PR #3, opened 2017-01-20
Poll for minions and show them on the UI

mssola:


mssola:
Supersedes #2

app/models/minion.rb
```diff
@@ -0,0 +1,4 @@
+# Minion represents the minions that have been registered in this application.
+class Minion < ApplicationRecord
+  default_scope { order hostname: :asc }
```
jimmykarily: I'm a little late but I suggest we avoid default_scope in general because it is hard to override. We can always use a named scope instead.

mssola: It should be possible to override with the [unscoped](http://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/unscoped) method no ? Otherwise, I'd agree to remove this line altogether to be honest 😬

jimmykarily: I didn't know about ```unscoped```. Probably because I removed ```default_scope``` from my options and never checked back again :). It might still be an issue if you have other scopes applied (chained) and you only want to remove the default_scope one. That's an edge case so we can leave it as is.

mssola: Right. That's why I said that we can remove it if you want. I think it has more drawbacks than advantages.

jimmykarily: agreed

PR #4, opened 2017-01-23
Add k8s manifest templates and script to raise all required services

ereslibre:


k8s/manifest-templates/pharos.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: Pod
+metadata:
+  name: pharos
+  labels:
+    name: pharos
+spec:
+  hostNetwork: true
+  containers:
+  - name: pharos-dashboard
+    image: ereslibre/pharos-dev:latest
```
ereslibre: Here we could take two different ways:

1. Build the image locally first. The kubelet in this case would never pull any image, it would be tagged something like `pharos:development` and the first step would be to build the image locally with `docker build -t pharos:development .`.

2. Rely on a remote image. Not sure in which account should be (opensuse?) as `pharos-development` or `pharos:development`? In this case we could add an automated build once this PR gets merged so we have a Dockerfile to instruct the automated build where to look for.

README.md
```diff
@@ -16,6 +16,29 @@ and start the server:
 - `PHAROS_KUBERNETES_CERT_DIRECTORY`: directory where the Kubernetes
   certificates are located.
 
+## Development
+
+You can develop on pharos with all its required dependencies right away using the `start` script
+within the `k8s` folder. You will need a standalone kubelet running in your machine.
```
jimmykarily: Maybe add a link to [this page](https://kubernetes.io/docs/admin/static-pods/)?

README.md
```diff
@@ -16,6 +16,29 @@ and start the server:
...
+`> ./start`
+
+This will start the pharos container as well as other required containers, such as `mariadb` for now.
+
+Your pharos folder will be mounted on the pharos container, so any change you do locally to pharos
+will be seen inside the container. Once that the service is up, you should be able to see the
+pharos service at `http://localhost:3000/` on your local machine.
+
+After any change that you have performed you can run any rails/rake task as if it was local, with
+the docker prefix, like `docker exec -it DASHBOARD_CONTAINER_ID rspec`.
```
jimmykarily: Can we specify a name for the generated containers? This command is quite big and I would like to create an alias in my bashrc so I don't have to type it all the time. I will need to have a consistent container name to do that.

k8s/manifest-templates/pharos.yaml
jimmykarily: It should be the same, building the image of fetching it unless someone has to change the Dockerfile for some reason in branch. In that case, we would build it locally and work with it as long as the branch is not merged and then we would push it to the registry. I suggest we use the ```imagePullPolicy: Never``` option to avoid confusion when the ```latest``` tag is used ([link](https://kubernetes.io/docs/user-guide/pods/multi-container/#containers)).
also.

k8s/manifest-templates/pharos.yaml
```diff
@@ -0,0 +1,46 @@
...
+    env:
+    - name: RAILS_ENV
+      value: development
+    - name: SECRET_KEY_BASE
+      value: fc0ece3811c0b1e0118dd7d619b385955e3ed146b37315d67d1c2d201c42531da398941ef27aab42493fff23afcba540390a26db51b5889e96d850a14178aabdc1926c013936a3c41abdbeb696ce440685c5d7be3a6c548434699d526a9e37657a76a43ca975b824905fd949f7e8d50aa1196a58e7e79dcfb08e2e397ee4f7b7
+    - name: PHAROS_DB_HOST
+      value: "127.0.0.1"
+    - name: PHAROS_DB_PORT
+      value: "3306"
```
jimmykarily: I think this will conflict with the host's MySQL (if any). I think some linux applications require MySQL so it's not safe to assume nobody runs MySQL on the host. Not using the host's network would mean that we would find our app in ```http://localhost:3000``` and we would need to seek out for the container's IP which is too much fuss so I'm not sure if its worth it. Shall we stick to the host's network for now?

ereslibre: I would go with host network for now, we can even move the mysql port because we are probably going to interact with it through rails at all times, but I'd go with the defaults for now.

k8s/manifest-templates/pharos.yaml
ereslibre: Yes, that's why I suggest using `pharos:development` so the kubelet won't try to fetch it every time (that only happens with the `latest` tag). With `imagePullPolicy: Never` we would be lacking any kind of pulling (but we want to use an image on the registry).

jimmykarily: Yes I know. I prefer to explicitly pull the image with ```docker pull``` when needed. Otherwise we must always keep this ```latest``` anomaly in mind.

README.md
ereslibre: You can alias it in some way, like `docker exec -it $(docker ps | grep pharos-dashboard) "$@"`. The kubelet will choose a fancy name for it so I'd just delegate this to docker + some grepping.

jimmykarily:
Can you split the Gem updates to a separate commit? They are irrelevant to the development env setup I think.

README.md
jimmykarily: That will work, thanks.

k8s/manifest-templates/pharos.yaml
jimmykarily: ok lets do that.

ereslibre:
Yes, doing that, it feels pretty wrong to mix those changes. I am pushing directly to pharos the gem update so the image and the code dependencies match (otherwise rails won't come up). And with our current setup, rails won't start because of https://github.com/cowboyd/therubyracer/issues/427.

jimmykarily:
ok I tryied it locally and it works like a charm. I only think we should use ```imagePullPolicy: Never``` on all images because the kubelet and docker logs doesn't inform you in any way when an image is being pulled. And then you are simply waiting for the env to come up without any information on what is going on. If we don't use the ```imagePullPolicy``` then we should never use the ```latest``` tag. Other than that, LGTM 👍 and the rest.

ereslibre:
I think it would be good for the script to work right away without the need to build the images (note that you would also need to build salt-master, salt-minion, salt-api, mariadb...). The kubelet won't report pull progress, but one can check network activity (not perfect, but works...). The idea was to create an automated build after this gets merged so we can create an image somewhere referring to this Dockerfile, and naming the image something like `pharos:development` or something like that (note that the tag is `development`, not `latest`, so the kubelet won't try to always pull). But for doing that we need to first include a Dockerfile inside pharos (or maybe don't use automated build at all and use other registry instead of the docker hub).

README.md
ereslibre: Updated.

ereslibre:
Removed Gemfile.lock changes.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
+.gitignore
+k8s
```
mssola: I'd prefer if the directory was named `kubernetes` since maybe all developers/users are not aware of the shorthand.

ereslibre:
Renamed!

PR #5, opened 2017-01-24
Handle new minion events

jimmykarily:
Handle "minion-start" events and write a new Minion in the database when needed.

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+  validates :hostname, uniqueness: true
 end
--- a/app/models/salt_event.rb
+++ b/app/models/salt_event.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+# This class wraps salt's events. Salt stores the events in the same database
+# we use for this application. This model is mapped to the same table salt-master
+# writes events into.
+class SaltEvent < ApplicationRecord
+  # Processing of a single event shouldn't take more than than 5 minutes.
+  # We consider events which were taken 5 minutes ago but not yet processed as
+  # if the worker died so we reassign them for processing.
+  PROCESS_TIMEOUT_MIN = 5
+  NO_EVENTS_TIMEOUT_SEC = 5
+
+  scope :not_processed, -> { where(processed_at: nil) }
+  scope :not_assigned, -> { where(worker_id: nil) }
+
+  # An event can be assigned to #my_worker when any of the following is true:
+  # - the event is not assigned to any worker
+  # - the event is already assigned to #my_worker
+  # - the event is assigned to any worker but more than
+  #   PROCESS_TIMEOUT_MIN minutes ago
+  #
+  # NOTE [Rails 5] Split this in separate scopes when we move to Rails 5 where
+  # ActiveRecord supports OR queries.
+  scope :assignable_to_worker, lambda { |worker_id|
+    where("worker_id IS NULL OR (worker_id = ?) OR taken_at < ?",
+          worker_id, PROCESS_TIMEOUT_MIN.minutes.ago)
+  }
+
+  # Processes one salt event at a time and sleeps
+  # NO_EVENTS_TIMEOUT_SEC seconds if none is available for processing.
+  def self.process(worker_id:)
+    # :nocov:
+    logger.info "Salt event processor #{worker_id} started at #{Time.current}"
+
+    loop do
+      sleep(NO_EVENTS_TIMEOUT_SEC) unless process_next_event(worker_id: worker_id)
+    end
+    # :nocov:
+  end
+
+  # Fetches the next processable event for the specified worker and calls
+  # process! on it. It returns false if no event is available for processing
+  # by the specified worker.
+  def self.process_next_event(worker_id:)
+    # rubocop:disable SkipsModelValidations
+    taken_event = SaltEvent
+                  .not_processed.assignable_to_worker(worker_id)
+                  .limit(1).update_all(worker_id: worker_id, taken_at: Time.current)
+
+    return false if taken_event.zero?
+
+    SaltEvent.where(worker_id: worker_id, processed_at: nil)
+             .where.not(taken_at: nil).limit(1).first.process!
+  end
+
+  # Memoizes the @handler instance which is responsible to process this event.
+  def handler
+    return @handler if @handler
+
+    case tag
+    when "minion_start"
+      @handler = SaltHandler::MinionStart.new(self)
+    end
+  end
+
+  # Calls the handler and marks this event as processed.
+  def process!
+    handler&.process_event
+
+    update_column(:processed_at, Time.current)
+  end
+
+  def parsed_data
+    JSON.parse(data)
+  end
+end
--- a/db/migrate/20170124141701_add_uniq_index_on_minions_id.rb
+++ b/db/migrate/20170124141701_add_uniq_index_on_minions_id.rb
@@ -1,0 +1,11 @@
+class AddUniqIndexOnMinionsId < ActiveRecord::Migration[5.0]
+  def up
+    remove_index :minions, :hostname
+    add_index :minions, :hostname, unique: true
+  end
+
+  def down
+    remove_index :minions, :hostname
+    add_index :minions, :hostname
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,13 +10,43 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170111154919) do
+ActiveRecord::Schema.define(version: 20170124141701) do
+
+  create_table "jids", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+    t.string "jid",                   null: false
+    t.text   "load", limit: 16777215, null: false
+    t.index ["jid"], name: "jid", unique: true, using: :btree
+  end
 
   create_table "minions", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.string   "hostname"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["hostname"], name: "index_minions_on_hostname", using: :btree
+    t.index ["hostname"], name: "index_minions_on_hostname", unique: true, using: :btree
+  end
+
+  create_table "salt_events", id: :bigint, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+    t.string   "tag",                                                                null: false
+    t.text     "data",         limit: 16777215,                                      null: false
+    t.datetime "alter_time",                    default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.string   "master_id",                                                          null: false
+    t.datetime "taken_at"
+    t.datetime "processed_at"
+    t.string   "worker_id"
+    t.index ["tag"], name: "tag", using: :btree
+  end
+
+  create_table "salt_returns", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+    t.string   "fun",        limit: 50,                                            null: false
+    t.string   "jid",                                                              null: false
+    t.text     "return",     limit: 16777215,                                      null: false
+    t.string   "id",                                                               null: false
+    t.string   "success",    limit: 10,                                            null: false
+    t.text     "full_ret",   limit: 16777215,                                      null: false
+    t.datetime "alter_time",                  default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.index ["fun"], name: "fun", using: :btree
+    t.index ["id"], name: "id", using: :btree
+    t.index ["jid"], name: "jid", using: :btree
   end
 
   create_table "users", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
--- a/spec/models/salt_event_spec.rb
+++ b/spec/models/salt_event_spec.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe SaltEvent do
+  describe ".process_next_event" do
+    let(:leftover_event) do
+      # not timed out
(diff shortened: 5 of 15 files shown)


db/schema.rb
```diff
@@ -10,13 +10,58 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170111154919) do
+ActiveRecord::Schema.define(version: 20170124141701) do
+
+  create_table "delayed_jobs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
jimmykarily: oops this is a leftover. Removing.

spec/factories/minion.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  factory :minion do
+    sequence(:hostname) { |n| "Minion#{n}" }
```
jimmykarily: @mssola introduced a similar factory in another PR  (minions_factory.rb). I will use that instead.

spec/factories/salt_event.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
jimmykarily: I see that we use ```_factory ``` suffix for these files. Adding that.

app/models/salt_event.rb
```diff
@@ -0,0 +1,48 @@
...
+    logger.info "Salt event processor #{worker_id} started at #{Time.current}"
+
+    loop do
+      sleep(NO_EVENTS_TIMEOUT_SEC) if !process_next_event(worker_id: worker_id)
+    end
+  end
+
+  def self.process_next_event(worker_id:)
+    taken_event = SaltEvent.where(
+      "processed_at IS NULL AND (worker_id IS NULL OR (worker_id = ?) OR taken_at < ?)",
```
ereslibre: I'd try to use scopes here. Something like `SaltEvent.not_processed.taken_by_or_nobody(worker_id).taken_before(PROCESS_TIMEOUT_MIN.minutes.ago)`.

app/models/salt_event.rb
```diff
@@ -0,0 +1,48 @@
...
+      worker_id, Time.current - PROCESS_TIMEOUT_MIN.minutes)
+    .limit(1).update_all(worker_id: worker_id, taken_at: Time.current)
+
+    return false if taken_event == 0
+
+    SaltEvent.where(worker_id: worker_id, processed_at: nil).
+      where("taken_at IS NOT NULL").limit(1).first.process!
```
ereslibre: We should use activerecord where possible, here: `where.not(taken_at: nil)`.

spec/models/salt_event_handlers/minion_start_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe SaltEventHandlers::SaltEvent do
+  let(:salt_event) {
+    event_data = "{\"_stamp\": \"2017-01-24T13:30:20.794326\", "\
```
ereslibre: We could avoid all this escaping if we used `event_data = ''`, it's somewhat easier to read. Even plain ruby hashes with a `.to_json`.

spec/models/salt_event_spec.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe SaltEvent do
+  describe ".process_next_event" do
+    it "processes new-minion events" do
+      event_data = "{\"_stamp\": \"2017-01-24T13:30:20.794326\", "\
```
ereslibre: Same here...

spec/models/salt_event_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      expect {
+        SaltEvent.process_next_event(worker_id: "this_worker")
+      }.to change{ event.reload.processed_at }.from(nil).to(Time)
+
+      expect(event.worker_id).to eq("this_worker")
+    end
+  end
+
+  describe "#process" do
```
ereslibre: Is the `#` a typo?

app/models/minion.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+  validates :hostname, uniqueness: true
```
mssola: `hostname` is an index already. Is this really needed ?

spec/models/salt_event_spec.rb
jimmykarily: No there is a convention to use "." for class method and "#" for instance methods. Not sure how wide spread it is. Do we follow a different rule?

app/models/salt_event.rb
```diff
@@ -0,0 +1,48 @@
+class SaltEvent < ApplicationRecord
```
mssola: Add class documentation

app/models/salt_event.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  def process!
+    handler.process_event if handler
+
+    update_column(:processed_at, Time.current)
+  end
+
+  def parsed_data
```
mssola: Maybe this can be memoized ?

app/models/salt_event_handlers/minion_start.rb
```diff
@@ -0,0 +1,13 @@
+class SaltEventHandlers::MinionStart
```
mssola: I'd rename SaltEventHandlers to simply Handlers (and also the directory). I think that the SaltEvent part is already assumed. Plus, documentation

app/models/salt_event_handlers/minion_start.rb
```diff
@@ -0,0 +1,13 @@
...
+  attr_reader :salt_event, :event_parsed_data
+
+  def initialize(salt_event)
+    @salt_event = salt_event
+    @event_parsed_data = @salt_event.parsed_data
+  end
+
+  def process_event
+    # false if a minion with this hostname already exists (uniqueness validation)
+    Minion.new(hostname: salt_event.parsed_data["id"]).save
```
mssola: Why not simple `Minion.create` ?

spec/models/salt_event_handlers/minion_start_spec.rb
mssola: In this case using single quotes is accepted, so please, use single quotes. Or even heredocs.

app/models/minion.rb
jimmykarily: It shouldn't. Just taking one extra precaution to avoid invalid data in the database.

app/models/salt_event.rb
jimmykarily: Added

app/models/minion.rb
mssola: Cool

app/models/salt_event.rb
jimmykarily: done

app/models/salt_event.rb
jimmykarily: changed

spec/models/salt_event_handlers/minion_start_spec.rb
jimmykarily: Changed to Hash -> to_json

spec/models/salt_event_spec.rb
jimmykarily: Changed to Hash -> to_json

spec/models/salt_event_spec.rb
mssola: I also follow this convention, so I think this is fine

app/models/salt_event.rb
jimmykarily: I'm not sure. The overhead is small and I'm afraid that if we ever change the data attribute (don't have a case yet) then parsed_data will return the old version leading to bugs. If we need to call parsed_data multiple times we should either way store it in a variable because in any case we would be invoking a method multiple times instead of accessing a variable.

app/models/salt_event_handlers/minion_start.rb
jimmykarily: Added documentation. Are you sure about ```Handlers```? What if we want a class that will handle something completely different?

app/models/salt_event_handlers/minion_start.rb
jimmykarily: The difference is that ```create``` will always return the Minion while ```save``` will return ```false``` if no Minion was created so we know if a minion was created or not. On the other hand none of these is a proper return value for a method named ```process_event```. The event has been processed in both cases so the return value should probably be ```true``` but that doesn't provide us any usable information.

app/models/salt_event.rb
mssola: Agreed

db/migrate/20170124141701_add_uniq_index_on_minions_id.rb
```diff
@@ -0,0 +1,11 @@
...
+  def up
+    remove_index :minions, :hostname
+    add_index :minions, :hostname, unique: true
+  end
+
+  def down
+    remove_index :minions, :hostname
+    add_index :minions, :hostname
+  end
+end
```
ereslibre: TBH I don't really like the idea of removing an index for adding it afterwards (the table will be empty, yeah), but maybe you can just modify the original migration at this point.

spec/models/salt_event_spec.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe SaltEvent do
+  describe ".process_next_event" do
+    let(:leftover_event) do
+      # not timed out
+      FactoryGirl.create(
+        :salt_event, worker_id: "this_worker", processed_at: nil,
+        taken_at: Time.current - (SaltEvent::PROCESS_TIMEOUT_MIN - 2).minutes
```
ereslibre: Here I'd use `(SaltEvent::PROCESS_TIMEOUT_MIN - 2).minutes.ago`, is more idiomatic.

spec/models/salt_event_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      )
+    end
+
+    let(:timedout_event) do
+      FactoryGirl.create(
+        :salt_event, worker_id: "other_worker", processed_at: nil,
+        taken_at: Time.current - (SaltEvent::PROCESS_TIMEOUT_MIN + 2).minutes
```
ereslibre: Same here.

app/models/salt_event_handlers/minion_start.rb
mssola: Sure, I just wanted to simplify the name :) What about `SaltHandlers` ?

app/models/salt_event.rb
```diff
@@ -0,0 +1,76 @@
...
+  # - the event is not assigned to any worker
+  # - the event is already assigned to #my_worker
+  # - the event is assigned to any worker but more than
+  #   PROCESS_TIMEOUT_MIN minutes ago
+  #
+  # NOTE [Rails 5] Split this in separate scopes when we move to Rails 5 where
+  # ActiveRecord supports OR queries.
+  scope :assignable_to_worker, lambda { |worker_id|
+    where("worker_id IS NULL OR (worker_id = ?) OR taken_at < ?",
+      worker_id, Time.current - PROCESS_TIMEOUT_MIN.minutes)
```
jimmykarily: Agreed, changed.

app/models/salt_event.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    case tag
+    when "minion_start"
+      @handler = SaltEventHandlers::MinionStart.new(self)
+    end
+  end
+
+  # Calls the handler and marks this event as processed.
+  def process!
+    handler&.process_event if handler
```
jimmykarily: I forgot to remove the check after I switched to save navigation operator. Fixing.

app/models/salt_event_handlers/minion_start.rb
```diff
@@ -0,0 +1,20 @@
...
+
+# This class is responsible to handle the salt events with tag "minion_start".
+# When such an event occurs, we want the minion to be saved in our database
+# if not already there.
+class SaltEventHandlers::MinionStart
+  attr_reader :salt_event, :event_parsed_data
+
+  def initialize(salt_event)
+    @salt_event = salt_event
+    @event_parsed_data = @salt_event.parsed_data
```
jimmykarily: I don't want to memoize parsed_data for the reasons explained [here](https://github.com/SUSE/pharos/pull/5#discussion_r97799805). Then I though I would need that data many times and I created this instance variable but after all I didn't even use it. I think I should remove this variable as you suggest but not memoize the ```parsed_data``` method. If some code needs parsed data mutliple times I prefer to store it in a local var.

If both of you @mssola @ereslibre think we won't step into bugs I will memoize the method.

db/migrate/20170124141701_add_uniq_index_on_minions_id.rb
jimmykarily: Changing past migrations is bad because it will force everyone to rebuilt their database from scratch. Somebody might need their seed data for development and forcing them to drop everything is not the way to go. Of  course its just us for now but I prefer to do it the right way anyway.

spec/models/salt_event_spec.rb
jimmykarily: fixed.

spec/models/salt_event_spec.rb
jimmykarily: fixed

app/models/salt_event_handlers/minion_start.rb
jimmykarily: So combining your comments I will go with ```SaltHandler```.

app/models/salt_event.rb
```diff
@@ -0,0 +1,76 @@
...
+  # Processing of a single event shouldn't take more than than 5 minutes.
+  # We consider events which were taken 5 minutes ago but not yet processed as
+  # if the worker died so we reassign them for processing.
+  PROCESS_TIMEOUT_MIN = 5
+  NO_EVENTS_TIMEOUT_SEC = 5
+
+  scope :not_processed, -> { where(processed_at: nil) }
+  scope :not_assigned, -> { where(worker_id: nil) }
+
+  # An event can be assigned to #my_worker when any of the following it true:
```
mssola: type: it -> is

mssola:
From my side, just fix that typo and squash everything into a single commit. Great work 👏

PR #7, opened 2017-01-25
Call orchestration from pharos dashboard

ereslibre:
We will randomly select roles for discovered nodes, and call the
async orchestration afterwards.

Roles will be set through "roles" grain.


.gitignore
```diff
@@ -11,3 +11,4 @@
 .rspec
 /kubernetes/tmp
 /kubernetes/manifests
+/kubernetes/salt/config/minion.d/_schedule.conf
```
jimmykarily: this is automatically generated on my machine too. Maybe we need to gitignore it?

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,18 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  def bootstrap
+    salt = Pharos::Salt.new
+    available_roles = %i(master)
+    salt.minions.keys.each do |minion|
+      Minion.transaction do
+        Minion.find_or_create_by!(hostname: minion).assign_role(available_roles)
```
jimmykarily: This will need to change as soon as we merge [this](https://github.com/SUSE/pharos/pull/5). We should only create a Minion in the database when we have information that this minion exists (e.g. when we get a minion-start event). That said, we only need to apply the master role to an existing minion (unless we have none).

.gitignore
ereslibre: You are reading the .gitignore :)

app/controllers/nodes_controller.rb
ereslibre: The master role will only be applied to one node. After that role is assigned (it's given priority for assignment), we will just assign minion roles.

jimmykarily: I think there is a bug here. I think ```assign_roles``` code assumes that any assigned role is removed from the available_roles array, which is true while inside the method but since Ruby is pass-by-value, that change isn't reflected on the available_roles here. So all the roles are present in every iteration which might result in multiple masters.

.gitignore
jimmykarily: :D sorry

app/controllers/nodes_controller.rb
ereslibre: It's pass by value, but there is no array cloning here. You are effectively modifying the available_roles. You can check this with a trivial example:

```
def hello
  a = %i(1 2 3)
  hi a
  puts a
end

def hi(a)
  a.delete_at 0
end

hello
```

But still I agree this isn't clear enough. Fixing it.

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,18 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  def bootstrap
+    salt = Pharos::Salt.new
+    available_roles = %i(master)
+    salt.minions.keys.each do |minion|
```
jimmykarily: I think we should have a single source of truth. This will make a synchronous request to salt-api to return the available minions (which I think is the equivalent of [```salt '*' grains.items```](https://github.com/saltstack/salt/blob/develop/salt/netapi/rest_cherrypy/app.py#L1156)). If we have the mechanism in place to know when a minion is down (which we should at some point), then maybe we can simply use the available minions from the database.

app/controllers/nodes_controller.rb
jimmykarily: You are right in both, it works and it is not very clear. An array is a reference so yes it is still pass-by-value. Mea culpa.

jimmykarily: But what if we don't have a minion with this hostname? In that case you are creating it right here which is not a responsibility of this method (because it might require a number of other actions which will be implemented [somewhere else](https://github.com/SUSE/pharos/pull/5/files#diff-784d5f9c206cafc566e70b874a69b0da)).

app/controllers/nodes_controller.rb
ereslibre: We'd need to accomodate this code after other changes make it inside the repo. I agree about the single source of truth, this is a way of getting the nodes without blocking because of other WIP PR, it will go away.

jimmykarily: I think I'm done with the other one. I'll check and I'll remove the [WIP] flag. Maybe we should merge that first and update this code.

app/models/minion.rb
```diff
@@ -1,4 +1,25 @@
 # frozen_string_literal: true
+
+require 'pharos/salt'
+
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+
+  def assign_role(available_roles)
```
jimmykarily: I would like it better if this method took only one role and assigned that. You can move the random role selection outside of this method.

lib/pharos/salt.rb
```diff
@@ -16,51 +17,102 @@ def initialize
...
-
-      id  = mid ? "/#{mid}" : ""
-      res = perform_request("/minions#{id}", "get")
+    def minions
+      res = perform_request(endpoint: "/minions", method: "get")
       JSON.parse(res.body)["return"].first
     end
 
+    # Return information for the minion linked to this instance.
+    def minion
```
jimmykarily: The methods of this class behave differently when a minion id is assigned and this can lead to confusion. For example if we see a line of code like: ```salt.minion``` somewhere in the code we don't know if it returns nil because a minion was not provided (bug?) or because the minion doesn't exist. Also the request will be sent even if no minion id is specified which will always return nil (because of the last line in this method). 

I think this library should not even be instantiated but rather be used as a module providing a number of methods directly mapped to API requests. Then the Minion model can implement all the methods related to that minion that will call the methods of this class with the right arguments.

app/controllers/nodes_controller.rb
ereslibre: Ok, I will just take minions from the database with unassigned roles.

app/models/minion.rb
ereslibre: Yes.

app/controllers/nodes_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'pharos/salt'
```
mssola: Strings go with double quotes. Rubocop will complain 😉

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,20 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  def bootstrap
```
mssola: Documentation

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,20 @@ def show
...
+    available_roles = %i(master)
```
mssola: Avoid using %i

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,20 @@ def show
...
+    Minion.where(role: nil).each do |minion|
```
mssola: Use `find_each` instead of `each`

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,20 @@ def show
...
+      if available_roles.blank?
+        random_role = :minion
+      else
+        random_role = available_roles.delete_at(rand(available_roles.length))
+      end
+      minion.assign_role(role: random_role)
+    end
```
mssola: nitpicking: add an empty line here

app/controllers/nodes_controller.rb
```diff
@@ -16,6 +18,20 @@ def show
...
 
+  def bootstrap
+    available_roles = %i(master)
+    Minion.where(role: nil).each do |minion|
+      if available_roles.blank?
+        random_role = :minion
+      else
+        random_role = available_roles.delete_at(rand(available_roles.length))
+      end
+      minion.assign_role(role: random_role)
```
mssola: For now this is ok, but this will take a lot of time. Would it be possible to add a websocket that informs the client what's going on ? For example, before you perform this call, send a "minion x will be deployed as a $role" message, and after that another message. Moreover, maybe it would also be worth it to send an initial message telling the client the amount of minions that are going to be assigned.

With all this, the client will be able to show more information to the user, so they are not clueless on what's going on 😄 

If you agree, maybe we can submit a new issue for this.

app/models/minion.rb
```diff
@@ -1,4 +1,19 @@
 # frozen_string_literal: true
+
+require 'pharos/salt'
+
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+
```
mssola: remove empty line

app/models/minion.rb
```diff
@@ -1,4 +1,19 @@
...
+  def assign_role(role:)
```
mssola: Add documentation 😄

app/models/minion.rb
```diff
@@ -1,4 +1,19 @@
...
+    self.update_column :role, role
+    salt.assign_role mid: self.hostname, role: role
+  end
+
+  private
+
+  def salt
```
mssola: doc

app/models/minion.rb
```diff
@@ -1,4 +1,19 @@
...
+    @salt ||= Pharos::Salt.new
+  end
+
```
mssola: Remove empty line.

app/views/nodes/index.html.slim
```diff
@@ -1,4 +1,9 @@
 h1
+  | Bootstrap cluster
+
+= link_to "Bootstrap cluster", bootstrap_nodes_path, method: :post, class: "btn btn-success btn-lg"
```
mssola: Add a view test for this (check the `link?` helper method)

db/migrate/20170124170136_add_role_to_minions.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToMinions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :minions, :role, :string, index: true, after: :hostname
```
mssola: I think it makes more sense to have this as an integer and then in the model add an enum that maps that ([as we do in portus and team members](https://github.com/SUSE/Portus/blob/master/app/models/team_user.rb#L25)).

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
 
+    # Return information for the minion linked to this instance.
+    def minion(mid:)
+      res = perform_request(endpoint: "/minions/#{mid}", method: "get")
+      JSON.parse(res.body)["return"].first[mid]
+    end
+
+    class << self
+      # Call the orchestration
+      def orchestrate
```
mssola: If it's only one method, it's better to simply write `def self.orchestrate`

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+        perform_run_request(data: { client: "runner_async",
+                                    fun: "state.orchestrate",
+                                    kwargs: { mods: "orch.kubernetes" } })
+      end
+    end
+
+    def has_roles?(mid:)
```
mssola: Documentation. Plus, does this need to be public ?

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+      !minion(mid: mid)["roles"].blank?
+    end
+
+    def map_role(role:)
```
mssola: Documentation, and don't use %i

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+      case role
+      when :master then %i(kube-master etcd)
+      when :minion then %i(kube-minion)
+      end
+    end
+
+    def assign_role(mid:, role:)
```
mssola: You guessed it, documentation 😁

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+      append_grain mid: mid, key: "roles", val: map_role(role: role)
+    end
+
     protected
 
+    def append_grain(mid:, key:, val:)
```
mssola: you now the drill 😋

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+      perform_request(endpoint: "/minions",
+                      method: "post",
+                      data: { client: "local",
```
mssola: I'm sure rubocop complains here 🤓

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
     def login!
-      hsh = {
-        username: ENV["PHAROS_SALT_USER"],
-        password: "password",
-        eauth:    "auto"
-      }
-
-      res = perform_request("/login", "post", hsh)
-      @token = res["X-Auth-Token"]
+      @token ||= perform_request(endpoint: "/login",
```
mssola: Same as before. They should be aligned

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
       req["X-Auth-Token"] = @token if @token
-      req.set_form_data(data)
+      req.body = data.to_json unless data.blank?
 
       opts = { use_ssl: false, open_timeout: 2 }
       Net::HTTP.start(uri.hostname, uri.port, opts) { |http| http.request(req) }
     end
+
+    class << self
+      def auth_details
```
mssola: documentation. Moreover, can't you just freeze the hash ?

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+        {
+          username: ENV["PHAROS_SALT_USER"],
+          password: ENV["PHAROS_SALT_PASSWORD"],
+          eauth:    "pam"
+        }
+      end
+
+      def hostname
```
mssola: documentation

lib/pharos/salt.rb
```diff
@@ -16,54 +15,113 @@ def initialize
...
+        "#{ENV["PHAROS_SALT_HOST"]}:#{ENV["PHAROS_SALT_PORT"]}"
+      end
+
+      # Perform an HTTP request to the Salt API server in a one-shot fashion.
+      def perform_run_request(data: {})
```
mssola: Can't you just merge this method with `perform_run_request` ? They look really similar.

lib/pharos/salt.rb
```diff
@@ -5,8 +5,7 @@ module Pharos
   # This class offers the integration between ruby and the Saltstack API.
   class Salt
```
mssola: You will need to provide test for this

app/controllers/nodes_controller.rb
ereslibre: Are we going to have this big number of minions when bootstrapping? `find_each` is more focused for batched finds (`find_in_batches`). The default batch is `1000` that I doubt we will reach on this result, so they are almost equivalent, but it's ok, we can use `find_each`.

app/controllers/nodes_controller.rb
ereslibre: Yes, we can open an issue and do that at any later time.

lib/pharos/salt.rb
ereslibre: Sure

lib/pharos/salt.rb
ereslibre: Style preference I guess, changing that.

lib/pharos/salt.rb
ereslibre: Yes, at least for now. If you have an already provisioned machine with roles assigned, you don't want to assign more roles to it, and that needs to be checked from the outside, or maybe ignore the assignment on `assign_role` method.

lib/pharos/salt.rb
ereslibre: Yes.

lib/pharos/salt.rb
ereslibre: This one only POSTS and won't take or generate any token, it's a bit different but we might want to merge them, yes. The one-shot one will probably only be used from static methods (no @token required or generated).

lib/pharos/salt_minion.rb
```diff
@@ -0,0 +1,53 @@
...
+    # Appends a grain with key and val to this minion.
+    def append_grain(key:, val:)
+      perform_request(endpoint: "/minions",
+                      method:   "post",
+                      data:     {
+                        client: "local",
+                        tgt:    @mid,
+                        fun:    "grains.append",
+                        kwarg:  { key: key,
+                                  val: val }
```
mssola: This last `kwarg`  can be put in the same line (too many newlines :P)

app/models/minion.rb
```diff
@@ -1,4 +1,25 @@
...
+    Minion.transaction do
+      update_column :role, role
+      salt.assign_role role: role
+    end
+    true
+  end
+
+  # Returns the proxy for the salt minion
+  def salt
+    @salt ||= Pharos::SaltMinion.new mid: hostname
```
jimmykarily: [MINOR] Maybe we don't need a keyword argument for this one but if we use one I suggest we rename it to minion_id. It's more verbose but it might not be obvious to someone as "mid" because after all we set it to "hostname".

db/migrate/20170124170136_add_role_to_minions.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToMinions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :minions, :role, :integer, index: true, after: :hostname
```
jimmykarily: I didn't know about ```after```. Good one 👍 .

lib/pharos/salt_api.rb
```diff
@@ -0,0 +1,86 @@
...
+      #
+      # Returns the response object.
+      def perform_run_request(data: {})
+        uri = URI.join("http://#{hostname}", "run")
+
+        req = Net::HTTP::Post.new uri
+        req["Accept"]       = "application/json; charset=utf-8"
+        req["Content-Type"] = "application/json; charset=utf-8"
+
+        req.body = data.merge(auth_details).to_json
```
jimmykarily: Can't we just use a token for this request too? If yes then this method is just ```perform_request``` with endpoint = "run", right?

jimmykarily: I the purpose is to avoid the extra request for login for one-off commands then maybe we should merge the 2 methods in one. That method will use a token if a "client" exists and will use the credentials if not.

lib/pharos/salt_minion.rb
```diff
@@ -0,0 +1,52 @@
...
+require "pharos/salt_api"
+
+module Pharos
+  # This class represents a salt minion
+  class SaltMinion
+    include SaltApi
+
+    # Initializes a new salt minion identified by mid.
+    def initialize(mid:)
+      @mid = mid
```
jimmykarily: [minor] add an ```attr_reader``` for this

lib/pharos/salt_minion.rb
```diff
@@ -0,0 +1,52 @@
...
+      JSON.parse(res.body)["return"].first[@mid]
+    end
+
+    # Check if this minion has any role assigned.
+    def roles?
+      !info["roles"].blank?
+    end
+
+    # Map our high-level roles to more salt specialized roles.
+    def map_role(role:)
```
jimmykarily: [MINOR] Maybe we can replace this method with a constant:

```
ROLES_MAP = {
  master: ["kube-master", "etcd"],
  minion: ["kube-minion"]
}

Calling ```ROLES_MAP[:master]``` or ```map_role(role: :master)``` isn't so different but we avoid a method invocation and tests for trivial things.
```

spec/models/minion_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require "rails_helper"
```
jimmykarily: I think it will be more readable if we add empty lines between examples.

lib/pharos/salt_api.rb
ereslibre: No, the whole idea of the `/run` endpoint is to provide the authentication details within the data. It is similar but different in important ways to `perform_request`. I think we could end up with a mangled and hard to use method:

- It takes `data` only as argument.
- It will never save `token` or use it. Always send raw [auth details](https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html#post--run). Token is ignored, auth details are sent within the post request.
- It is always a `POST` request.

I'd keep them separate for the sake of readability.

lib/pharos/salt_minion.rb
ereslibre: Yes, I agree.

lib/pharos/salt_api.rb
jimmykarily: With all this duplicate code it is hard to tell the difference between the 2 methods. So instead of this:

```
perform_run_request(data: { client: "runner_async", fun:    "state.orchestrate", kwargs: { mods: "orch.kubernetes" }})
```
one would call this:

```
perform_request(endpoint: "/run", method: "POST", data: { client: "runner_async", fun:    "state.orchestrate", kwargs: { mods: "orch.kubernetes" }})
```

Imho the second version is as readable as the first and we avoid one extra method and all the duplication.

app/models/minion.rb
ereslibre: Done.

db/migrate/20170124170136_add_role_to_minions.rb
ereslibre: :tada:

lib/pharos/salt_api.rb
ereslibre: No, they are not equivalent. `perform_run_request` will merge your data with the auth data `(username, password, eauth)`. Also, the `/run` endpoint only makes sense to be run with `POST`. You also don't use any kind of token with `perform_run_request`. They are similar, but are different in really important ways.

spec/models/minion_spec.rb
ereslibre: Done.

lib/pharos/salt_api.rb
jimmykarily: Yes that's why I suggested my second comment above. We can use a token if a client exists (which means "use a token if this is not an one-off command") and send the credentials in params when it doesn't. If we create a different method for each sub-case then what do we need perform_request for? I thought this was the method capable of any request to the API.

ereslibre: Fixed, but we want to mess that method as less as possible :)

ereslibre:
Updated!

app/models/minion.rb
```diff
@@ -1,4 +1,25 @@
 # frozen_string_literal: true
+
+require "pharos/salt_minion"
+
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+  enum role: [:master, :minion]
+
+  # rubocop:disable SkipsModelValidations
```
mssola: Remember to enable it back again at the end of the method

spec/lib/pharos/salt_api_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(client.token).to be_nil
+    end
+
+    it "saves the token in the client token cache" do
+      VCR.use_cassette("salt/fetch_minion", record: :none) do
+        client.perform_request(endpoint: "/minions/minion1", method: "get")
+        expect(client.token).not_to be_nil
+      end
+    end
+
```
mssola: remove empty line

spec/lib/pharos/salt_api_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
```
mssola: same here

spec/lib/pharos/salt_minion_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  describe "minion roles" do
+    context "has roles" do
+      it "returns true for roles?" do
+        VCR.use_cassette("salt/fetch_minion_with_roles", record: :none) do
+          expect(described_class.new(minion_id: "minion1").roles?).to be true
+        end
+      end
+    end
```
mssola: Separate context blocks with an empty line :)

app/models/minion.rb
ereslibre: Done.

spec/lib/pharos/salt_minion_spec.rb
ereslibre: Done

spec/lib/pharos/salt_api_spec.rb
ereslibre: Done

spec/lib/pharos/salt_api_spec.rb
ereslibre: Done

lib/pharos/salt_api.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        req = Net::HTTP.const_get(method.capitalize).new(uri)
+        req["Accept"]       = "application/json; charset=utf-8"
+        req["Content-Type"] = "application/json; charset=utf-8"
+        req["X-Auth-Token"] = token if token
+        if client.nil?
+          req.body = data.merge(auth_details).to_json
+        else
+          req.body = data.to_json unless data.blank?
+        end
```
jimmykarily: Maybe:

```
if token
  req["X-Auth-Token"] = token
  req.body = data.to_json unless data.blank?
else
  req.body = data.merge(auth_details).to_json
end
```

I think this is equivalent

lib/pharos/salt_api.rb
```diff
@@ -0,0 +1,79 @@
...
+          eauth:    "pam"
+        }.freeze
+      end
+
+      # Returns salt hostname for the current environment.
+      def hostname
+        "#{ENV["PHAROS_SALT_HOST"]}:#{ENV["PHAROS_SALT_PORT"]}"
+      end
+
+      # Returns true if it is a tokenless request
```
jimmykarily: Maybe add this link to the comments?
https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html#run

ereslibre: Done.

PR #8, opened 2017-01-26
Add minion missing test

ereslibre:
If we cannot set the role on the remote, make sure that the transaction
gets a rollback, so we have the same information on the local database
than the remote salt instance.

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -18,6 +18,8 @@ class Minion < ApplicationRecord
       salt.assign_role role: role
     end
     true
+  rescue Pharos::SaltApi::SaltConnectionException
+    false
   end
   # rubocop:enable SkipsModelValidations
 
--- a/lib/pharos/salt_api.rb
+++ b/lib/pharos/salt_api.rb
@@ -4,6 +4,24 @@ require "net/http"
 module Pharos
   # This class offers the integration between ruby and the Saltstack API.
   module SaltApi
+    class SaltConnectionException < StandardError; end
+
+    HTTP_EXCEPTIONS = [
+      SocketError,
+      Errno::ETIMEDOUT,
+      Net::ReadTimeout,
+      Net::OpenTimeout,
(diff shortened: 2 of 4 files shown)


app/models/minion.rb
```diff
@@ -18,6 +18,8 @@ def assign_role(role:)
       salt.assign_role role: role
     end
     true
+  rescue
```
jimmykarily: Since we know which exception we want to rescue from we'd better be explicit here:

```rescue Net::HTTPBadResponse```

mssola:
LGTM 👍 Just fix @jimmykarily's comment.

app/models/minion.rb
ereslibre: Actually this was just one of them. With HTTP lots of things can go wrong and we should be able to rescue from a lot more of possible exceptions (https://github.com/rainforestapp/http-exceptions/blob/master/lib/http/exceptions.rb#L7-L21). Should we rescue from all of them? A generic rescue could be hiding future problems and maybe it's not optimal.

jimmykarily: I suggest we rescue explicitly only what makes sense. For example I would rollback the transaction but I would reraise in the case of ```Errno::ECONNREFUSED``` because that probably means something is wrong.

mssola: > A generic rescue could be hiding future problems and maybe it's not optimal.

That was my train of thought. I'd prefer white-listing the exceptions that we want to catch. Maybe other exceptions have to be handled differently.

On another note, maybe we can do as we do in [portus](https://github.com/SUSE/Portus/blob/master/lib/portus/http_helpers.rb#L65). That is, to re-route HTTP exceptions into a handler. This way we can also provide better logging.

ereslibre:
Updated, and wrapped in a single salt exception (that in the future could include the original exception). This is really handy for SaltAPI users, because we are wrapping all http/socket related exceptions into a single one.

PR #12, opened 2017-02-01
Check that we are trying to reassign available roles if they fail

ereslibre:
If we have certain roles to be assigned (mandatory, not `minion` ones). We should retry to assign them if a previous assignment fails.

### DIFF
--- a/app/controllers/nodes_controller.rb
+++ b/app/controllers/nodes_controller.rb
@@ -22,14 +22,12 @@ class NodesController < ApplicationController
   # assigned role, assign a random role to it, and then call the salt
   # orchestration.
   def bootstrap
-    available_roles = [:master]
+    @available_roles = [:master]
     Minion.where(role: nil).find_each do |minion|
-      random_role = if available_roles.blank?
-        :minion
-      else
-        available_roles.delete_at(rand(available_roles.length))
+      random_role = @available_roles.pop || :minion
+      unless minion.assign_role(role: random_role)
+        @available_roles << random_role if random_role != :minion
       end
-      minion.assign_role(role: random_role)
     end
     Pharos::Salt.orchestrate
     redirect_to nodes_path
--- a/spec/controllers/nodes_controller_spec.rb
+++ b/spec/controllers/nodes_controller_spec.rb
@@ -95,5 +95,19 @@ RSpec.describe NodesController, type: :controller do
       end
       expect(response.status).to eq 302
     end
+
+    context "a role fails to be assigned" do
+      # rubocop:disable AnyInstance
+      before do
+        allow_any_instance_of(Minion).to receive(:assign_role).and_return(false)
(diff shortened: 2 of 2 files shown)


spec/controllers/nodes_controller_spec.rb
```diff
@@ -95,5 +95,19 @@
       end
       expect(response.status).to eq 302
     end
+
+    context "a role fails to be assigned" do
+      # rubocop:disable AnyInstance
+      before do
+        allow_any_instance_of(Minion).to receive(:assign_role).and_return(false)
```
ereslibre: I couldn't find a better way to mock this response in a stylish way without messing too much with the inners of the bootstrap method. Maybe this is a code smell, but if you come up with a better mocking I'd be happy to hear about it.

ereslibre: Note that this forces us to disable `AnyInstance` rubocop check, because the usage of `allow_any_instance_of` is discouraged. However at this time, we haven't still instantiated the elements. We could do so manually and provide them as an argument to the bootstrap action, but it looked strange.

app/controllers/nodes_controller.rb
```diff
@@ -22,14 +22,16 @@ def show
   # assigned role, assign a random role to it, and then call the salt
   # orchestration.
   def bootstrap
-    available_roles = [:master]
+    @available_roles = [:master]
```
jimmykarily: The use of an instance variable here is a little cryptic (this is only needed for the tests). I think it would be better if we changed the signature of this method to:

```
def bootstrap(roles=[:master])
  ...
  return assigned_roles
end
```
This way you don't need to change the method calls and you can check the return value for assigned roles.

mssola: @jimmykarily we cannot `return assigned_roles` because this is a method that responds to a request no ?

Otherwise, a suggestion/nitpick that I have is to move the `[:master]` into a constant in the beginning of the class:

```ruby
INITIAL_AVAILABLE_ROLES = [:master].freeze
```

And then assign the `@available_roles` to it. This is more of a cosmetic change, I know, but I think it reflects better the intention of this first initialization.

That being said, even if it smells ugly, I don't see another way around creating an instance variable honestly 😞

spec/controllers/nodes_controller_spec.rb
mssola: Fine by me

app/controllers/nodes_controller.rb
mssola: Maybe a comment should reflect why it's an instance variable ?

ereslibre:
We won't need this after https://github.com/SUSE/pharos/pull/13, that is the right way to approach the problem. This PR gets superseded by https://github.com/SUSE/pharos/pull/13.

PR #14, opened 2017-02-02
Update the mine after assigning roles

ereslibre:
When we assign roles to all machines, we need to explicitly call
mine.update, to ensure that all machines mine their information
depending on the role assigned. We can ensure that the orchestration
is going to have the required mined information.

### DIFF
--- a/lib/pharos/salt.rb
+++ b/lib/pharos/salt.rb
@@ -7,14 +7,14 @@ module Pharos
     include SaltApi
 
     # This method is the entrypoint of any Salt call. It will simply apply the
-    # given function 'fun' with the given arguments 'arg' to the given target
-    # machines.
+    # given function 'action' with the given arguments 'arg' to the given target
+    # machines provided by targets.
     #
     # Returns two values:
     #   - The response object.
     #   - A hash containing the parsed JSON response.
-    def self.call(tgt:, fun:, arg: nil)
-      hsh = { tgt: tgt, fun: fun, client: "local" }
+    def self.call(action:, targets: "*", arg: nil)
+      hsh = { tgt: targets, fun: action, client: "local" }
       hsh[:arg] = arg if arg
 
       res = perform_request(endpoint: "/", method: "post", data: hsh)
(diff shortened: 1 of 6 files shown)


lib/pharos/salt.rb
```diff
@@ -27,6 +27,11 @@ def self.minions
       JSON.parse(res.body)["return"].first
     end
 
+    # Call mine.update on salt for every minion specified by targets.
+    def self.mine_update(targets: "*")
```
mssola: Do we really need an extra method which is a simple alias of the `call` method ?

ereslibre: Do we want to `call` directly from the outside? I think it exposes too many salt information to users (the controller in this case). It would expose `tgt` and `fun`, I don't know if we want to expose that much.

mssola: Right. But I'd like to avoid adding a new method, which is simply an alias for `call`, for every salt function that we might end up calling. Is it really that different from a user perspective to say "I want to call mine_update", than "I want salt to call mine.update" ?

I'd be in favor of adding a function `all` which would be an alias of `call` for targets='*'. Then your call would be:

```ruby
Pharos::Salt.all("mine.update")
```

Another possible option would be to add a function like this:

```ruby
def self.perform(action, target = "*")
    # ...
end
```

This `action` parameter would expect a symbol `:mine_update` or any other allowed function, and it would do the translation from this symbol to the proper salt function. Then your call would've been:

```ruby
Pharos::Salt.perform(:mine_update)
```

And this would work for similar functions. That being said, this abstraction would not probably translate well whenever we want to add parameters 😅 What do you think ?

ereslibre: I think we can use `call` for this directly, and use other keyword arguments, maybe we could narrow the problem down to `Pharos::Salt.call(action: "mine.update")`. By default this method's signature is like: `def self.call(action:, targets: "*", arg: nil)`. It's really close to the `perform` solution that you provide, we don't need to map actions (you'd call the salt action directly), and it's somewhat better to be used from the outside.

ereslibre: Updated.

mssola:
LGTM 👍

ereslibre:
Do not merge, this is related to bug https://github.com/saltstack/salt/issues/11501, and we can fix this at salt level.

PR #15, opened 2017-02-02
Create assign_roles method on Minion and remove keyword args

jimmykarily:
Also trust our database and don't query salt for already applied roles

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,8 @@ unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
     gem "rubocop", "~> 0.46", require: false
     gem "brakeman", require: false
     gem "database_cleaner"
+    gem "pry"
+    gem "pry-nav"
   end
 
   group :test do
--- a/app/controllers/nodes_controller.rb
+++ b/app/controllers/nodes_controller.rb
@@ -22,17 +22,17 @@ class NodesController < ApplicationController
   # assigned role, assign a random role to it, and then call the salt
   # orchestration.
   def bootstrap
-    available_roles = [:master]
-    Minion.where(role: nil).find_each do |minion|
-      random_role = if available_roles.blank?
-        :minion
-      else
-        available_roles.delete_at(rand(available_roles.length))
-      end
-      minion.assign_role(role: random_role)
+    if Minion.where(role: nil).count > 1
+      Minion.assign_roles(roles: [:master], default_role: :minion)
+      Pharos::Salt.orchestrate
+
+      redirect_to nodes_path
+    else
+      redirect_back(
+        fallback_location: nodes_path,
+        alert:             "Not enough Workers to bootstrap. Please start at least one worker."
+      )
     end
-    Pharos::Salt.orchestrate
-    redirect_to nodes_path
   end
 
   # TODO
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -4,18 +4,43 @@ require "pharos/salt_minion"
 
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+  class NotEnoughMinions < StandardError; end
+  class CouldNotAssignRole < StandardError; end
+
   enum role: [:master, :minion]
 
   validates :hostname, uniqueness: true
 
+  # This method is used to assign the specified roles to Minions which do not
+  # have a roles already assigned (available minions).
+  # roles param are the roles we absolutely have to assign. If we can't assign
+  # one of those, the method will raise.
+  # default_role param can be set if we want all the rest of the available
+  # minions to get a default role.
+  def self.assign_roles(roles: [], default_role: nil)
+    minions = Minion.where(role: nil)
+
+    # only the needed number of minions or all if we have a default role
+    minions = minions.limit(roles.size) unless default_role
+
+    raise NotEnoughMinions if minions.count < roles.size
+
+    minions.find_each do |minion|
+      unless minion.assign_role(roles.pop || default_role)
+        raise CouldNotAssignRole
+      end
+    end
+  end
+
   # rubocop:disable SkipsModelValidations
   # Assigns a role to this minion locally in the database, and send that role
   # to salt subsystem.
-  def assign_role(role:)
-    return false if salt.roles?
+  def assign_role(new_role)
+    return false if role.present?
+
     Minion.transaction do
-      update_column :role, role
-      salt.assign_role role: role
+      update_column :role, new_role
+      salt.assign_role new_role
     end
     true
   rescue Pharos::SaltApi::SaltConnectionException
--- a/lib/pharos/salt_minion.rb
+++ b/lib/pharos/salt_minion.rb
@@ -30,8 +30,9 @@ module Pharos
     end
 
     # Assign role to this minion.
-    def assign_role(role:)
+    def assign_role(role)
       append_grain key: "roles", val: ROLES_MAP[role]
+
       role
     end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -60,6 +60,7 @@ GEM
     cliver (0.3.2)
     codeclimate-test-reporter (1.0.5)
       simplecov
+    coderay (1.1.1)
     concurrent-ruby (1.0.4)
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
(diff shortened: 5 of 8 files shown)


app/controllers/nodes_controller.rb
```diff
@@ -22,17 +22,17 @@ def show
...
-        available_roles.delete_at(rand(available_roles.length))
-      end
-      minion.assign_role(role: random_role)
+    if Minion.where(role: nil).count > 1
+      Minion.assign_roles(roles: [:master], default_role: :minion)
+      Pharos::Salt.orchestrate
+
+      redirect_to nodes_path
+    else
+      redirect_back(
```
ereslibre: Shouldn't we just redirect it to the `nodes_path` ?

lib/pharos/salt_minion.rb
```diff
@@ -30,9 +30,9 @@ def roles?
     end
 
     # Assign role to this minion.
-    def assign_role(role:)
-      append_grain key: "roles", val: ROLES_MAP[role]
-      role
+    def assign_role(new_role)
```
ereslibre: Here I'd just call the parameter `role`. This semantics are a bit different than the one in the `Minion` class, where it has a `role` and we are assigning a `new_role`, here we could just keep this as `def assign_role(role)`.

Gemfile
```diff
@@ -51,6 +51,8 @@ unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
     gem "rubocop", "~> 0.46", require: false
     gem "brakeman", require: false
     gem "database_cleaner"
+    gem "pry"
+    gem "pry-nav"
```
ereslibre: Are those leftovers? We really want to have `pry` and `pry-nav`?

app/models/minion.rb
```diff
@@ -4,18 +4,43 @@
...
+    minions = Minion.where(role: nil)
+
+    # only the needed number of minions or all if we have a default role
+    minions = minions.limit(roles.size) unless default_role
+
+    raise NotEnoughMinions if minions.count < roles.size
+
+    minions.find_each do |minion|
+      unless minion.assign_role(roles.pop || default_role)
+        raise CouldNotAssignRole
```
ereslibre: If one minion fails to assign the role, we stop the loop because we raise the exception. Maybe we could do something different here, maybe a more friendly failure, where we try to do our best to assign roles even if some minion fails to be assigned the role.

app/models/minion.rb
```diff
@@ -4,18 +4,43 @@
...
-  def assign_role(role:)
-    return false if salt.roles?
+  def assign_role(new_role)
+    return false if role.present?
+
     Minion.transaction do
-      update_column :role, role
-      salt.assign_role role: role
+      update_column :role, new_role
+      salt.assign_role(new_role)
```
ereslibre: I'd add the parentheses to `update_column` or remove them from `assign_role`.

app/controllers/nodes_controller.rb
jimmykarily: well, that's "back". I just don't want to hardcode it in case we add this button elsewhere. I will change the default though from root_path to node_path.

lib/pharos/salt_minion.rb
jimmykarily: you are right. Fixing.

Gemfile
jimmykarily: I asked as couple of days ago and I think @mssola agreed. I use the regularly and I have to remember to skip them when indexing changes in git. They live only in :development and :test so it shouldn't be much of a problem. I just have to push the docker image for dev env.

app/models/minion.rb
jimmykarily: I couldn't decide if we want to be strict or flexible here so I thought we should better raise for now. It might reveal underlying problems that would go unnoticed otherwise. If we prefer a different strategy in the future we can change this.

app/models/minion.rb
jimmykarily: removed

Gemfile
mssola: I don't personally use it, but if you find it useful, go ahead. I don't have any strong opinions on this.

app/controllers/nodes_controller.rb
ereslibre: Ok

Gemfile
ereslibre: Ok, fine with me.

app/models/minion.rb
ereslibre: Ok, we will be able to change this in the future.

app/controllers/nodes_controller.rb
mssola: I'd honestly just redirect to `nodes_path`. If we change the location of the button, there's other parts that will change anyways.

app/models/minion.rb
```diff
@@ -4,18 +4,43 @@
 
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
+  class NotEnoughMinions < StandardError; end
+  class CouldNotAssignRole < StandardError; end
```
mssola: How are these exceptions handled by controllers ? IIRC, they will be handled as an 500, no ? If that's the case, could you change the application_controller so these exceptions are simple 404 responses (they are not really a server failure).

mssola: Plus, it would be cool to add documentation on which cases we expect this exceptions to be raised.

app/models/minion.rb
```diff
@@ -4,18 +4,43 @@
...
+  # have a roles already assigned (available minions).
+  # roles param are the roles we absolutely have to assign. If we can't assign
+  # one of those, the method will raise.
+  # default_role param can be set if we want all the rest of the available
+  # minions to get a default role.
+  def self.assign_roles(roles: [], default_role: nil)
+    minions = Minion.where(role: nil)
+
+    # only the needed number of minions or all if we have a default role
+    minions = minions.limit(roles.size) unless default_role
```
mssola: Dumb question: won't this produce a second SQL query ? If so, could you simply chain this method call with the previous one for this case ?

app/models/minion.rb
mssola: I'd appreciate a more fine-grained error. Try to assign as many roles as possible, and return the minions that failed (or raise an exception holding such information so it can later be consumed by the controller, which will be able to alert to user on the UI).

app/models/minion.rb
```diff
@@ -4,18 +4,43 @@
...
+      end
+    end
+  end
+
   # rubocop:disable SkipsModelValidations
   # Assigns a role to this minion locally in the database, and send that role
   # to salt subsystem.
-  def assign_role(role:)
-    return false if salt.roles?
+  def assign_role(new_role)
```
mssola: Why ?

lib/pharos/salt_minion.rb
```diff
@@ -30,8 +30,9 @@ def roles?
     end
 
     # Assign role to this minion.
-    def assign_role(role:)
+    def assign_role(role)
```
mssola: Same as before. At a first glance it looks like a good idea to have this as a keyword argument (let's be consistent, we are using keyword arguments in lots of places already).

lib/pharos/salt_minion.rb
```diff
@@ -30,8 +30,9 @@ def roles?
...
       append_grain key: "roles", val: ROLES_MAP[role]
+
```
mssola: Empty line :P

lib/pharos/salt_minion.rb
mssola: Ah, forget it. I've reconsider it :)

app/models/minion.rb
mssola: Forget it, I've reconsider it :)

app/models/minion.rb
jimmykarily: An Exception should only be raised in exceptional cases and as such, we shouldn't know when that happens (otherwise we should handle that case too). For example, our code should never call assign_roles with more roles that available minions. If that happens then something is wrong in the code. Also we should always be able to assign a role on salt or we should handle the edge cases. That's why I prefer to let it fail for now. See my [comment](https://github.com/SUSE/pharos/pull/15#discussion_r99125280) on @ereslibre's review too.

app/models/minion.rb
jimmykarily: This line produces the first query. Line 21 above doesn't produce a query, it only returns a Relation. I use the same relation in line 28 above so I only want to add ```limit(roles.size)``` in the chain if we don't have default_role.

app/models/minion.rb
mssola: Got it

app/models/minion.rb
mssola: Got it

PR #17, opened 2017-02-03
Handle Highstate completion

jimmykarily:


db/migrate/20170201155426_add_highstate_to_minions.rb
```diff
@@ -0,0 +1,5 @@
+class AddHighstateToMinions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :minions, :highstate, :boolean
```
ereslibre: Maybe add an `after: :role`, so we keep `created_at` and `updated_at` the last ones.

ereslibre:
Just tried and it worked fine here, we might want to improve a little bit this PR.

app/models/minion.rb
```diff
@@ -10,6 +10,13 @@ class NotEnoughMinions < StandardError; end
   # Raised when we fail to assign a role on a minion
   class CouldNotAssignRole < StandardError; end
 
+  # Human readable names for highstate attribute values
+  HIGHSTATES = {
+    pending:  nil,
+    complete: true,
+    failed:   false
+  }.freeze
```
mssola: Wouldn't it be simpler to just use an enum for this ?

app/models/salt_event.rb
```diff
@@ -10,6 +10,11 @@ class SaltEvent < ApplicationRecord
   PROCESS_TIMEOUT_MIN = 5
   NO_EVENTS_TIMEOUT_SEC = 5
 
+  ENABLED_EVENT_HANDLERS = [
+    SaltHandler::MinionStart,
+    SaltHandler::MinionHighstate
+  ].freeze
```
mssola: Same here

app/models/salt_event.rb
```diff
@@ -57,17 +63,20 @@ def self.process_next_event(worker_id:)
...
     return @handler if @handler
 
-    case tag
-    when "minion_start"
-      @handler = SaltHandler::MinionStart.new(self)
+    klass = ENABLED_EVENT_HANDLERS.find do |handler_class|
+      handler_class.can_handle_event?(self)
     end
+
+    klass&.new(self)
```
mssola: Won't work for ruby < 2.3

db/migrate/20170201155426_add_highstate_to_minions.rb
jimmykarily: :D ok will do

app/models/minion.rb
jimmykarily: What do you mean? I was thinking of making this column and integer in the database and keep a mapping hash here (like in the code above).

app/models/salt_event.rb
jimmykarily: But the alternative makes rubocop angry. I will fix it as soon as we downgrade.

app/models/salt_event.rb
jimmykarily: I don't think I can do that here. There is no database column involved. This is a list of classes which we ask wether they are capable of handling a particular event or not. They should all implement the ```can_handle_event?``` method.

app/models/minion.rb
mssola: I mean something like this: https://github.com/SUSE/pharos/pull/17/files#diff-fa4c43e571bbfe7bd1ae914e5794a39fL13. Then the column can be an integer, and you map with this enum, which is cheaper.

jimmykarily: ok , got it. Fixing.

app/models/salt_event.rb
mssola: Yeah, you are right

PR #18, opened 2017-02-03
Update the salt interface to take action and a default targets: '*'

ereslibre:
This exposes less salt slang to the Pharos::Salt users.

### DIFF
--- a/lib/pharos/salt.rb
+++ b/lib/pharos/salt.rb
@@ -7,14 +7,13 @@ module Pharos
     include SaltApi
 
     # This method is the entrypoint of any Salt call. It will simply apply the
-    # given function 'fun' with the given arguments 'arg' to the given target
-    # machines.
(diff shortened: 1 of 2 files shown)


lib/pharos/salt.rb
```diff
@@ -7,14 +7,13 @@ class Salt
...
-    # machines.
+    # given function 'action' with the given arguments 'arg' to the given targets.
     #
     # Returns two values:
     #   - The response object.
     #   - A hash containing the parsed JSON response.
-    def self.call(tgt:, fun:, arg: nil)
-      hsh = { tgt: tgt, fun: fun, client: "local" }
+    def self.call(action:, targets: "*", arg: nil)
+      hsh = { tgt: targets, fun: action, client: "local" }
```
jimmykarily: Maybe we can add some documentation about what each parameter does. An usage example would also be nice.

PR #20, opened 2017-02-07
setup: check for kubelet before trying to execute it

MaximilianMeister:
otherwise the setup-environment process will be spawned into the
background and continues to run and spam the teminal even when the
start script failed


ereslibre:
LGTM :+1:

PR #21, opened 2017-02-09
downgrade to rails 4.2.7.1

MaximilianMeister:


MaximilianMeister:
I've already prepared the bump to `4.2.7.1` but we're still waiting for some input on the logistics of the rails update for SLE12, so I'll push the change (probably tomorrow) once we have some clarity there.

ereslibre:
LGTM :+1:

.rubocop.yml
```diff
@@ -29,6 +29,10 @@ Bundler/OrderedGems:
 RSpec/DescribeClass:
   Enabled: false
 
+# https://github.com/bbatsov/rubocop/issues/3629
```
mssola: Could you also add a "NOTE: [Rails 5]" stating that this is no longer needed there ?

.travis.yml
```diff
@@ -1,9 +1,7 @@
 language: ruby
 rvm:
-  - 2.3.3
+  - 2.1.2
```
mssola: Copy what we have in Portus. We are testing more versions there. And don't remove ruby-head

Gemfile
```diff
@@ -1,19 +1,16 @@
-# frozen_string_literal: true
```
mssola: I don't think removing this is needed.

mssola: And what about nodejs. Is there a way to pin its version here ? Or that's something RPM-specific ?

MaximilianMeister: > And what about nodejs. Is there a way to pin its version here ? Or that's something RPM-specific ?

i think that's only rpm specific where we precompile the assets

MaximilianMeister:
@mssola updated, thanks

mssola:
LGTM

mssola:
@jimmykarily what do you think ?

kubernetes/manifest-templates/pharos.yaml
```diff
@@ -8,7 +8,7 @@ spec:
   hostNetwork: true
   containers:
   - name: pharos-dashboard
-    image: ereslibre/pharos:development
+    image: mmeister/pharos_dev:latest
```
jimmykarily: Maybe we should create a "kubic" repo or something like that.

MaximilianMeister: agree, this should be only temporary until we have some repo. maybe we could use the opensuse namespace

jordimassaguerpla:
lgtm

jimmykarily:
LGTM

kubernetes/manifest-templates/pharos.yaml
ereslibre: The idea behind the `development` tag was to avoid the kubelet try to find new versions of the image everytime (this affects on networks with high latencies and low bandwidth). The kubelet will try to fetch a new image version if the tag is `latest`.

jimmykarily: You are right. Let's use the "development" tag.

PR #26, opened 2017-02-14
Gemfile: bump mysql to latest version

MaximilianMeister:
we are not so tied anymore to what we have in SLE12, so we can
bump the version up again


ereslibre:
Great! LGTM :+1:

PR #28, opened 2017-02-15
Add an etcd instance to the development environment.

ereslibre:
When there is no dashboard, the development environment is the
dashboard itself. Add an etcd instance that the rest of the cluster
can use.

### DIFF
--- a/lib/velum/salt_minion.rb
+++ b/lib/velum/salt_minion.rb
@@ -9,7 +9,7 @@ module Velum
     attr_accessor :minion_id
 
     ROLES_MAP = {
-      master: ["kube-master", "etcd"],
+      master: ["kube-master"],
       minion: ["kube-minion"]
     }.freeze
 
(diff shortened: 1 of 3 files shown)


ereslibre:
This kind of depends on https://github.com/openSUSE/docker-containers/pull/40. We can merge it with the current image location, but it would be great if we could directly use opensuse/etcd.

lib/velum/salt_minion.rb
```diff
@@ -9,7 +9,7 @@ class SaltMinion
     attr_accessor :minion_id
 
     ROLES_MAP = {
-      master: ["kube-master", "etcd"],
+      master: ["kube-master"],
```
jimmykarily: Right now all minions will be etcd-proxies but in the future we might need to revisit this one. Looks good for now.

jimmykarily:
I haven't tried the orchestration but if you did and it works I have no further comments. Merge at your own risk :p

lib/velum/salt_minion.rb
ereslibre: Yes, we will improve how and to whom we assign roles in the very near future.

PR #30, opened 2017-02-16
Use --non-interactive flag for non interactive mode in dev env start script

jimmykarily:
We need the non interactive mode for the script to be used from other
scripts (e.g. in the e2e-tests project)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,13 @@ within the `kubernetes` folder. You will need a [standalone kubelet](https://kub
 
 This will start the velum container as well as other required containers.
 
+You can use the `--non-interactive` flag if you prefer to run in non interactive mode.
+In that case you will be asked before old containers are removed.
+
+E.g.
+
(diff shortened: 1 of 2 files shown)


ereslibre:
I would do the other way around. It's easier to add "--non-interactive" to our e2e-tests that to ask the user for a new flag. However, LGTM, I always clean up anyways :)

jimmykarily:
getopts is [limited to single character flags](http://stackoverflow.com/questions/402377/using-getopts-in-bash-shell-script-to-get-long-and-short-command-line-options). I will use `-r` as in `*r*eplace containers`.

jimmykarily:
Used `--non-interactive` after all

mssola:
@jimmykarily @ereslibre yeah, it's what an opensuse guy would expect 😉

ereslibre:
Anyone interested in this project would be the user. We are users too,
actually 🤗

PR #31, opened 2017-02-16
Draft UI Implementation

santiph:
Draft UI Implementation

screen shot 2017-02-16 at 16 52 46
screen shot 2017-02-16 at 16 53 18
screen shot 2017-02-16 at 16 53 35


MaximilianMeister:
@santiph thanks! could you still remove the commits from @ereslibre about the vagrantfile and squash your work into one commit?

also would be great if you could upload a screenshot

app/controllers/config_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# ConfigController shows the main page.
+class ConfigController < ApplicationController
+  def index
```
ereslibre: If they are blank, you can skip writing them at all, same with the others. If you only add the view it should work.

app/helpers/application_helper.rb
```diff
@@ -7,4 +7,27 @@ def user_image_tag(owner)
...
+      {id: 5,  name: 'Stuart',  ip: '220.68.155.90', network: 'Asia',    master: false, status: 'warning',  new: false, isSelected: false},
+      {id: 6,  name: 'Dave',    ip: '192.168.15.55', network: 'Europe',  master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 7,  name: 'Steve',   ip: '192.168.15.20', network: 'America', master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 8,  name: 'Donny',   ip: '192.168.15.20', network: 'America', master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 9,  name: 'Ken',     ip: '220.68.155.93', network: 'Asia',    master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 10, name: 'Mike',    ip: '220.68.155.94', network: 'Asia',    master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 11, name: 'Paul',    ip: '220.68.155.96', network: 'Asia',    master: false, status: 'ready',    new: true,  isSelected: false},
+      {id: 12, name: 'Larry',   ip: '220.68.155.97', network: 'Asia',    master: false, status: 'ready',    new: true,  isSelected: false},
+    ]
+  end
```
ereslibre: It's OK for now, we can refactor this a bit after merging it.

app/helpers/application_helper.rb
```diff
@@ -7,4 +7,27 @@ def user_image_tag(owner)
...
+
+  def title_helper(content)
```
ereslibre: Since this is already a helper we can remove the `_helper` ending, just naming it `title`, because we are going to use it from the view. However it's maybe even better to just use `h1` on the view if there is no more logic here, because we aren't adding too much value here. It's just longer, and doesn't do anything else than `h1`.

app/helpers/dashboard_helper.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+# ApplicationHelper contains all the view helpers.
+module DashboardHelper
+
+  # Render Nodes Table row
+  def nodes_table_row(icon, name, ram, cpu, network, master)
```
ereslibre: It's OK for now. In rails there exists the notion of `partials` that we can use here, instead of using the helpers for doing some composition. The idea behind helpers is that they implement logic that would be harder to track and mantain on a view (or something several views can use). Here I'd go with a partial for implementing each row, or even on the same view. We could add a helper for the icon, so the helper method would only include the `case` statement. The rest actually belongs more to a view than to a helper.

app/assets/javascripts/application.js
```diff
@@ -1,16 +0,0 @@
...
-// It's not advisable to add code directly here, but if you do, it'll appear at the bottom of the
-// compiled file. JavaScript code in this file should be added after the last require_* statement.
-//
-// Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
-// about supported directives.
-//
-//= require jquery
-//= require jquery_ujs
-//= require turbolinks
-//= require_tree .
```
jimmykarily: I would keep this file since it is just what Rails generates by default when you bootstrap a new project. Simply removing all the not-needed files from the app/assets/javascripts directory will generate an empy application.js file which is what you want for now I think.

jimmykarily:
[REMINDER] When @santiph is done with the front end, we should bring back the previous functionality (bootstrap cluster) using the new views, so please don't merge this yet.

app/assets/javascripts/application.js
jimmykarily: Also the application.js is referenced in the app/views/layouts/application.html.slim and it now creates a 404 request.

app/helpers/application_helper.rb
jimmykarily: I suggest we jump in after the initial review and fix things before merging. I would prefer our master to be in good shape. Since we don't have a "stable" branch yet, the master will be the first thing someone would try out.

app/views/layouts/application.html.slim
```diff
@@ -1,26 +1,29 @@
...
+    meta(charset='utf-8')
+    meta(http-equiv='X-UA-Compatible', content='IE=edge')
+    meta(name='viewport', content='width=device-width, initial-scale=1')
+    // The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags
+    meta(name='description', content='')
+    meta(name='author', content='')
+    link(rel='icon', href='/favicon.ico')
     title Velum
     = stylesheet_link_tag 'application', media: 'all'
     = javascript_include_tag 'application'
```
jimmykarily: this is a 404. See my other comment above (app/assets/javascripts/application.js)

config/routes.rb
```diff
@@ -16,6 +16,12 @@
     root to: "auth/sessions#new"
   end
 
+  get '/installer/options', to: 'installer#options'
+  get '/installer/master', to: 'installer#master'
+  get '/nodes/add', to: 'nodes#add'
+  get '/nodes', to: 'nodes#list'
+  get '/config', to: 'config#index'
```
jimmykarily: Are we using these as "example" pages? I'm not sure about what purpose each one serves.

jimmykarily:
In general this is a good cleanup, well done @santiph! When you are done with your changes, ping someone to start working on the (now broken) functionality.

config/routes.rb
santiph: `/installer/config` is intended to be the first page the user will see when Yast installation is over.
`/installer/master` is the following page in the flow, where the master is selected by the user. Once clicked on Next, the installation process with proceed and once completed, move to the next page.
`/nodes` is the list of available up-and-running nodes in the cluster, and the page the user would see once logged into the system, once it's up and running (honestly, I think this should be the real dashboard, with the tachometers and other control components... but is up to the product owner and designer)
`/nodes/add` is the page where new nodes will be displayed, selected and added to the existing cluster.
`/config` is the configuration page for velum where, among other things, the autoupdate checkbox should be displayed. Options from `/installer/config` could be displayed and modified (Federica?).

app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# InstallerController shows the main page.
+class InstallerController < ApplicationController
```
mssola: Could you provide more helpful documentation ? This class has the same documentation as `ConfigController`. What's the difference between the two of them ? Could you write this on both classes ?

app/assets/javascripts/application.js
santiph: Since no javascript is being used so far, I'd prefer to completely remove it and include it back once it's really needed

app/views/layouts/application.html.slim
santiph: Since no javascript is being used so far, I'd prefer to completely remove it and include it back once it's really needed

app/controllers/installer_controller.rb
mssola: Moreover, the following methods look like pages from the ConfigController. Maybe both classes can be merged ?

app/controllers/nodes_controller.rb
```diff
@@ -34,4 +34,13 @@ def bootstrap
 
   # TODO
   def destroy; end
+
```
mssola: remove all these empty lines

app/controllers/nodes_controller.rb
```diff
@@ -34,4 +34,13 @@ def bootstrap
...
+
+
+
+  def list
```
mssola: Again, please document what's these methods are all about. At first glance I have no idea what's going on.

config/routes.rb
mssola: Instead of `nodes#list`, please use `nodes#index`, which is the REST-way of handling this. For `nodes#add`, you should go for `nodes#new`. The installer paths can also be RESTified.

That being said, that's something we can fix after this review.

mssola:
@santiph thanks a bunch for all this 👏

config/routes.rb
santiph: @mssola `nodes#index` is already being used.
`nodes#add`fixed.

ereslibre:
Superseded by https://github.com/kubic-project/velum/pull/50.

PR #32, opened 2017-02-16
packaging: add a VERSION file

MaximilianMeister:
this is helpful during specfile creation, and it's done like that
in Portus as well


cyphar:
Note: When creating `rpm` versions, I've found that the best method for creating `VERSION` files is the following:

* Before doing a release, make the version `old_version+dev` -- `rpm` recognises this as meaning "newer than old version".
* After doing a release, make the version `new_version` for a single commit then `new_version+dev`.
* When doing `rc` releases use `new_version~rcN` which is recognised by `rpm` as meaning "older than `new_version`".

And of course, `new_version~rcN+dev` _is_ handled correctly.

jordimassaguerpla:
Using "-" confuses rpm, better follow cyphar scheme. I am creating a PR

PR #35, opened 2017-02-17
Add Vagrantfile

ereslibre:


santiph:
+1

mssola:
I wonder if it would be a good idea to create a new directory `deploy` there. In there we would have two subdirectories for now: `kubernetes` and `vagrant`. The `kubernetes` one would be the same as the one we have in the root directory, and the `vagrant` would contain the script that you are passing to vagrant inline. What do you think ?

ereslibre:
We can do that, the `Vagrantfile` is kind of hackish anyways, and I'd like to port it to opensuse with a newer image (leap at least). The only thing that I see by adding `deploy` and inside there `kubernetes` and `vagrant` is that the `kubernetes/start.sh` script is meant to be run manually, but `vagrant/deploy.sh` isn't, and I don't know if that can cause some kind of confusion. That's why I preferred to keep the `Vagrantfile` in one piece (because it's highly hackish right now). However if you see benefits in splitting it I am not against it.

mssola:
I see your point. Ok, let's do that in the future then 😉

PR #38, opened 2017-02-17
travis: removed the ugly travis_mariadb script

mssola:
This is provided as an addon from Travis, and it's more reliable than
what we had before.


ereslibre:
Great! LGTM :+1: Thank you for this!

PR #39, opened 2017-02-17
packaging: reference master instead of HEAD

MaximilianMeister:
when we call make_spec.sh within a concourse git resource, the git
command renders HEAD instead of master.


MaximilianMeister:
i was so free to merge this even w/o a +1, as it only affects the packaging, not the app itself. and this fixes the packaging

PR #40, opened 2017-02-18
packaging: add script to fetch velum buildresults

MaximilianMeister:


MaximilianMeister:
oops i still have to make small change, will send a followup...

PR #41, opened 2017-02-18
packaging: followup to exit with 1 when build failed

MaximilianMeister:
followup for #40


MaximilianMeister:
merging directly, as it only affects the packaging pipeline

PR #43, opened 2017-02-19
add api to let the user to choose the master

MaximilianMeister:


app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,42 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  # rubocop:disable Metrics/AbcSize
+  def assign
+    @minion = Minion.find(params[:node_id])
+
+    unless Minion.roles.keys.include?(params[:role])
```
MaximilianMeister: thought about that too, but at this point the instance of the model is already created, we're just querying with `find`

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,42 @@ def show
...
+      raise Minion::InvalidRole, "Could not assign role #{params[:role]}"
+    end
+
+    respond_to do |format|
+      if @minion.assign_role(params[:role])
+        res, body = Velum::Salt.orchestrate
```
MaximilianMeister: good question, i was assuming we have to do that to propagate the role to the nodes, but as you explained, this seems wrong. so shall I remove the orchestration bit again?

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,42 @@ def show
...
+        http_status = res.code == "200" ? :ok : :unprocessable_entity
+        flash[:error] = "Salt failed to orchestrate: #{body.inspect}" unless http_status == :ok
+
+        format.html { redirect_to nodes_path }
+        format.json { head http_status }
```
MaximilianMeister: at least i would. I think it always makes sense to have a json api for enabling integration with other clients, and it doesn't cost much to have it already now, even if we don't need it yet

config/routes.rb
```diff
@@ -17,6 +17,8 @@
   end
 
   resources :nodes, only: [:index, :show, :destroy] do
+    post :assign
```
MaximilianMeister: i tried to keep the api generic, but i am not 100% sure if that makes sense. do we want to assign e.g. a minion role explicitly?

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,42 @@ def show
...
+        format.html do
+          flash[:error] = @minion.errors.full_messages.first
+          redirect_to nodes_path
+        end
+        format.json { render json: @minion.errors, status: :unprocessable_entity }
+      end
+    end
+  rescue Velum::SaltApi::SaltConnectionException,
+         ActiveRecord::RecordNotFound,
+         Minion::InvalidRole => e
```
jimmykarily: We raised this exception just 20 lines above. I think we should better return early instead of raising.

config/routes.rb
ereslibre: Right now the selection the user performs is master selection. The rest of the nodes will be assigned minion/worker role. That's why I proposed to use `assign_master` (maybe mapped to `assign-master` route). In this action we could set the master role to the selected node, set the minion/worker to the rest of the nodes, and that's it. Depending on how the setup on the UI is we would call the orchestration at the end (after setting roles for all machines involved in the cluster), or don't call the orchestration at all (if that's another `Bootstrap now` step in the UI, that I don't know right now).

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,42 @@ def show
...
+      else
+        format.html do
+          flash[:error] = @minion.errors.full_messages.first
+          redirect_to nodes_path
+        end
+        format.json { render json: @minion.errors, status: :unprocessable_entity }
+      end
+    end
+  rescue Velum::SaltApi::SaltConnectionException,
+         ActiveRecord::RecordNotFound,
```
jimmykarily: This rescues a quite bit block of code. Inside that block there might be more than one places where this exception can occur and some of them we might not even know right now. If we want a beautiful error in case the user tampers with the params maybe it would be better to explicitly check for that case at line 23 and let the rest of the code raise Exceptions.

app/controllers/nodes_controller.rb
MaximilianMeister: makes sense, but then i think we'd have to write another rendering block, which i wanted to avoid. and it would look almost the same as the one in the rescue section i believe. this raise and rescue is also an early return. is this somehow bad practice?

config/routes.rb
jimmykarily: I would keep it generic

spec/controllers/nodes_controller_spec.rb
```diff
@@ -116,4 +116,134 @@
...
+
+        it "assigns the master role" do
+          allow(salt).to receive(:orchestrate).and_return(net_httpok_instance, {})
+          allow(net_httpok_instance).to receive(:code).and_return("200")
+          allow_any_instance_of(Minion).to receive(:assign_role).with("master").and_return(true)
+          post :assign, node_id: Minion.find_by(hostname: "master").id, role: :master
+          expect(response.redirect_url).to eq "http://test.host/nodes"
+        end
+
+        # rubocop:disable RSpec/MessageChain
```
jimmykarily: I think (I'm not sure) that you have to enable it again when you are done or else it stays disabled.

config/routes.rb
ereslibre: Maybe `assign_roles` with path `assign-roles`? `assign` is too generic from my POV.

app/controllers/nodes_controller.rb
ereslibre: The orchestration should be called only once, after all roles have been set. It isn't called from within `assign_role` or `assign_roles`. The idea of the orchestration is to `bootstrap` everything for the first time, but for that to happen all roles need to have been set first.

spec/controllers/nodes_controller_spec.rb
ereslibre: Yes, you need to restore disabled cops, or they will stay disabled.

app/controllers/nodes_controller.rb
jimmykarily: [Most people](https://www.google.gr/search?q=exceptions+as+flow+control&ie=utf-8&oe=utf-8&client=firefox-b-ab&gfe_rd=cr&ei=Hc6qWOfDBKjA8ge7tbPYCA) seem to agree it is bad practice. And to be honest, I also think exceptions should only be raised when something exceptional happens. That is a case which we could not predict and thus we couldn't handle gracefully.

config/routes.rb
jimmykarily: ```assign_role``` maybe? We can only assign role to one minion at a time as it is now.

ereslibre: The path can be `assign-roles` because as arguments we get the `master` the user choose, but we are going to set the minion role for the rest of the discovered machines. So you are effectively assigning roles (plural). The master that was chosen, and the rest, that will be minions by default (at least for now).

app/controllers/nodes_controller.rb
MaximilianMeister: then why do we throw custom exceptions in the `Minion` model? in our case it's not a classical exception, it's a custom one, which doesn't even have "Exception" in the name. it's something known, and i just followed what we already do in the model

app/controllers/nodes_controller.rb
MaximilianMeister: i have tried to change that, but rubocop gave me a `Metrics/CyclomaticComplexity` violation. I think it's because of the additional `respond_to` block which i wanted to avoid, as i have stated below.

```ruby
  def assign
    if Minion.exists?(id: params[:node_id])
      @minion = Minion.find(params[:node_id])
    else
      error_msg = "Cannot assign role to non-existing Minion with id: #{params[:node_id]}"
      respond_to do |format|
        format.html do
          flash[:error] = error_msg
          redirect_to nodes_path && return
        end
        format.json { render json: error_msg, status: :unprocessable_entity && return }
      end
    end
    ...
  end
```

I can try to reduce the complexity further, by saving some errors in a hash or variables, and only have one intelligent `respond_to` block but is it really worth adding this complexity?

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,61 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def assign_roles
+    minion = Minion.find(params[:master_id])
```
MaximilianMeister: still not sure if we should send an id, or the hostname as a parameter

MaximilianMeister:
@ereslibre @jimmykarily I have kept the changes now as separate commits, for you to see. but i can squash them again, once you're fine with it

app/controllers/nodes_controller.rb
jimmykarily: I think any of them is fine.

app/controllers/nodes_controller.rb
```diff
@@ -18,6 +18,61 @@ def show
...
+    minion_errors = []
+
+    if minion.assign_role(:master)
```
jimmykarily: Now that we have other requirements, the assign_roles method in app/models/minion.rb is not that useful. I would try to refactor that method to accept a Hash as the `roles` argument so it can handle this case too.

E.g.

```
  Minion.assign_roles({master: ["master.example.com"], minion: ["must_be_minion.example.com"]}, default_role: :minion)
```

This way we keep the controller thin and that method becomes useful again.

app/models/minion.rb
```diff
@@ -9,37 +9,69 @@ class Minion < ApplicationRecord
...
+    if roles[:master].any? && Minion.find_by(hostname: roles[:master].first).nil?
+      raise NonExistingMinion, "Failed to process non existing minion: #{roles[:master].first}"
+    end
+    master = Minion.find_by(hostname: roles[:master].first)
+    # choose requested minions or all other than master
+    minions = if roles[:minion]
+      Minion.where(hostname: roles[:minion])
+    else
+      Minion.where(role: nil).where.not(hostname: roles[:master].first)
+    end
```
jimmykarily: Can we take a more generic approach? As in, not check each roles separately but iterate over the ```roles```. E.g.

```
assigned_ids = []
roles.each do |role, minion_ids|
  minions = Minion.where(hostname: minion_ids).each do |minion|
    if minion.assign_role(role)
      assigned_ids << minion.id
    else
      raise CouldNotAssignRole, "Failed to assign #{role} role to #{minion.hostname}"
    end
  end
end
```
(Warning: not tested code)

app/controllers/nodes_controller.rb
```diff
@@ -18,12 +18,31 @@ def show
...
+  end
+
   # Bootstraps the cluster. This method will search for minions missing an
   # assigned role, assign a random role to it, and then call the salt
   # orchestration.
   def bootstrap
     if Minion.where(role: nil).count > 1
-      Minion.assign_roles(roles: [:master], default_role: :minion)
+      # choose first minion to be the master
+      Minion.assign_roles(roles: { master: [Minion.all.first.hostname] })
```
ereslibre: `Minion.first.hostname` will do it :)

app/models/minion.rb
```diff
@@ -9,37 +9,69 @@ class Minion < ApplicationRecord
...
-      unless minion.assign_role(roles.pop || default_role)
-        raise CouldNotAssignRole
+      unless minion.assign_role(:minion)
+        raise CouldNotAssignRole, "Failed to assign minion role to #{minion.hostname}"
+      end
+      assigned_ids << minion.id
+    end
+
+    # assign default role if there is any minion left with no role
+    Minion.where(role: nil).find_each do |minion|
```
jimmykarily: I think you want to run this only if default_role is not nil. We need to have the option to pass a nil default_option. In that case we won't set a default role to all the remaining minions.

app/models/minion.rb
```diff
@@ -9,37 +9,69 @@ class Minion < ApplicationRecord
...
+  #       master: ["master.example.com"],
+  #       minion: ["minion1.example.com"]
+  #     },
+  #     default_role: :dns
+  #   )
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def self.assign_roles(roles: {}, default_role: :minion)
+    if roles[:master].any? && Minion.find_by(hostname: roles[:master].first).nil?
```
ereslibre: Since we accept an `Array`, maybe we can do a best effort try and say something like `if roles[:master].any? && !Minion.exists?(hostname: roles[:master])`, so we ensure that we have at least one existing `master` assigned. With this logic we accept an `Array` but only take into account the first one into that given array.

app/models/minion.rb
ereslibre: I think the entire if can be reduced to: `minions = Minion.where(role: roles[:minion]).where.not(hostname: roles[:master])`

config/routes.rb
```diff
@@ -19,6 +19,7 @@
   resources :nodes, only: [:index, :show, :destroy] do
     collection do
       post :bootstrap
+      post :assign_roles
```
ereslibre: On the route it's strange to see underscores, if I recall correctly this can be done using `post :assign_roles, path: 'assign-roles'`.

lib/velum/salt_api.rb
```diff
@@ -95,10 +95,8 @@ def perform_request(endpoint:, method:, data: {}, client: nil)
 
         opts = { use_ssl: false, open_timeout: 2 }
         Net::HTTP.start(uri.hostname, uri.port, opts) { |http| http.request(req) }
-      rescue *HTTP_EXCEPTIONS
-        # TODO: we might want to wrap the original exception inside the SaltConnectionException
-        # so we can show detailed problems to the user in the future.
-        raise SaltConnectionException
+      rescue *HTTP_EXCEPTIONS => e
+        raise SaltConnectionException, e
```
ereslibre: Why are we raising with `e` here?

config/routes.rb
mssola: This is not very *Resty*. Would it make sense to create a new controllers `RolesController` and then add this functionality as its `update` method ?

app/models/minion.rb
MaximilianMeister: makes sense somehow, but then we'd have to always pass the minion hostnames, don't we? we also then need to have another `each` block within `roles.each` for the minions. we might save some lines of code, but wouldn't that increase the complexity too much?

app/models/minion.rb
MaximilianMeister: good point, i have changed the default to `:minion` but it can always be overwritten

app/controllers/nodes_controller.rb
```diff
@@ -18,12 +18,31 @@ def show
     @minion = Minion.find(params[:id])
   end
 
+  def assign_roles
+    Minion.assign_roles(roles: { master: [params[:hostname]] })
```
mssola: Use strong params.

config/routes.rb
MaximilianMeister: you mean undescores in the url endpoint? not sure what is best here TBH, but we also have `/users/sign_in`, `/users/sign_out`

lib/velum/salt_api.rb
MaximilianMeister: to wrap the exception details into `SaltConnectionException`, for not getting lost as they will bubble up in the json response or flash message i think

app/controllers/nodes_controller.rb
```diff
@@ -18,12 +18,31 @@ def show
...
+
+    respond_to do |format|
+      format.html { redirect_to nodes_path }
+      format.json { head :ok }
+    end
+  rescue Minion::CouldNotAssignRole,
```
mssola: You shouldn't be using exceptions for code flow. There are a couple of issues here:

1. If `assign_roles` can raise an error and it performs an bang action, then it should be named with a bang `assign_roles!`. Otherwise, as a reader I have no way of knowing this (it's kind of a Rails convention 😉).
2. In my head, exceptions should really occur only in "exceptional" cases (well, I'm against exceptions in general, but that's a different story 😁). Not being able to assign a role or that the user wanted to change a non-existing minion is not that exceptional:
    - If the minion never existed, then this should be a 404. You could handle that by adding to this class (or the application_controller): `rescue_from Minion::NonExistingMinion, with: :not_found`. This way, you don't have to handle this explicitely, Rails and the frontend will already guess what a 404 stands for, and other methods that might face this situation, can act in the same way.
    - I'd actually expect the assign_roles method to return something meaningful (e.g. it can return whether it was successful or not, like Rails' `save` method). In this case, a good ol' if/else statement would fix this (note also that there's some shared code between these two cases: in html the only difference is in the flashy message).

app/models/minion.rb
```diff
@@ -9,37 +9,69 @@ class Minion < ApplicationRecord
...
+    minions = if roles[:minion]
+      Minion.where(hostname: roles[:minion])
+    else
+      Minion.where(role: nil).where.not(hostname: roles[:master].first)
+    end
 
+    # assign master if requested
     assigned_ids = []
+    if master
+      # rubocop:disable Style/GuardClause
```
mssola: Rubocop's complain is completely valid. This should be written like this:

```ruby
raise CouldNotAssignRole, "Failed to assign master role to #{master.hostname}" unless master.assign_role(:master)
assigned_ids << master.id
```

mssola: That being said, and as I said earlier, I don't see the point of raising an exception here.

app/models/minion.rb
```diff
@@ -9,37 +9,69 @@ class Minion < ApplicationRecord
...
+  #   Minion.assign_roles(
+  #     roles: {
+  #       master: ["master.example.com"],
+  #       minion: ["minion1.example.com"]
+  #     },
+  #     default_role: :dns
+  #   )
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
```
mssola: If rubocop is complaining that this method is complex, it's because it's true 😄 There should be a way to simplify this by splitting this into logical units. If you think that this is beyond this PR (which is fair), I'd be fine if you could write a TODO note.

app/models/minion.rb
jimmykarily: The point is to not hardcode the roles. This method shouldn't care what the roles are. Its responsibility is to assign the specified roles to the specified minions whatever they may be. If we want to assign a new role in the future we shouldn't have to change this method.

app/models/minion.rb
mssola: For example, this method is used like: "I want this role to be applied for this node, and I don't care about the rest (take a default)". Why not do this simpler:

- A method that assigns "expected" roles (I think that you don't even need to create it 😋)
- A method that sweeps all the marked nodes and effectively does the assignment in the kubernetes/salt context.

This two-step mechanism would give you simpler methods, while giving you the same functionality. Other questions like "finding a non-existing" minion would be handled by the callee, and if you want it can return an exception that would be handled as a mere 404 as I said before 😉

spec/controllers/nodes_controller_spec.rb
```diff
@@ -116,4 +116,110 @@
       end
     end
   end
+
+  # rubocop:disable RSpec/AnyInstance
+  # rubocop:disable RSpec/ExampleLength
+  # rubocop:disable RSpec/NestedGroups
```
mssola: Some of these are easily avoidable. For example, NestedGroups can be avoided iirc by using `describe` instead of `context` all the time. I think that this is how the `context` method was meant to be used.

spec/models/minion_spec.rb
```diff
@@ -4,29 +4,61 @@
 describe Minion do
   it { is_expected.to validate_uniqueness_of(:hostname) }
 
+  # rubocop:disable RSpec/ExampleLength
```
mssola: The fact that you have to disable this is already telling you that this method is too complex 😉

app/models/minion.rb
MaximilianMeister: i see that this is obvious, but then it complains that the line is too long...

spec/models/minion_spec.rb
MaximilianMeister: which method? afaics it means that there are too many examples, but what is the problem with that?

config/routes.rb
MaximilianMeister: good question, i'd maybe discuss that first and do a follow up, as that requires some architectural planning like what should the roles controller do beyond that etc.

ereslibre: You are right :( They are not very fancy. Anyways...

lib/velum/salt_api.rb
ereslibre: It's OK. I was thinking something more in the lines of `raise SaltConnectionException.new(e)` and keep the original exception as an attribute of `SaltConnectionException`, but we can completely ignore this for now.

config/routes.rb
ereslibre: Other option is to use `PUT/PATCH` on `/nodes` endpoint. With a payload of the type:
```
{
  "hostname1": { roles: ["kubernetes-master"] },
  "hostname2": { roles: ["kubernetes-minion"] }
}
```

Or maybe `ID's` instead of `hostnames`.

app/controllers/nodes_controller.rb
ereslibre: In this case there's no security issue (we are explicitly accessing :hostname, but it's a good practice).

app/controllers/nodes_controller.rb
```diff
@@ -5,6 +5,16 @@
 # NodesController is responsible for everything related to nodes: showing
 # information on nodes, deleting them, etc.
 class NodesController < ApplicationController
+  rescue_from Minion::NonExistingMinion do |e|
```
mssola: Nitpick: could you change this with:

```ruby
rescue_from Minion::NonExistingMinion, :minion_not_found
```

And then define a private method `minion_not_found` that does what you wrote ? I think that writing the whole thing in a lambda is confusing in a controller.

app/controllers/nodes_controller.rb
```diff
@@ -5,6 +5,16 @@
...
+    respond_to do |format|
+      format.html do
+        flash[:error] = e.message
+        redirect_to nodes_path
+      end
+      format.json { render json: e.message, status: :unprocessable_entity }
```
mssola: I'd say that the status should be 404 instead, since the problem is that the requester referenced a resource that could not be found.

config/routes.rb
```diff
@@ -18,6 +18,7 @@
 
   resources :nodes, only: [:index, :show, :destroy] do
     collection do
+      put :update
```
mssola: It should be added in the array two lines above instead:

```ruby
resources :nodes, only: [:index, :show, :update, :destroy] do
```

MaximilianMeister: for now it's not a single resource that we're patching, remember we're sending a payload with multiple node assignments, although it could be done like that too, but routing it then to the `assign_role` instance method. but maybe that's sth for later

mssola: Oh, you are right

PR #44, opened 2017-02-20
packaging: make sure build is running, obs is slow

MaximilianMeister:
as obs is a bit slow picking up a new build once a new change has
been submitted, we need an intermediate step to make sure the build
has begun, before checking for success/failure


ereslibre:
LGTM :+1:

PR #48, opened 2017-02-22
packaging: adapt parameters for reusability and add copypac script

MaximilianMeister:


MaximilianMeister:
i just realized i need to enable passing the api url as well in `wait_for_build.sh`

packaging/suse/concourse/buildresult.sh
```diff
@@ -2,24 +2,31 @@
 
 set -e
 
-if [ -z "$1" ] || [ -z "$2" ]; then
+if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ] || [ -z "$4" ]; then
```
jimmykarily: I think this is equivalent to this ```if [ -z "$4" ]; then```
I mean your check is not "if any of the 1,2,3,4 arguments is empty" which is the same as saying "if the argument 4 is empty" since both conditions return true if there is a 4th argument and both return false if there isn't.

Maybe I'm missing something?

MaximilianMeister: no you're right, we don't have to be so verbose here :+1:

MaximilianMeister:
merging to trigger the new pipeline

PR #49, opened 2017-02-23
packaging: fix error in order of execution

MaximilianMeister:


MaximilianMeister:
merged directly as this was an obvious fix

PR #51, opened 2017-02-23
packaging: enhancements and fixes

MaximilianMeister:
* make sure to write .oscrc
* enhance logging
* cache result after success
  otherwise we will always fail, due to an outdated result
  if the apiurl is different than https://api.opensuse.org


jordimassaguerpla:
lgtm

PR #53, opened 2017-02-24
Set dashboard location on the pillar.

ereslibre:
When starting the development environment, we can take the vars.sls
from the pillar, and mount it on the salt-master, so we can add some
variables without having to modify the original salt states.

### DIFF
--- a/kubernetes/start
+++ b/kubernetes/start
@@ -34,24 +34,31 @@ done
 default_interface=$(awk '$2 == 00000000 { print $1 }' /proc/net/route)
 ip_address=$(ip addr show $default_interface | awk '$1 == "inet" {print $2}' | cut -f1 -d/)
 
+# Use SALT_DIR is set and not empty else use our default directory
+# This can be used to point to different state files for orchestration.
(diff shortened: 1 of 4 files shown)


kubernetes/start
```diff
@@ -34,24 +34,31 @@ done
 default_interface=$(awk '$2 == 00000000 { print $1 }' /proc/net/route)
 ip_address=$(ip addr show $default_interface | awk '$1 == "inet" {print $2}' | cut -f1 -d/)
 
+# Use SALT_DIR is set and not empty else use our default directory
+# This can be used to point to different state files for orchestration.
+if [ -z ${SALT_DIR+x} ]
+then
+    log "You did not provide a SALT_DIR. $(dirname $PWD)/kubernetes/salt will be used..."
+    salt_dir="$(dirname $PWD)/kubernetes/salt"
+#    sleep 5
```
MaximilianMeister: isn't it better to leave this uncommented?

ereslibre: Yes, sorry! :+1:

PR #54, opened 2017-02-27
Add an external pillar.

ereslibre:
The pillar is not programatically modifiable. Since we need to set
some pillar values during the bootstrapping process, we need to rely
on an external pillar that we can modify.

The original module requires a '%s' argument to be provided, or the
mysql python module will complain. We added a `minion_id` that can be
used for certain pillar key discovery. However, since we are going to
use this pillar information for global configuration, we can just let
`minion_id` be NULL, and that will be readable by all salt clients.

We also add a Pillar model that will be able to modify these Pillar
values.

### DIFF
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+# Pillar represents a pillar value on Salt.
+#
+# This can override already existing pillar values, or create completely new ones.
+#
+# For example, given that we have on our static pillar:
+#
+#   certificate_information:
+#     subject_properties:
+#       C: DE
+#       O: SUSE
+#
+# We could override them by creating the following Pillar records:
+#
+#   Pillar.create pillar: "certificate_information.subject_properties.C", value: "FR"
+#   Pillar.create pillar: "certificate_information.subject_properties.O", value: "My Company"
+#
+# Lists can be represented as collisions. So, creating the following Pillar records:
+#
+#   Pillar.create pillar: "my_service.extra_ips", value: "127.0.0.1"
+#   Pillar.create pillar: "my_service.extra_ips", value: "127.0.0.2"
+#
+# Would match the following YAML representation:
+#
+#   my_service:
+#     extra_ips:
+#       - 127.0.0.1
+#       - 127.0.0.2
+#
+# This is because how we configure our salt master using this table as an external pillar. For
+# further information you can visit:
+#   - https://github.com/kubic-project/velum/blob/master/docs/salt.md
+#   - https://github.com/kubic-project/velum/blob/master/kubernetes/salt/config/master.d/salt-master.conf
+class Pillar < ApplicationRecord
+  validates :pillar, presence: true
+  validates :value, presence: true
+
+  scope :global, -> { where minion_id: nil }
+end
--- a/docs/salt.md
+++ b/docs/salt.md
@@ -1,0 +1,58 @@
+# Salt usage
+
+This is a technical document explaining how we use and take advantage of `Salt`.
+
+Terminology:
+
+- We use the term *minion* to refer to the salt minions, because that is the way they are called in
+  Salt documentation.
+
+# Salt API
+
+We do use the [Salt API](https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html)
+for almost all operations we perform on `Salt` from within `Velum`. There are some exceptions, that
+we will describe in this document.
+
+## Bootstrapping process
+
+During the bootstrapping process, we do several tasks with `Salt`. Mainly:
+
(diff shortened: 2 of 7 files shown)


app/models/pillar.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# Pillar represents a pillar value on Salt.
```
jimmykarily: I think it would be good to add an example and the actual use case here. I would like to understand why this is needed and what kind of pillar data we push this way.

jimmykarily:
Do you think it would make sense to add an end to end test for this one? E.g. Create a pillar value from velum and check the value on the minion.

app/models/pillar.rb
ereslibre: Every setting that will be part of the bootstrapping process and that affects the whole cluster will be set as pillar information. E.g: the number of total etcd instances that will be created, certificate information (company name, e-mail, country...), possibly the timezone...

ereslibre:
It makes a lot of sense, @jimmykarily. Actually, we would be checking that the merge of the information is properly done. It is actually a little tricky: https://github.com/kubic-project/velum/pull/54/files#diff-7afc35426a5be810373b1b1917642c22R16.

Considering we have this on our pillar:

```
certificate_information:
  subject_properties:
    C: DE
...
```

We can override this information by doing `Pillar.create minion_id: nil, pillar: 'certificate_information.subject_properties.C', value: 'ES'`, for example. We don't need to change our current pillars or states, and we can just override the keys that we want, being able to fallback to the default on our static pillar if not specified.

jimmykarily:
ok this sounds great. I would:

1. Add this information on the top of the Pillar class (the example you sent and the fact that if overrides).
2. Create an end to end test as soon as [this](https://github.com/kubic-project/e2e-tests/pull/3/files) is merged.

kubernetes/salt/config/master.d/salt-master.conf
```diff
@@ -3,8 +3,18 @@ auto_accept: True
...
+mysql:
+  host: '127.0.0.1'
+  user: 'root'
+  pass: 'salt'
+  db: 'velum_development'
+  port: 3306
+
+ext_pillar:
+  - mysql:
+    - query: 'SELECT SUBSTRING_INDEX(SUBSTRING_INDEX(pillar, ".", 1), ".", -1) AS key1,
```
jimmykarily: Can we add a comment here? An example of a database record and a link to documentation would be nice ([docs](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.mysql.html)?)

ereslibre:
Added an explanation and a link to SaltStack's documentation.

ereslibre:
Added a specific document under `docs` explaining how we use `Salt`. Please, if you think there is something missing or something that can be reworded to make it more clear I'd love to know.

docs/salt.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Grains
+
+Grains are tiny amounts of information, automatically collected, or manually set.
+
+During the bootstrapping process, we will set `roles` for all minions. Right now, from the UI it
+will be possible to select which machine will take the kubernetes master role (that means, the
+machine that will run the `apiserver`, `controller-manager`...).
+
+This `role` is a grain, and is set through the `Salt API` from the `Velum` dashboard. The rest of
```
jimmykarily: Maybe make `velum` a link to the repo?

jimmykarily: sorry this repo. ignore this.

ereslibre: Yup :smile:

docs/salt.md
```diff
@@ -0,0 +1,56 @@
...
+process, and never change again (at least in a long time).
+
+The pillar cannot be modified using the `Salt API`. For this reason, we are using an external
+Pillar, that complements the static Pillar that we already have.
+
+#### External Pillar
+
+Since SaltStack is highly modular, there are several ways to set up an external pillar. In our case,
+we are setting an external pillar that will rely on our `mysql` instance running on the controller
+node.
```
jimmykarily: Link to mysql ext_pillar module missing ([this](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.mysql.html)?)

app/models/pillar.rb
ereslibre: Added a comment on the class explaining its behavior.

docs/salt.md
ereslibre: Added, also to the far more detailed documentation at https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.sql_base.html

inercia:
LGTM :+1:

flavio:
I know I'm late to the party, but this LGTM too 👍

PR #50, opened 2017-02-23
Features/ui base

santiph:
Base implementation of latest design so far.
To see them in action, 6 new urls has been created for demo purposes:
`/ui-demo/1`
`/ui-demo/2`
`/ui-demo/3`
`/ui-demo/4`
`/ui-demo/5`
`/ui-demo/6`

[Annotations.pdf](https://github.com/kubic-project/velum/files/796624/Annotations.pdf)

PR #55, opened 2017-02-27
Ui integration Part1

MaximilianMeister:
I've included https://github.com/kubic-project/velum/pull/50 here and started to add the integration part on top.

i have dropped all views that we don't need yet to have some clean state in the application, that we can add the other views afterwards step by step.

there are still some undone tasks:

- [ ] decide on which page (url/route) to show the node list where we can select the master
- [ ] fix polling and dynamically update the node list table
- [ ] fix/add unit tests once we have a fixed place for the views
- [ ] cleanup: make sure we don't have any old files/code left around

For anyone that wants to try it out, it should already work to select a master at the `/nodes` view

velum-ui-choose-master


app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,68 @@
...
+      | &times;
+  i.fa.fa-4x.pull-left aria-hidden="true"
+  span
+    | After choosing the master and clicking "Apply" all other nodes will be set to the minion role
+h1 Nodes
+.row
+  .col-xs-12.discovery-control
+    span #{Minion.count} Nodes found
+    a.btn.btn-default[href="/nodes"] Refresh
+= form_tag(nodes_path, method: "put")
```
santiph: Shouldn't this replace line 15?

app/views/shared/_header.html.slim
```diff
@@ -1,14 +1,5 @@
...
-.text-right.user-header
-  .button-container
-    .username-logout
-      .hidden-xs
-        = link_to edit_user_registration_path, class: 'btn-inline-header', data: {placement: 'bottom', toggle: 'tooltip'}, title: 'My profile' do
-          = user_image_tag(current_user)
-          span.username = current_user.email
-      = link_to destroy_user_session_path, method: :delete, class: 'topbar btn btn-default', id: 'logout', data: {placement: 'bottom', toggle: 'tooltip'}, title: 'Logout' do
-        i.fa.fa-sign-out
+nav.navbar.navbar-default
```
santiph: Header is already resolved in the layout: 
https://github.com/kubic-project/velum/pull/55/files#diff-1ccd3e5828c7330cdc2450c78847cd35R17

This fine should should be removed.

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,68 @@
...
+              - else
+                = radio_button_tag "roles[#{m.hostname}]", ["master"]
+  .clearfix.text-right.steps-container
+    a.btn.btn-danger[href="/"] Back
+    - if Minion.all.collect(&:role).include?("master")
+      = submit_tag "Apply", class: "btn btn-primary", disabled: true
+    - else
+      = submit_tag "Apply", class: "btn btn-primary"
 h1
   | Bootstrap cluster
```
santiph: Only one `h1` per page should be used

app/views/nodes/index.html.slim
MaximilianMeister: right, fixed

app/views/shared/_header.html.slim
MaximilianMeister: i have added the code from layout to the shared header, as it is somehow convention to keep the header file separate

app/views/nodes/index.html.slim
MaximilianMeister: dropped. i guess @ereslibre will catch up on the bootstrapping part, I just needed to add the button somewhere ; )

app/views/layouts/application.html.slim
```diff
@@ -1,26 +1,31 @@
...
-    .container-footer
-      .row-0
-        .col-xs-6
-          p
-            'Created with
-            i.fa.fa-heart
-            'by the SUSE team
+  body class="#{yield :body_class}"
+    .container
+      header.row
```
santiph: I'd suggest to also move the header container inside the partial

app/views/nodes/index.html.slim
santiph: Shouldn't the `.nodes-container` be indented to be contained by the `form_tag`?

app/views/layouts/application.html.slim
MaximilianMeister: I tried that, but it either breaks the whole layout, or i get a rendering error... : (

santiph: This works for me:
Layout:
```
  body class="#{yield :body_class}"
    .container
      = render :partial => "shared/header"
      .row
```

Partial:
```
header.row
  .col-xs-12
      .page-header
        h2 SUSE&reg; Container as a Service Platform
```

MaximilianMeister: oh now i see, had `.header.row` on top of the partial ... that broke it.
fixed

app/views/nodes/index.html.slim
santiph: Bootstrap button is not present in Ken's design. That should be removed at some point

santiph:
Looks good to me, so far.
Please, in further changes, do not squash/amend/force push here. Having 49 files so far is very difficult to keep reviewing your patches and changes that way. With individual commits is simpler to focus on reviewing small pieces.
Once approved, Squash & Merge github option can be used

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+                i.fa.fa-refresh.fa-stack-1x.fa-spin.fa-inverse aria-hidden="true"
+          td
+            | #{m.hostname}
+          td
+            | #{m.role}
+          td
+            | #{m.highstate}
+          td
+            - if m.role == "master"
+              = radio_button_tag "roles[#{m.hostname}]", ["master"], true, disabled: true
```
jimmykarily: The radio buttons should have the same name to behave like radio buttons (only one selected). Right now the user can select all of them.

MaximilianMeister: that somehow slipped in, seems like a regression. thanks for the hint

any idea how i can achieve that without having to change the back-end? i am a bit dependent now on that way, as only like this it will generate the correct payload `{ roles: { "hostname1" => ["master"] }`

it's our custom payload that makes the thing a bit more tricky

jimmykarily: I would prefer if we changed the payload to match the view logic. The name of the radio button is supposed to be "thing" we are trying to set and the value is the value of that "thing". In other words, we are using the radio button in order to set the master node. That means the name of the elements should be "master-minion" or master-node" and the value should be the id of the selected node. How is that translated on the back-end is up to us.

I suggest we change our controller code so it takes the simplest payload ( "master-node" => minion_id ) and prepare it for the assign_roles method. I'm not sure but I think we will also need to add a hidden field which will be an array of the workers (minion ids) currently visible. We will need to use this in order to determine the workers that will get the minion role.

MaximilianMeister: makes sense.

> 'm not sure but I think we will also need to add a hidden field which will be an array of the workers (minion ids) currently visible. We will need to use this in order to determine the workers that will get the minion role.

this is handled automatically, as the default is always minion

jimmykarily: We have to double check this but in a past meeting I think we discussed that only the current visible "workers" will be assigned. Any other machine that registers after we click on "Apply", will be ignored. That means we need to "freeze" the workers to the currently visible or else we might assign other machines that the user never saw. This might be an edge case or of minor importance so we need to check with the product managers.

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+    span #{Minion.count} Nodes found
+    a.btn.btn-default[href="/nodes"] Refresh
+= form_tag(nodes_path, method: "put")
+.nodes-container
+  table.table
+    thead
+      tr
+        th
+          | Id
+        th
```
santiph: Use `th.text-center`

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+          | Highstate
+        th
+          | Master
+    tbody
+      - Minion.all.each do |m|
+        tr
+          th
+            | #{m.id}
+          td
```
santiph: Use `td.text-center`

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+          | Id
+        th
+          | Status
+        th
+          | Hostname
+        th
+          | Role
+        th
+          | Highstate
+        th
```
santiph: Use `th.text-center`

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+              span.fa-stack aria-hidden="true"
+                i.fa.fa-circle.fa-stack-2x.text-success aria-hidden="true"
+                i.fa.fa-refresh.fa-stack-1x.fa-spin.fa-inverse aria-hidden="true"
+          td
+            | #{m.hostname}
+          td
+            | #{m.role}
+          td
+            | #{m.highstate}
+          td
```
santiph: Use `td.text-center`

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+            - if m.role
+              i.fa.fa-check-circle-o.text-success.fa-2x aria-hidden="true"
+            - else
+              span.fa-stack aria-hidden="true"
+                i.fa.fa-circle.fa-stack-2x.text-success aria-hidden="true"
+                i.fa.fa-refresh.fa-stack-1x.fa-spin.fa-inverse aria-hidden="true"
+          td
+            | #{m.hostname}
+          td
+            | #{m.role}
```
santiph: Given the purpose of the page is to select the master for the new cluster installation (line 49-54), isn't this column wrong?

MaximilianMeister: i think it's nice to see the role directly after you assigned it

app/assets/stylesheets/pages/nodes_list.scss
```diff
@@ -0,0 +1,28 @@
+.discovery-control {
```
jimmykarily: Maybe it is better if we use css file names that map to controllers. We can have a class on the body element of each page (generated by the current controller and current action) so inside the `app/assets/stylesheets/pages/nodes.scss` we will have the css for nodes#index nodes#show nodes#edit etc. I can send a commit that does that if you agree.

app/assets/stylesheets/velum_general.scss
```diff
@@ -0,0 +1,64 @@
...
+        width: 100%;
+
+        p {
+            margin: 0px;
+            font-size: 11px;
+        }
+    }
+}
+
+@import 'pages/nodes_list';
```
jimmykarily: I think is would be simpler if only used `@import` inside the application.scss. It seems to me that this file acts as the "super" css file that should import all the velum css. The application.scss can already take that role. We will only need to extract the rules above in a separate css file (common.scss or somthing like that) and import it in the application.scss. I can take care of that if we agree.

app/views/layouts/application.html.slim
```diff
@@ -1,26 +1,30 @@
...
-  footer
-    .aside-footer
-    .container-footer
-      .row-0
-        .col-xs-6
-          p
-            'Created with
-            i.fa.fa-heart
-            'by the SUSE team
+  body class="#{yield :body_class}"
```
jimmykarily: Shoudn't there be a `content_for :body_class` somewhere? This seems like an implementation of what I proposed above (regarding classes on the body element) but it doesn't seem to be complete.

ereslibre: Yes, you can find them on the original PR (https://github.com/kubic-project/velum/pull/50/files)

jimmykarily: ok but there we had non-restful demo pages. Now I think we can use this body_class to organize our css files as well (as per my previous comment). Let me know what you think so I can prepare a commit.

app/views/nodes/index.html.slim
jimmykarily: In the original design the page where we add nodes is not the same as the monitoring one although this pages serves both roles now. In the add nodes page, what @santiph says, makes sense although @eotchi suggested that we keep this column because in the future we might want to assign roles to minions that already have a role (Federica please confirm this).

Whatever the case, we can create the 2 separate pages (add nodes and monitor) either here or in a separate branch after we merge this one. I suggest we do the latter because this one already includes a lot of cleaning up and it is important to merge it so others can work in parallel on other issues.

santiph: I think that's what the radio button defined on line 49-54 do. I haven't seen it on the design, but you can consult Ken about UI changes

app/views/nodes/index.html.slim
```diff
@@ -1,12 +1,65 @@
...
+        th
+          | Id
+        th.text-center
+          | Status
+        th
+          | Hostname
+        th
+          | Role
+        th
+          | Highstate
```
jimmykarily: This is Salt terminology. We agreed to not expose Salt terms in the UI since this is something internal and not relevant to the user. Also the same information this provides, `status` column also does so it is redundant.

app/assets/stylesheets/pages/nodes_list.scss
santiph: I like the idea, but in this case, I think the `nodes_list` is more like a reusable component across several pages than a page itself.
Maybe for the backend it makes more sense to convert it into its own controller... IDK

jimmykarily: Well in that case we should move it from `pages/...` to `widgets` or `components` or something like that. Only the part of the code that is reusable of course.

jimmykarily:
I added a Logout button to fix a couple of tests. It need styling but I think we can fix this together with other things that do not match the design in another branch.

The rest of the failing tests have to do with the minion polling javascript code being removed. We either bring back the polling functionality here or we remove the tests.

app/views/shared/_header.html.slim
```diff
@@ -1,7 +1,7 @@
 header.row
-  .col-xs-6
-    .page-header
-      h2 SUSE&reg; Container as a Service Platform
-  .col-xs-6
-    - if user_signed_in?
-     = link_to('Logout', destroy_user_session_path, class: "btn btn-default pull-right", :method => :delete)
+  .col-xs-12
+    .page-header.col-xs-6
```
santiph: columns cannot be child of another column. You need a row for that.

app/views/shared/_header.html.slim
```diff
@@ -1,7 +1,7 @@
...
+        h2 SUSE&reg; Container as a Service Platform
+    .col-xs-6
+      - if user_signed_in?
+      = link_to('Logout', destroy_user_session_path, class: "btn btn-default pull-right", :method => :delete)
```
santiph: I've never seen a logout action using a delete method.

ereslibre: On a REST context is very common, "destroy this session".

jimmykarily: I just copied [the docs](https://github.com/plataformatec/devise/wiki/How-To:-Add-sign_in,-sign_out,-and-sign_up-links-to-your-layout-template)

app/views/shared/_header.html.slim
jimmykarily: You are right. I think the col-xs-12 is not needed anymore. Right?

app/views/shared/_header.html.slim
santiph: Do you have any example working on another web application that handles users? I'm only seeing Get and Posts. (And Get is being highly discouraged because of browsers pre-fetching GET pages).

Stackoverflow users seem to prefer POST
http://stackoverflow.com/questions/5868786/what-method-should-i-use-for-a-login-authentication-request
http://softwareengineering.stackexchange.com/questions/196871/what-http-verb-should-the-route-to-log-out-of-your-web-app-be

MaximilianMeister: that's how `devise` is implemented, and i think we don't need to care much about that

santiph: Just saw your comment @jimmykarily 
Ignore this

app/views/shared/_header.html.slim
santiph: Login/out buttons are usually part of the headers in the front end. You have to validate it with Ken.

app/assets/stylesheets/pages/nodes_list.scss
santiph: I'll leave ruby organisation and structure to you guys

app/views/nodes/index.html.slim
santiph: I'm not sure what implementation is best in Ruby (single page vs two pages), but in terms of what the user will see, this doesn't match Ken's design for none of the pages.
And as I told you guys already, we're still on time to iterate over the design one more time, but you need to talk with the designer and raise this inconsistencies! He's responsible for the UI of the product, after all.

app/assets/stylesheets/pages/nodes_list.scss
ereslibre: I think this MR is already a monster, we can reorganize later on extra PR's, unless there's something "blocking".

app/assets/stylesheets/velum_general.scss
ereslibre: I agree on trying to avoid `@import` outside the `application.scss`, but I'd left this changes for future PR's.

app/controllers/nodes_controller.rb
```diff
@@ -42,12 +42,11 @@ def update
   # assigned role, assign a random role to it, and then call the salt
   # orchestration.
   def bootstrap
-    if Minion.where(role: nil).count > 1
-      # choose first minion to be the master
-      Minion.assign_roles!(roles: { Minion.first.hostname => ["master"] })
+    if Minion.all.collect(&:role).include?("master")
```
ereslibre: This could be more rails-like: `Minion.exists?(role: :master)`, it's faster too.

app/models/minion.rb
```diff
@@ -15,20 +15,20 @@ class NonExistingMinion < StandardError; end
...
   #     default_role: :dns
   #   )
   def self.assign_roles!(roles: {}, default_role: :minion)
-    requested_master = roles.detect { |_name, r| r.include?("master") }.first
-    requested_minions = roles.select { |_name, r| r.include?("minion") }.keys
-    if roles.values.flatten.include?("master") && !Minion.exists?(hostname: requested_master)
-      raise NonExistingMinion, "Failed to process non existing minion: #{requested_master}"
+    requested_master = roles[:master]
+    requested_minions = roles[:minion]
+    if roles[:master] && !Minion.exists?(id: requested_master)
```
ereslibre: I'd use `!requested_master.blank? && !Minion.exists?(id: requested_master)`. In this kind of checks when taking external information is better to use `blank?` because it also checks for emptyness, also refer to `requested_master` since we already bound the variable.

app/views/layouts/application.html.slim
ereslibre: Yes, I'd try to merge this as soon as possible so we can work in parallel on different improvements.

PR #56, opened 2017-02-28
docs: added internal research on salt job callbacks

mssola:
Open sourcing an internal research that was done by @jimmykarily.

### DIFF
--- a/docs/salt-job-callbacks.md
+++ b/docs/salt-job-callbacks.md
@@ -1,0 +1,34 @@
+# Salt Job Callbacks
+
+Completion of commands sent to Salt usually affect our Minion's and their
+current state.  For example, when we send an "orchestrate" command (see
+`Velum::Salt.orchestrate`) a number of states are applied (a highstate) and the
+final result is that the Minions already assigned a role are actually setup for
+that role. This is something we need to show on our UI and also something we
+need to be aware of in order to perform other tasks (e.g. bring down a Minion).
+
+For the rest of this document we will call one standalone action on a Minion as
+"Action".  An example of an action is "Turn a Minion to the assigned
+role". Although an action might include a number of salt commands (and salt
+jobs) we will assume for now that there is one salt job that designates the
+action's completeness. To better understand this lets use an example.
+
+Lets assume we have 2 salt minions available and we want to "orchestrate"
+them. This is done in 2 steps.  First we assign the roles of master and minion
+to the 2 minions. Then we run the ochestrate method.  The first step will simply
+set the `role` column of the minion. Another integer column called `highstate`
+will contain the current state:
+
+- `:not_applied`: salt orchestration has not been applied yet.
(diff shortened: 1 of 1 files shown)


ereslibre:
This is actually a bit outdated and doesn't really match reality. Should we update it?

mssola:
@ereslibre yeah. The idea is to include this first version, and then add more info on top of it. We could have this PR as [WIP]

docs/salt-job-callbacks.md
```diff
@@ -0,0 +1,26 @@
...
+also something we need to be aware of in order to perform other tasks (e.g. bring down a Minion).
+
+For the rest of this document we will call one standalone action on a Minion as "Action".
+An example of an action is "Turn a Minion to the assigned role". Although an action
+might include a number of salt commands (and salt jobs) we will assume for now that
+there is one salt job that designates the action's completeness. To better understand this lets use an example.
+
+Lets assume we have 2 salt minions available and we want to "orchestrate" them. This is done in 2 steps.
+First we assign the roles of master and minion to the 2 minions. Then we run the ochestrate method.
+The first step will simply set the ```role``` column of the minion. Another column
```
ereslibre: Too many backticks, I'd leave only one backtick. Same on next line.

docs/salt-job-callbacks.md
```diff
@@ -0,0 +1,26 @@
...
+called ```highstate``` will be left with a nil value. That means, a highstate
```
ereslibre: Since this document was written, `highstate` can take different values: `:not_applied, :pending, :failed and :applied`. Mainly: `not_applied` is when no orchestration has been run at all (but the minion exists in our database because it has registered itself against the salt master); `pending` is when the orchestration has been started and we are waiting for its result; `failed` is when we got an orchestration failure for that minion; and `applied` is when we got a success coming from the orchestration for that node.

They are a rails `enum`, so we will actually write integers to the database.

docs/salt-job-callbacks.md
```diff
@@ -0,0 +1,26 @@
...
+has not been applied yet on that Minion.
+
+When we run the orchestrate method a highstate will be triggered on all Minions.
+We will monitor the highstate completion events and as soon as we get one, we
+will update the "highstate" column of that minion. This way we know which Minions
+have been setup and which not.
```
ereslibre: And if they returned an error or not.

PI-Victor:
we should probably get this merged

mssola:
Just fixed the comments from @ereslibre. Sorry for the delay (I simply forgot about this PR 😅)

MaximilianMeister:
jsut docs, merging

mssola:
> Ideally we need to expand it but let's get it merged first 👏

Agreed. If I recall correctly, this PR was just about open sourcing documentation from our gitlab instance. We should improve it further.

PR #59, opened 2017-03-03
Use jQuery plugin for minion polling

ereslibre:


spec/views/nodes/index.html.slim_spec.rb
```diff
@@ -2,27 +2,26 @@
...
 
     it "polls for minions" do
-      render
-
-      script = assert_select("script").children.text
-      expect(script.strip).to eq "MinionPoller.poll();"
+      section = assert_select("#nodes")
+      expect(JSON.parse(section.attribute("data-plugin").value).first["name"]).to eq "velumMinionPoller"
     end
+
```
mssola: nit: remove this empty line

app/assets/javascripts/application.js
```diff
@@ -11,6 +11,7 @@
 // about supported directives.
 //
 //= require jquery
+//= require bootstrap-sprockets
```
mssola: You don't really need this. Take a look at what we are doing in Portus: https://github.com/SUSE/Portus/blob/master/app/assets/javascripts/application.js

mssola: Just saying because this hit me when building the RPM prepared for containerizing Portus 😉

app/assets/javascripts/velumMinionPoller.js
```diff
@@ -0,0 +1,123 @@
+// the semi-colon before function invocation is a safety net against concatenated
+// scripts and/or other plugins which may not be closed properly.
+;( function( $, window, document, undefined ) {
+
+    "use strict";
+
+        // undefined is used here as the undefined global variable in ECMAScript 3 is
```
mssola: I don't understant the extra indentation from this line forward.

mssola: Moreover, lots of these comments looks like they provide from somewhere else. Could you remove them, or trim as much as possible ? (while adding a comment giving credits)

app/assets/javascripts/velumMinionPoller.js
```diff
@@ -0,0 +1,123 @@
...
+
+            processNodes: function (nodesResponse) {
+                var self = this;
+                $.each(nodesResponse, function (index, value) {
+                     self.renderNode(value);
+                });
+            },
+
+            renderNode: function (node) {
+
```
mssola: Remove empty line

app/assets/javascripts/velumMinionPoller.js
```diff
@@ -0,0 +1,123 @@
...
+                if ( $(this.element).has('#node-' + node.id).length ) {
+
```
mssola: Same

app/assets/javascripts/velumMinionPoller.js
```diff
@@ -0,0 +1,123 @@
...
+        // preventing against multiple instantiations
+        $.fn[ pluginName ] = function( options ) {
+            return this.each( function() {
+                if ( !$.data( this, "plugin_" + pluginName ) ) {
+                    $.data( this, "plugin_" +
+                        pluginName, new Plugin( this, options ) );
+                }
+            } );
+        };
+
```
mssola: Same

app/assets/javascripts/velumMinionPoller.js
```diff
@@ -0,0 +1,123 @@
...
+                            '</dl>',
+                        '</div>'
+                    ].join('');
+
+                    $( this.element ).append( nodeRow );
+                }
+            },
+
+            handleError: function () {
+                // To be defined
```
mssola: Could you implement this ?

santiph: This behaviour is yet to be defined by PO

app/assets/javascripts/velumMinionPoller.js
santiph: I'd prefer to keep it for readability purposes

app/assets/javascripts/velumMinionPoller.js
santiph: Maybe... How? Where?
Source: https://jqueryboilerplate.com/

santiph: I haven't noticed indentation until now. Will be fixed

app/assets/javascripts/application.js
santiph: I have no idea how the building RPM works, but I've tried to follow bootstrap instructions to make it work correctly with Ruby. (https://github.com/twbs/bootstrap-sass#a-ruby-on-rails)
Feel free to change it as needed

app/assets/javascripts/velumMinionPoller.js
mssola: Sure, then put a comment on top stating that the skeleton of this poller comes from that URL 😉

app/assets/javascripts/application.js
mssola: Just saying because we are not using sprockets afaik, and this has some consequences when building the RPM. One easy step would be to replace this line with:

```javascript
//= require bootstrap
```

app/assets/javascripts/velumMinionPoller.js
mssola: Ok, thanks.

app/assets/javascripts/velumMinionPoller.js
MaximilianMeister: i'd also prefer to drop this empty line. and why is it better readable only for this method and not the others?

santiph: It's easier for me to read multiline method's definitions with spaces than without them:

```
            renderNode: function (node) {
                if ( $(this.element).has('#node-' + node.id).length ) {
                    return;
                } else {
                    var nodeRow = [
                        '<div id="node-' + node.id + '">',
                            '<a href="/nodes/' + node.id + '">' + node.hostname + '</a>',
                            '<dl>',
                                '<dt>Created at</dt>',
                                '<dd>' + node.created_at + '</dd>',
                                '<dt>Updated at</dt>',
                                '<dd>' + node.updated_at + '</dd>',
                            '</dl>',
                        '</div>'
                    ].join('');
                    $( this.element ).append( nodeRow );
                }
            },

```

vs

```

            renderNode: function (node) {

                if ( $(this.element).has('#node-' + node.id).length ) {

                    return;

                } else {
                    var nodeRow = [
                        '<div id="node-' + node.id + '">',
                            '<a href="/nodes/' + node.id + '">' + node.hostname + '</a>',
                            '<dl>',
                                '<dt>Created at</dt>',
                                '<dd>' + node.created_at + '</dd>',
                                '<dt>Updated at</dt>',
                                '<dd>' + node.updated_at + '</dd>',
                            '</dl>',
                        '</div>'
                    ].join('');

                    $( this.element ).append( nodeRow );
                }
            },
```

santiph: A coding style guide needs to be defined for the project and enforced through `eslint` or any similar linting tool for javascript

MaximilianMeister:
@santiph Unfortunately we can't add more moving pieces here regarding the javascript, as we're getting really tight in the schedule and we decided it doesn't make much sense to invest in a custom javascript architecture right now, as we we'll probably move away to some proper js library some time soon.

Thank's for the work on this, but we'll rather update the minion poller to do what we need for the nodes page

PR #61, opened 2017-03-06
[WIP] setup: don't launch setup_environment in the background

MaximilianMeister:
the logs are lost in our ci testing and there seems to be no real
reason to not run it in the foreground

not 100% sure if my assumption is really true here... but i'll check in the e2e gating

## related to
https://github.com/kubic-project/e2e-tests/pull/13

### DIFF
--- a/kubernetes/start
+++ b/kubernetes/start
@@ -63,7 +63,7 @@ done
 
 if $(which kubelet >&/dev/null); then
   log "Running setup-environment script"
-  bash setup-environment &
+  ./setup-environment
 else
   log "To launch kubelet make sure to install kubernetes-node package first and run this script again..."
   exit 1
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
build got stuck for some reason ...

MaximilianMeister:
so with the test commit to re-enable the backgrounding it seems to work again. seems that i am missing something in the logic here. sorry for the noise

kubernetes/start
```diff
@@ -63,7 +63,7 @@ done
 
 if $(which kubelet >&/dev/null); then
   log "Running setup-environment script"
-  bash setup-environment &
+  ./setup-environment
```
jimmykarily: If you don't run this in the background, the script will stop here until it returns. But it will never do, because setup-environment script needs the kubelet to be started which only happens on line 73 below.

PR #62, opened 2017-03-06
Add setup process

ereslibre:
If no minions have any role assigned, we will redirect to `/setup` where we will wait for user input for certain settings.

`/setup` cannot be visited if there is any minion with an assigned role.

We will also run the discovery, assign roles and run the orchestration.

The root path `/` will contain a basic monitoring tool.


ereslibre:
I need to leave in a while and I don't want to block anybody. There are some tests missing, and others that need to be adapted to the new structure.

Pillar information is actually written into database and navigation works properly.

ereslibre:
Please, note that I changed the base branch, instead of `master` it will be merged into `setup-process` branch so you can continue from there on.

PR #63, opened 2017-03-06
bootstrap right after choosing the master

MaximilianMeister:


MaximilianMeister:
might have a small influence on https://github.com/kubic-project/velum/pull/62 but whatever gets meged first should be rebased then

spec/controllers/nodes_controller_spec.rb
```diff
@@ -134,6 +134,7 @@
           { hostname: "minion0.example.com" },
           { hostname: "minion1.example.com" }
         ]
+        allow(Velum::Salt).to receive(:orchestrate)
```
ereslibre: Here we are allowing message, but we are not using its counterpart `expect(Velum::Salt).to receive(:orchestrate)`, right? Here we are basically setting the double, but not using it.

spec/controllers/nodes_controller_spec.rb
```diff
@@ -181,6 +182,7 @@
...
```
ereslibre: Same here.

spec/controllers/nodes_controller_spec.rb
MaximilianMeister: updated

app/views/nodes/index.html.slim
```diff
@@ -48,15 +47,13 @@ h1 Nodes
...
+            - elsif Minion.exists?(role: "master")
               = radio_button_tag "roles[master]", m.id, false, disabled: true
             - else
               = radio_button_tag "roles[master]", m.id
   .clearfix.text-right.steps-container
     a.btn.btn-danger[href="/"] Back
-    - if Minion.all.collect(&:role).include?("master")
-      = submit_tag "Apply", class: "btn btn-primary", disabled: true
+    - if Minion.exists?(role: "master")
+      = submit_tag "Bootstrap Cluster", class: "btn btn-primary", disabled: true
```
jimmykarily: Clicking on this button when no minions have registered (and thus no master selected) results to:

```

ActionController::ParameterMissing in NodesController#update_nodes
param is missing or the value is empty: roles
```

MaximilianMeister: I've added a generic rescue_from in the application controller, you think that's fine? not sure what's the best practice in those cases

jimmykarily: I think, UX wise, the correct thing to do would be to not allow the user to click on a button that will result in an error. Or maybe, show a client side message that will explain why this action is not permitted. On the other hand, since we are not the UX experts here, I would return a friendlier message to the user, not the exception message. E.g. "Please spawn some minions before you try to bootstrap a cluster" or something like that.

PR #65, opened 2017-03-07
UI: setup process

MaximilianMeister:
this includes the initial work from @ereslibre https://github.com/kubic-project/velum/pull/62 and some fixes and enhancements on top

this should make the initial wizard complete and usable. further changes, like the polling and adapting the texts is still to be done, but it can happen in a separate PR

### NOTE
the texts in the screenshots are not the final ones, and also don't represent what we have now in the code

casp-1

casp-2

casp-3

casp-4


app/models/minion.rb
```diff
@@ -22,7 +25,7 @@ class NonExistingMinion < StandardError; end
   #   )
   def self.assign_roles!(roles: {}, default_role: :minion)
     requested_master = roles[:master]
-    requested_minions = roles[:minion]
+    requested_minions = roles[:minion] || (Minion.unassigned_role.pluck(:id) - [requested_master])
```
jimmykarily: [MINOR] After this addition, the check on line 34 below is no longer needed I think.

app/views/dashboard/index.html.slim
```diff
@@ -1,4 +1,53 @@
-h1 Dashboard
+h1 Simple Monitoring
 .row
   .col-xs-12.discovery-control
-    a.btn.btn-primary[href="/nodes"] Choose master
+    span #{Minion.count} Nodes found
```
santiph: Simple Monitoring page doesn't have the Nodes count, but Add Nodes button

app/views/dashboard/index.html.slim
```diff
@@ -1,4 +1,53 @@
...
+        th
+          | Id
+        th.text-center
+          | Status
+        th
+          | Hostname
+        th
+          | Role
+        th
+          | Highstate
```
santiph: Role and Highstate are not present in the design.
Status is the first column, Node Identifier is the second one
Missing RAM, CPU and Model columns

app/views/dashboard/index.html.slim
```diff
@@ -1,4 +1,53 @@
...
+            | #{m.highstate}
+          td.text-center
+            - if m.role == "master"
+              = radio_button_tag "roles[master]", m.id, true, disabled: true
+            - elsif Minion.all.collect(&:role).include?("master")
+              = radio_button_tag "roles[master]", m.id, false, disabled: true
+            - else
+              = radio_button_tag "roles[master]", m.id
+
+script
```
santiph: Why are you using html tags inside a script?

app/views/setup/discovery.html.slim
```diff
@@ -0,0 +1,34 @@
+.alert.alert-info.alert-dismissible role="alert"
+  button.close[type="button" data-dismiss="alert" aria-label="Close"]
```
santiph: Bootstrap Alerts won't be dismissed if its javascript components are not loaded.
Please double check that, I realised I didn't include them in my original PR so the functionality was broken.

app/views/setup/discovery.html.slim
```diff
@@ -0,0 +1,34 @@
...
+    span[aria-hidden="true"]
+      | &times;
+  i.fa.fa-4x.pull-left aria-hidden="true"
+  span
+    | After choosing the master and clicking "Bootstrap cluster" all other nodes will be set to the minion role
+h1 Nodes
```
santiph: There is no "Nodes" page in the design. I think this is "Add Nodes" instead

app/views/setup/discovery.html.slim
```diff
@@ -0,0 +1,34 @@
...
+.row
+  .col-xs-12.discovery-control
+    span #{Minion.count} Nodes found
+= form_tag(setup_bootstrap_path, method: "post")
+.nodes-container
```
santiph: Shouldn't `.nodes-container` be indented 1 extra level to be contained by the form?

app/views/setup/discovery.html.slim
```diff
@@ -0,0 +1,34 @@
...
+  table.table
+    thead
+      tr
+        th
+          | Id
+        th
+          | Hostname
+        th.text-center
+          | Master
```
santiph: Missing columns: "Status", "Node Identifier", "RAM", "CPU", "Model"

app/views/setup/discovery.html.slim
```diff
@@ -0,0 +1,34 @@
...
+        tr
+          th
+            | #{m.id}
+          td
+            | #{m.hostname}
+          td.text-center
+            = radio_button_tag "roles[master]", m.id
+  .clearfix.text-right.steps-container
+    = link_to "Back", setup_worker_bootstrap_path, class: "btn btn-danger"
+    = submit_tag "Bootstrap cluster", class: "btn btn-primary"
```
santiph: Button text should be "Add Nodes"

app/views/setup/welcome.html.slim
```diff
@@ -0,0 +1,29 @@
+= content_for :body_class, "demo-1"
```
santiph: body_class needs to be renamed

app/models/pillar.rb
```diff
@@ -36,4 +36,23 @@ class Pillar < ApplicationRecord
   validates :value, presence: true
 
   scope :global, -> { where minion_id: nil }
+
+  class << self
+    def value(pillar:)
+      pillar = Pillar.find_by pillar: all_pillars[pillar]
```
jimmykarily: I suggest we don't overwrite the original argument because it was a string and now it is a Pillar object. It makes it harder to mentally run the code because have to keep track of what the variable is at each point.

app/models/pillar.rb
```diff
@@ -36,4 +36,23 @@ class Pillar < ApplicationRecord
...
+      return pillar.value if pillar
+      nil
```
jimmykarily: [minor] these 2 lines can be one: `pillar && pillar.value` or even `pillar.try(:value)`

app/models/pillar.rb
```diff
@@ -36,4 +36,23 @@ class Pillar < ApplicationRecord
...
+    end
+
+    def all_pillars
```
jimmykarily: What is this? Can we add documentation that explains what this method provides?

app/views/dashboard/index.html.slim
```diff
@@ -1,4 +1,53 @@
...
+          td
+            | #{m.hostname}
+          td
+            | #{m.role}
+          td
+            | #{m.highstate}
+          td.text-center
+            - if m.role == "master"
+              = radio_button_tag "roles[master]", m.id, true, disabled: true
+            - elsif Minion.all.collect(&:role).include?("master")
```
jimmykarily: This creates N+1 queries but I have fixed that already [here](https://github.com/kubic-project/velum/blob/fix_minion_polling/app/views/nodes/index.html.slim#L14) so you can leave it as is. I will take care of it.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -3,18 +3,53 @@
...
 
-    it "gets all the available minions" do
+    it "gets redirected to setup initially" do
       sign_in user
+      get :index
+      expect(response.status).to eq 302
+      expect(response.redirect_url).to eq "http://test.host/setup"
+    end
 
+    it "shows a simple monitoring" do
```
jimmykarily: maybe add "...when roles have already been assigned"?

spec/controllers/setup_controller_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+
+        get :welcome
+        expect(response.status).to eq 200
+      end
+
+      it "renders with HTML if no format was specified" do
+        sign_in user
+
+        get :welcome
+        expect(response["Content-Type"].include?("application/json")).to be_falsey
```
jimmykarily: This assertion doesn't check what the actual format is but rather what it's not. The test description says otherwise.

app/views/setup/welcome.html.slim
santiph: same thing with `demo-2` in the next page

app/views/setup/discovery.html.slim
MaximilianMeister: this could then be a flash message, but i think even that is broken atm, i'd say lets fix it in a follow up. the PR is already big enough again

app/views/setup/discovery.html.slim
MaximilianMeister: I wouldn't worry too much about the wording here, as we can't add nodes from the UI anyways. it needs to be done in the background by the user, and then the nodes are just popping up

app/views/setup/discovery.html.slim
MaximilianMeister: seems so, i tried it, but it doesn't make any difference visually, i also checked the html and it looks the same

app/views/setup/discovery.html.slim
MaximilianMeister: we don't have that info, and we decided to not require it. i think that was just some example anyways in the design

app/views/setup/welcome.html.slim
MaximilianMeister: fixed

app/views/setup/welcome.html.slim
```diff
@@ -0,0 +1,29 @@
...
+.alert.alert-info.alert-dismissible role="alert"
+  button.close[type="button" data-dismiss="alert" aria-label="Close"]
+    span[aria-hidden="true"]
+      | &times;
+  i.fa.fa-eye.fa-4x.pull-left aria-hidden="true"
+  span Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.
```
MaximilianMeister: right, as i said in the description, i'd like to do that in a follow up

app/models/minion.rb
MaximilianMeister: fixed

app/views/setup/discovery.html.slim
santiph: As far as the html (not the DOM within developer tool) is the same

app/views/setup/discovery.html.slim
kwwii: Andreas specifically said that we will have this information and it should be included. Please do so.

kwwii:
Why is there a column with "Role"? Please remove it or explain why it is needed (and perhaps why this information wasn't shared with me).

jimmykarily:
@kwwii this PR is not a complete implementation of your design. We are still working on bringing everything as close as possible to what you designed. Some information is still missing (e.g. RAM/CPU) and it is not yet provided from the back-end. Let us get this as close to the design as we can and we will arrange a meeting to discuss any differences. How does that sound?

MaximilianMeister:
@kwwii just to add more confusion, as @jimmykarily said, this is not complete yet, we just have to work on parts iteratively as we have to split the work somehow, e.g. we can't expect to implement all things (design, javascript, backend, tests, e2e-tests) in a single PR as more people work on the same codebase on different issues right now. unfortunately there is no other way without getting stuck too much at the moment, but we'll make sure that the final version meets all requirements.

app/views/setup/discovery.html.slim
MaximilianMeister: sorry for the assumption, i was missing that bit

PR #66, opened 2017-03-07
Fix minion polling

jimmykarily:
Re-enable polling. Also:

- Remove "highstate" column since it is redundant with status column in place
- Use a different icon in status column for every state

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,33 +1,95 @@
 MinionPoller = {
-    poll: function() {
-        this.request();
-        return setInterval(this.request, 5000);
-    },
+  poll: function() {
+    this.request();
+    return setInterval(this.request, 5000);
+  },
 
-    request: function() {
-        $.ajax({
-            url: $('#nodes').data('url'),
-            dataType: "json",
-            success: function(data) {
-                var rendered = "";
+  request: function() {
+    $.ajax({
+      url: $('.nodes-container').data('url'),
+      dataType: "json",
+      success: function(data) {
+        var rendered = "";
+        MinionPoller.selectedMaster = $("input[name='roles[master]']:checked").val();
 
-                for (i = 0; i < data.length; i++) {
-                    rendered += MinionPoller.render(data[i]);
-                }
-                $("#nodes").html(rendered);
-            }
-        });
-    },
+        // depending on the url we might have these data:
+        // when monitoring:
+        //  [assigned_minions: [<minion1>,<minion2>,<minion3>], unassigned_minions: []]
+        // when discovering:
+        //   [<minion1>,<minion2>,<minion3>]
+        var minions = data.assigned_minions || data;
 
-    render: function(minion) {
-        return "<div> \
-  <a href=\"/nodes/" + minion.id + "\">" + minion.hostname + "</a>\
-  <dl>\
-    <dt>Created at</dt>\
-    <dd>" + minion.created_at + "</dd>\
-\
-    <dt>Updated at</dt>\
-    <dd>" + minion.updated_at + "</dd>\
-  </dl>"
+        for (i = 0; i < minions.length; i++) {
+          if (MinionPoller.renderMode == "discovery") {
+            rendered += MinionPoller.renderDiscovery(minions[i]);
+          } else {
+            rendered += MinionPoller.render(minions[i]);
+          }
+        }
+        $(".nodes-container tbody").html(rendered);
+      }
+    });
+  },
+
+  render: function(minion) {
+    var statusHtml;
+    var checked;
+    var masterHtml;
+
+    switch(minion.highstate) {
+      case "not_applied":
+        statusHtml = '<i class="fa fa-circle-o text-success fa-2x" aria-hidden="true"></i>';
+        break;
+      case "pending":
+        statusHtml = '\
+          <span class="fa-stack" aria-hidden="true">\
+            <i class="fa fa-circle fa-stack-2x text-success" aria-hidden="true"></i>\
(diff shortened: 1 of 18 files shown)


jimmykarily:
Ruby and Javascript render different views right now. We can either skip Ruby rendering completely and let Javascript render the minions or we can duplicate the view logic in both sides. The first solution has the drawback that if Javascript is disabled (or broken for whatever reason) the user won't see any Minions. The second solution has the drawback of duplication. If we go for the second solution I would like to introduce Decorators ([draper](https://github.com/drapergem/draper) or other) to avoid polluting the view with lots of logic. What do you think? @mssola @MaximilianMeister @ereslibre

ereslibre:
If we are going for a complete JS view framework in the future (react.js, vue.js...) I'd go with the "render on JS" solution for now, because we can reuse the endpoints for live update (return data from the api, nothing view related, just data) with the new implementation. My 2 cents.

app/views/dashboard/_polling.html.erb
```diff
@@ -1,3 +1,5 @@
 <script>
-  MinionPoller.poll();
+  document.addEventListener("DOMContentLoaded", function(event) {
```
santiph: jQuery is available in this project, I'd suggest to use it instead.
Take a look at https://api.jquery.com/ready/

jimmykarily: Yes it is but it is not loaded at this point. There are ways to fix this but I didn't want to introduce any Javascript loading logic or take any decisions now that will have to be changed later. Vanilla Javascript to the rescue...

jimmykarily:
ok went with the Javascript only rendering solution for now.

jimmykarily:
I couldn't find the relevant information in the design or anywhere else so I just used an icon that looked good to me for not_applied highstate. That is the state of a new minion before it is bootstrapped. @santiph should I use something else? Keep in mind, I'm only referring to the Status icon in the screenshot below. I know there are differences between what is shown and the original design but there will be addressed in another PR. 

selection_001

santiph:
@jimmykarily I cannot give you that answer.
Icons to be used are specified in the provided design. Anything else must be clarified with Ken.

mssola:
@jimmykarily let's use these icons for now, and we can fix this later on. I'd like to get this merged asap 😸

jimmykarily:
@mssola this PR was completely broken after merging [this one](https://github.com/kubic-project/velum/pull/65). I'm working to bring this back into shape. I will tell you when I'm done.

jimmykarily:
It should be ok now.

mssola:
That looks much nicer. Thanks 👍

app/assets/javascripts/dashboard.js
```diff
@@ -1,33 +1,83 @@
...
+
+  request: function() {
+    $.ajax({
+      url: $('.nodes-container').data('url'),
+      dataType: "json",
+      success: function(data) {
+        var rendered = "";
+        MinionPoller.selectedMaster = $("input[name='roles[master]']:checked").val();
+
+        minions = data.assigned_minions || data;
```
santiph: Where is `minions` declared?

app/assets/javascripts/dashboard.js
```diff
@@ -1,33 +1,83 @@
...
+        }
+        $(".nodes-container tbody").html(rendered);
+      }
+    });
+  },
+
+  render: function(minion) {
+    switch(minion.highstate) {
+      case "not_applied":
+        statusHtml = '<i class="fa fa-circle-o text-success fa-2x" aria-hidden="true"></i>';
```
santiph: This variable is not declared

app/assets/javascripts/dashboard.js
```diff
@@ -1,33 +1,83 @@
...
+        minions = data.assigned_minions || data;
+
+        for (i = 0; i < minions.length; i++) {
+          if (MinionPoller.renderMode == "discovery") {
+            rendered += MinionPoller.renderDiscovery(minions[i]);
+          } else {
+            rendered += MinionPoller.render(minions[i]);
+          }
+        }
+        $(".nodes-container tbody").html(rendered);
```
santiph: You can avoid rewriting the content of the node with so little checks: For example
https://github.com/kubic-project/velum/blob/97fdf6e13f0f27865f03a307edf94e7138c935b9/app/assets/javascripts/velumMinionPoller.js#L84

app/assets/javascripts/dashboard.js
jimmykarily: Nowhere, I got too used to Coffescript I guess. Fixing.

app/assets/javascripts/dashboard.js
jimmykarily: A node might already exist in the table but with outdated data. At least for the monitoring page this is the case. I wouldn't try to make this script smart enough to handle that case too. At some point we will switch to a framework which will handle all these in a better way.

app/views/setup/_polling.html.slim
```diff
@@ -0,0 +1,4 @@
+= content_for :page_javascript do
+  javascript:
```
jimmykarily: turbolinks make a mess of this. Javascript keeps running in the background and after visiting another page and coming back to this, a new polling loop is started. I suggest we remove turbolinks. I could never make it work in anything other than a demo. @mssola @MaximilianMeister @ereslibre

mssola: We removed turbolinks in portus for other reasons (e.g. nginx couldn't pick it up in a containerized environment). Since we are not using it (and we might move into React or Vue), I'm ok with removing turbolinks.

mssola:
@jimmykarily thanks a lot 👏

PR #69, opened 2017-03-10
Update Dockerfile to also install phantomjs

jimmykarily:
because now we have js test that need it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,13 @@
 FROM ruby:2.1-onbuild
 
-RUN apt-get update && apt-get install -y nodejs
+RUN apt-get update && apt-get install -y nodejs fontconfig
+
+ARG PHANTOM_JS="phantomjs-2.1.1-linux-x86_64"
+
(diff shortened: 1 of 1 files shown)


jimmykarily:
This grows the image up to 1.058Gb but Debian Jessie (on which the ruby onbuild image is based) doesn't provide a package for phantomjs. I think this is the best way to get it.

jimmykarily:
This grows the image up to 1.058Gb but Debian Jessie (on which the ruby onbuild image is based) doesn't provide a package for phantomjs. I think this is the best way to get it.

Dockerfile
```diff
@@ -1,5 +1,13 @@
 FROM ruby:2.1-onbuild
 
-RUN apt-get update && apt-get install -y nodejs
+RUN apt-get update && apt-get install -y nodejs fontconfig
+
+ARG PHANTOM_JS="phantomjs-2.1.1-linux-x86_64"
+
+RUN curl  https://bitbucket.org/ariya/phantomjs/downloads/$PHANTOM_JS.tar.bz2 -O -L
```
jimmykarily: Link taken from here: http://phantomjs.org/download.html

PR #71, opened 2017-03-10
update Gemfile.lock

MaximilianMeister:
in our tests there was an error:
  Bundler::GemNotFound: Your bundle is locked to i18n (0.8.0), but that
version could not be found in any of the sources listed in your Gemfile

## NOTE
not tested yet


MaximilianMeister:
as we need to prepare on the packaging side this doesn't make sense now

PR #73, opened 2017-03-13
packaging: remove .orig files

MaximilianMeister:
unfortunately there is no option to patch that prevents this
it seems that patch creates an .orig file in some cases even
when it was successful, e.g. when the patch has an offset


jordimassaguerpla:
lgtm

PR #74, opened 2017-03-14
Show message about unassigned minions in monitoring page

jimmykarily:


jimmykarily:
With message:
selection_001

Without message:
selection_002

spec/features/monitoring_feature_spec.rb
```diff
@@ -18,4 +18,22 @@
       expect(page).to have_selector(".nodes-container tbody tr i.fa.fa-refresh")
     end
   end
+
+  # rubocop:disable RSpec/ExampleLength
+  # rubocop:disable RSpec/MultipleExpectations:
```
jimmykarily: I can avoid the many lines and many expectations if I remove the first 2 "expects". In that case the test is prone to false positives. E.g. if the flash message is already there before I create the minion.

mssola: That's fine I'd say

app/assets/javascripts/dashboard.js
```diff
@@ -27,6 +27,16 @@ MinionPoller = {
           }
         }
         $(".nodes-container tbody").html(rendered);
+
+        var unassignedMinions = data.unassigned_minions;
+
+        if (unassignedMinions.length > 0) {
+          $('#unassigned_count').html(unassignedMinions.length + " \
+            <strong>new</strong> nodes are available but have not been added to the cluster yet"
```
mssola: You should handle also the singular case 😉 (1 node vs 1 nodes)

app/views/dashboard/index.html.slim
```diff
@@ -1,7 +1,11 @@
 h1 Simple Monitoring
 .row
   .col-xs-12.discovery-control
-    span #{Minion.count} Nodes found
+    span#unassigned_count
+      - if (unassigned = @unassigned_minions.count) > 0
+        = unassigned
+        strong  new
+        | nodes are available but have not been added to the cluster yet
```
mssola: Same here

app/assets/javascripts/dashboard.js
jimmykarily: It is not that simple: 

- **1** new **node** **is** available but *has* not been...
- **2** new **nodes** **are** available but **have** not been...

I don't think it is worth it. This text will probably be gone soon.

ereslibre: With rails `I18n` it's really simple to do this (with `one` and `other` fields), but we are ignoring `I18n` for now.

PR #75, opened 2017-03-15
Change "nodes" to "workers"

jimmykarily:
"workers" is the term we decided to use for salt minions.


ereslibre:
Hmm, is this right? I thought `nodes` included `master` (`kubernetes-master`) and `worker` (`kubernetes-worker` -- `kubelet`, container runtime...) roles. While `workers` only referred to worker (kubernetes-worker) role.

So, all `workers` are also `nodes`, but not all `nodes` are `workers`. I thought we were going in this direction, but it would be great to clarify.

PR #72, opened 2017-03-13
Add instructions on how to spawn workers

jimmykarily:

PR #77, opened 2017-03-15
Add AutoYaST profile serving

ereslibre:
This profile will be read by workers during the bootstrapping process
and when adding new workers to the cluster. It will provide the
subscription details of the main installation to the workers.

Here we take into account that we are using our SUSE Secrets fork of
Docker, that mounts `/run/secrets` on every container. This directory
will include `/etc/SUSEConnect` and `/etc/zypp/credentials.d` from the
host mounted inside the container.

- `/run/secrets/SUSEConnect` includes the SMT if any (the default is SCC
  if not provided). If there is a SMT present, the regcode will be empty.

- If we are using SUSE's Customer Center (SCC), we need a regcode, that
  is obtained using the credentials in
  `/run/secrets/credentials.d/SCCcredentials` file, accessing the SCC API
   and checking for an active subscription. Filtering the CaaSP active
   subscription we can fetch the registration code for the subscription
   and serve it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
   - bundle exec rspec spec
   - bundle exec rubocop -V
   - bundle exec rubocop -F
-  - bundle exec brakeman -z -A
+  - bundle exec brakeman -c config/brakeman.ignore -z -A
 
 addons:
   code_climate:
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
 require "velum/kubernetes"
+require "velum/suse_connect"
 
 # DashboardController shows the main page.
 class DashboardController < ApplicationController
+  # TODO: move autoyast to its own controller (following a different logic flow). It would never get
+  # authenticated (as login/password -- since it's machines requesting this endpoint). It would
+  # never get redirected to setup the cluster, and it should actually read some security setting for
+  # only serving the autoyast profile to a set of IP ranges (provided by the customer).
+  skip_before_action :authenticate_user!, only: :autoyast
+  skip_before_action :redirect_to_setup, only: :autoyast
+
   def index
     @assigned_minions = Minion.assigned_role
     @unassigned_minions = Minion.unassigned_role
@@ -17,6 +25,39 @@ class DashboardController < ApplicationController
     end
   end
 
+  # Return the autoyast XML profile to bootstrap other worker nodes. They will read this response in
+  # order to start an unattended installation of CaaSP.
+  #
+  # It will return the content of the autoyast profile along with a 200 HTTP response code if the
+  # operation was successfull, or a 503 HTTP response code (service unavailable) if there was any
+  # problem while contacting to the subscription backend.
+  #
+  # This method skips authentication (workers won't authenticate using the typical username/password
+  # fields that customer uses) and also skips the redirection to the setup process (when a worker
+  # asks for the autoyast profile we will either serve it, or return an error).
+  def autoyast
+    @controller_node = Socket.gethostname
+    begin
+      suse_connect_config = Rails.cache.fetch("SUSEConnect_config") do
+        Velum::SUSEConnect.config
+      end
+      @suse_smt_url = suse_connect_config.smt_url
+      @suse_regcode = suse_connect_config.regcode
+      @do_registration = true
+    rescue Velum::SUSEConnect::MissingRegCodeException,
+           Velum::SUSEConnect::MissingCredentialsException
+      @do_registration = false
+    end
+    render "autoyast.xml.erb", layout: false, content_type: "text/xml"
+  rescue Velum::SUSEConnect::SCCConnectionException
+    head :service_unavailable
+  end
+
+  # Return the kubeconfig file that allows the customer to use the cluster using the kubectl tool.
+  #
+  # If everything is successfull, the kubeconfig file will be served. Otherwise (e.g. the cluster is
+  # not yet provisioned or is still being provisioned) it will redirect to the main page giving the
+  # user feedback about the current situation.
   def kubectl_config
     kubeconfig = Velum::Kubernetes.kubeconfig
     @apiserver_host = kubeconfig.host
--- a/app/views/setup/worker_bootstrap.html.slim
+++ b/app/views/setup/worker_bootstrap.html.slim
@@ -12,20 +12,20 @@ h1 Bootstrap your Container as a Service Platform
 p
   | In order to complete the installation, it is necessary to bootstrap a few additional nodes, those will be the Kubernetes Master and Workers.
   |  This process leverages AutoYaST and is (almost) fully automated.
-  |  In case you are not familiar with it, you can find more information about AutoYaST in the 
+  |  In case you are not familiar with it, you can find more information about AutoYaST in the&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html" official documentation.
-  |  The automatic installation gets invoked by adding autoyast=&lt;URL_TO_PROFILE&gt; to the kernel parameter list.
+  |  The automatic installation gets invoked by adding autoyast=#{autoyast_url} to the kernel parameter list.
   |  As installation media, you can use the very same image you bootstrapped the admin node with.
 
   |  A ready to use AutoYaST profile has been already generated for you during the bootstrap of the admin node.
   |  Bootstrap all the nodes you want to make part of this platform by adding the following boot parameter
-  |  autoyast=http://myserver/myconfig.xml
+  |  autoyast=#{autoyast_url}
 
 h2 Tips
 p
-  | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available 
+  | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#bootmedium.pxe" here
-  | . Where http://myserver/myconfig.xml is the real, generated path to the AutoYaST profile served by the dashboard.
+  | . Where #{autoyast_url} is the real, generated path to the AutoYaST profile served by the dashboard.
 
 .clearfix.text-right.steps-container
   = link_to "Back", setup_path, class: "btn btn-danger"
--- a/lib/velum/suse_connect.rb
+++ b/lib/velum/suse_connect.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+require "net/http"
+require "velum/http_exceptions"
+
+module Velum
+  # This class handles interaction with the SUSE Connect service (whether is the SCC or SMT).
+  class SUSEConnect
+    # Raised when there is a connection exception with the SMT/SCC service
+    class SCCConnectionException < StandardError; end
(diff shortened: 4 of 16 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -14,4 +16,16 @@ def index
       end
     end
   end
+
+  def autoyast_profile
+    @suse_register_email = Pillar.value pillar: "suse.register.email"
+    @suse_register_code = Pillar.value pillar: "suse.register.code"
+    @suse_register_server = Pillar.value pillar: "suse.register.server"
+    if @suse_register_email.blank? || @suse_register_code.blank? || @suse_register_server.blank?
+      render status: 503
```
jimmykarily: Maybe we should add validations for the first page of the setup. We don't want the user to skip these values.

ereslibre: There are two possibilities here that I'd like to know beforehand. I wrote this as a proof of concept, but:

1. If YaST asks for this information during the installation of the controller node, it will write this settings somewhere. We would need to mount this information in the velum container and wouldn't use the pillar at all.

2. YaST does not request this information during the installation. We do during the setup process. We write this information on the pillar and retrieve it from the pillar. In this case it's very important that this information is filled `before` other nodes are powered on (or they can try to get a not yet ready autoyast profile).

Depending on the option we would need to do one thing or the other.

flavio:
> Also, it isn't completely clear if we need to ask for the hostname on the setup process (and add it as a pillar) or if can count on YaST having set it always (so we can just ask for the hostname on the dashboard) -- and where would it have been placed it.

This is indeed something that has to be figured out. I would prefer the hostname to be set by the user during the YaST installation. I fear a change of the hostname at a later point would make things more complicated (eg: imagine we start to serve velum over https using an automatically generated certificate. A change of hostname would cause some headaches)

flavio:
Adding @eotchi to the thread, she has to be aware of these questions too

jimmykarily:
@ereslibre you might want to take care of this too: https://github.com/kubic-project/velum/pull/72/files#r106136028

ereslibre:
Sure, @jimmykarily. :+1:

ereslibre:
We actually don't need to template anything here, just serve a static file that is actually on the FS.

ereslibre:
Please, review again! :clap:

MaximilianMeister:
seems that the e2e tests were stuck at terraform applying (happens 1/100 times not sure why). i just retriggered them...

app/controllers/dashboard_controller.rb
```diff
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
 require "velum/kubernetes"
+require "velum/suse_connect"
 
 # DashboardController shows the main page.
 class DashboardController < ApplicationController
+  skip_before_action :authenticate_user!, only: :autoyast
```
mssola: Isn't this a bit dangerous ? The autoyast profile will contain sensitive data like the registration code. Isn't there a way to only serve this file to machines we know we can serve ? Just saying so things like the registration code don't leak into the outer world.

app/controllers/dashboard_controller.rb
```diff
@@ -17,6 +21,24 @@ def index
     end
   end
 
+  def autoyast
```
mssola: Document :)

.travis.yml
```diff
@@ -28,7 +28,7 @@ script:
   - bundle exec rspec spec
   - bundle exec rubocop -V
   - bundle exec rubocop -F
-  - bundle exec brakeman -z -A
+  - bundle exec brakeman -c config/brakeman.ignore -z -A
```
jimmykarily: Both commands seem to return the same output. Are you sure this is needed?

lib/velum/suse_connect.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+require "net/http"
+require "velum/http_exceptions"
+
+module Velum
+  # This class handles interaction with the SUSE Connect service (whether is the SCC or SMT).
+  class SUSEConnect
+    class SCCConnectionException < StandardError; end
```
mssola: Document the purpose of each exception

mssola: Also document methods and such when it makes sense.

app/controllers/dashboard_controller.rb
jimmykarily: As discussed in a previous meeting, this poses a security issue. Inside the autoyast file we have the @suse_regcode rendered and with no authentication/authorization in place anybody with network access to the dashboard can read the reg_code.

I think we should start planning for a solution. I think we can ignore this for now (beta1).

app/controllers/dashboard_controller.rb
```diff
@@ -17,6 +21,24 @@ def index
...
+      @suse_smt_url = suse_connect_config.smt_url
+      @suse_regcode = suse_connect_config.regcode
+      @do_registration = true
+    rescue Velum::SUSEConnect::MissingRegCodeException,
+           Velum::SUSEConnect::MissingCredentialsException
+      @do_registration = false
+    end
+    render "autoyast.xml.erb", layout: false, content_type: "text/xml"
+  rescue Velum::SUSEConnect::SCCConnectionException
+    render status: 503, nothing: true
```
jimmykarily: nothing: true [will be deprecated](http://stackoverflow.com/questions/34688726/the-nothing-option-is-deprecated-and-will-be-removed-in-rails-5-1). Maybe `head: 503` ?

.travis.yml
ereslibre: They do on your branch :) Here they make brakeman pass.

Brakeman fails because https://github.com/kubic-project/velum/pull/77/files#diff-b964d4ad50e36a0b0c7da1295058f697R112 is insecure. However, we already know that, so I had to create https://github.com/kubic-project/velum/pull/77/files#diff-1fb4db507da51ec0282450037ef38a47 in order for this warning to be omitted and have a green build.

I mean, we *know* insecure is insecure, that's why our flag is called `insecure` and is `false` by default. However, if someone wants to use an insecure SMT server (bad certificate or `http`) is their sole responsability.

app/controllers/dashboard_controller.rb
ereslibre: We cannot serve the autoyast profile to other workers then, they would get redirected to the login page :)

ereslibre: Yes, my guess is that we should add a restriction of what IP ranges can actually request this autoyast configuration, for example, but this is beyond the beta1 scope.

app/controllers/dashboard_controller.rb
MaximilianMeister: https://trello.com/c/4pDXbtUV/81-create-and-maintain-api-documentation-for-velum

so maybe we could do that for all endpoints at a later time

app/controllers/dashboard_controller.rb
mssola: Then, add a TODO comment 😉

app/views/setup/worker_bootstrap.html.slim
```diff
@@ -12,19 +12,19 @@ h1 Bootstrap your Container as a Service Platform
 p
   | In order to complete the installation, it is necessary to bootstrap a few additional nodes, those will be the Kubernetes Master and Workers.
   |  This process leverages AutoYaST and is (almost) fully automated.
-  |  In case you are not familiar with it, you can find more information about AutoYaST in the 
-  a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html" official documentation.
-  |  The automatic installation gets invoked by adding autoyast=&lt;URL_TO_PROFILE&gt; to the kernel parameter list.
+  |  In case you are not familiar with it, you can find more information about AutoYaST in the
+  a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html"  official documentation.
```
jimmykarily: [minor] extra space

app/views/setup/worker_bootstrap.html.slim
```diff
@@ -12,19 +12,19 @@ h1 Bootstrap your Container as a Service Platform
...
   |  Bootstrap all the nodes you want to make part of this platform by adding the following boot parameter
-  |  autoyast=http://myserver/myconfig.xml
+  |  autoyast=#{autoyast_url}
 
 h2 Tips
 p
-  | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available 
-  a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#bootmedium.pxe" here
+  | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available
+  a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#bootmedium.pxe"  here
```
jimmykarily: extra space

app/views/setup/worker_bootstrap.html.slim
ereslibre: Otherwise, the link appears next to the regular text :(

It's the first time I use slim, but this is the only way I found to show everything correctly.

app/views/setup/worker_bootstrap.html.slim
ereslibre: Otherwise, the link appears next to the regular text :(

It's the first time I use slim, but this is the only way I found to show everything correctly.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -59,6 +61,80 @@
...
+    end
+
+    context "if the credentials are missing for SMT or SCC" do
+      before do
+        allow(Velum::SUSEConnect).to receive(:config).and_raise(
+          Velum::SUSEConnect::MissingCredentialsException
+        )
+      end
+
+      it "serves the autoyast content" do
```
jimmykarily: Is there a specific case where an autoyast file with empty regcode is actually useful?

ereslibre: Yes, when using `SMT` (say `https://smt.mycompany.com` instead of `https://scc.suse.com`), the `regcode` will be empty, as that information is already in your local instance of `SMT` (that is the piece that will actually connect to `SCC` directly).

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -59,6 +61,80 @@
...
+    end
+
+    context "if there is a connectivity problem with SMT or SCC" do
+      before do
+        allow(Velum::SUSEConnect).to receive(:config).and_raise(
+          Velum::SUSEConnect::SCCConnectionException
+        )
+      end
+
+      it "renders a 503 status and a blank page" do
```
jimmykarily: So we care about connectivity but we don't care if we are authorized to talk to SMT or not?

ereslibre: I'd say so... I have never had access to a real `SMT` instance from where I could take some responses with VCR and play with it...

ereslibre: Actually, I *think* (not sure) that `SMT` is actually an instance of `SCC`, so API and stuff should be the same.

.travis.yml
jimmykarily: agreed, thanks

spec/controllers/dashboard_controller_spec.rb
jimmykarily: great, thanks

app/controllers/dashboard_controller.rb
ereslibre: Addressed.

PR #78, opened 2017-03-16
ui: fixes part2

MaximilianMeister:


app/views/devise/registrations/new.html.slim
```diff
@@ -18,7 +18,7 @@
       .form-group
         = f.email_field :email, autofocus: true, class: 'form-control', placeholder: 'Introduce your Email', required: true
       .form-group
-        = f.password_field :password, class: 'form-control', placeholder: 'Select your password (8 characters min.)', required: true
+        = f.password_field :password, class: 'form-control', pattern: "^[a-zA-Z0-9]{#{Devise.password_length.begin},#{Devise.password_length.end}}$", placeholder: "Select your password (#{Devise.password_length.begin} characters min.)", required: true
```
mssola: `[a-zA-Z0-9]` Why are other characters not allowed ?

MaximilianMeister: no specific reason, that's just a suggestion, should we restrict this or not?

mssola: Not in my opinion, but I'd like to hear from others.

MaximilianMeister: i think if there's no restriction from the back-end side, it's better to not restrict here either

jimmykarily:
Css is broken:

selection_004

jimmykarily:
Also I get no flash message at all in the registration page (e.g. when I enter 2 different passwords and submit).

app/views/devise/registrations/new.html.slim
ereslibre: I would not restrict password allowed symbols at all, I am all in for enforcing a minimum length, but characters should be any. Here is an example: http://html5pattern.com/Passwords (`(?=^.{8,}$)((?=.*\d)|(?=.*\W+))(?![.\n])(?=.*[A-Z])(?=.*[a-z]).*$`). Something similar to that?

MaximilianMeister:
> Css is broken

fixed

jimmykarily:
selection_006

PR #80, opened 2017-03-16
Provide kubectl configuration

ereslibre:
Provide a download option for the kubeconfig file.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "velum/kubernetes"
+
 # DashboardController shows the main page.
 class DashboardController < ApplicationController
   def index
@@ -12,6 +14,21 @@ class DashboardController < ApplicationController
         render json: { assigned_minions:   @assigned_minions,
                        unassigned_minions: @unassigned_minions }
       end
+    end
+  end
+
+  def kubectl_config
+    kubeconfig = Velum::Kubernetes.kubeconfig
+    @apiserver_host = kubeconfig.host
+    @ca_crt = kubeconfig.ca_crt
+    @client_crt = kubeconfig.client_crt
+    @client_key = kubeconfig.client_key
+    if [@apiserver_host, @ca_crt, @client_crt, @client_key].none?(&:blank?)
+      response.headers["Content-Disposition"] = "attachment; filename=kubeconfig"
+      render "kubeconfig.erb", layout: false, content_type: "text/yaml"
+    else
+      redirect_to root_path,
+                  alert: "Provisioning did not yet finish. Please wait until the cluster is ready."
     end
   end
 end
--- a/lib/velum/kubernetes.rb
+++ b/lib/velum/kubernetes.rb
@@ -1,8 +1,35 @@
 # frozen_string_literal: true
+
+require "velum/salt"
+
 module Velum
   # Kubernetes deals with the Kubernetes integration of this application.
   class Kubernetes
     attr_reader :client
+
+    KubeConfig = Struct.new :host, :ca_crt, :client_crt, :client_key
+
+    # Returns the Kubernetes apiserver configuration. It returns a KubeConfig struct.
+    def self.kubeconfig
+      _, ca = Velum::Salt.call action:  "mine.get",
+                               targets: "ca",
+                               arg:     ["ca", "x509.get_pem_entries"]
+      _, apiserver_crt = Velum::Salt.call action:      "cmd.run",
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         "cat /etc/pki/minion.crt"
+      _, apiserver_key = Velum::Salt.call action:      "cmd.run",
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         "cat /etc/pki/minion.key"
+      host = Minion.master.applied.pluck(:hostname).first
+      # rubocop:disable Style/RescueModifier
+      ca_crt = ca["return"].first["ca"]["ca"]["/etc/pki/ca.crt"] rescue nil
+      client_crt = apiserver_crt["return"].first.values.first rescue nil
+      client_key = apiserver_key["return"].first.values.first rescue nil
+      # rubocop:enable Style/RescueModifier
+      KubeConfig.new host, ca_crt, client_crt, client_key
+    end
 
     def initialize
       host = ENV["VELUM_KUBERNETES_HOST"]
--- a/lib/velum/salt.rb
+++ b/lib/velum/salt.rb
@@ -12,8 +12,8 @@ module Velum
     # Returns two values:
     #   - The response object.
     #   - A hash containing the parsed JSON response.
-    def self.call(action:, targets: "*", arg: nil)
-      hsh = { tgt: targets, fun: action, client: "local" }
+    def self.call(action:, targets: "*", target_type: "glob", arg: nil)
+      hsh = { tgt: targets, fun: action, expr_form: target_type, client: "local" }
       hsh[:arg] = arg if arg
 
       res = perform_request(endpoint: "/", method: "post", data: hsh)
--- a/spec/controllers/dashboard_controller_spec.rb
+++ b/spec/controllers/dashboard_controller_spec.rb
@@ -2,11 +2,15 @@
 require "rails_helper"
 
 RSpec.describe DashboardController, type: :controller do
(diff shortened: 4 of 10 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -14,4 +16,22 @@ def index
       end
     end
   end
+
+  def kubectl_config
```
mssola: I'd try to put as much of this code as possible into a lib (for example in `lib/velum/kubernetes` ?). Then this method should be:

1. Get values from library and set them into instance variables.
2. Render the thing

lib/velum/salt.rb
```diff
@@ -12,8 +12,8 @@ class Salt
     # Returns two values:
     #   - The response object.
     #   - A hash containing the parsed JSON response.
-    def self.call(action:, targets: "*", arg: nil)
-      hsh = { tgt: targets, fun: action, client: "local" }
+    def self.call(action:, targets: "*", target_type: "glob", arg: nil)
```
mssola: Doesn't this break other calls ? Just wondering, haven't checked myself.

ereslibre: No, because by default `target_type` was already `glob`, and we were using it this way. We are extending the functionality to allow other `expr_type` here, but still using the default which is `glob`.

app/controllers/dashboard_controller.rb
PI-Victor: +1 functionality around kubectl might grow a lot.

ereslibre: Yes, it for sure should be refactorized. Problem is that this is `kubernetes` related, but it gets all the information from `salt`, so yeah...

ereslibre:
This is going through a refactor, will ask for review!

lib/velum/kubernetes.rb
```diff
@@ -1,9 +1,46 @@
...
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         ["cat /etc/pki/minion.crt"]
+      _, apiserver_key = Velum::Salt.call action:      "cmd.run",
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         ["cat /etc/pki/minion.key"]
+      host = Socket.gethostname
+      begin
+        ca_crt = ca["return"].first["ca"]["ca"]["/etc/pki/ca.crt"]
```
MaximilianMeister:
```ruby
ca_crt = ca["return"].first["ca"]["ca"]["/etc/pki/ca.crt"] rescue nil
```
should do the trick as well

ereslibre: Oh! Thanks!

ereslibre:
```
lib/velum/kubernetes.rb:26:16: C: Avoid using rescue in its modifier form.
      ca_crt = ca["return"].first["ca"]["ca"]["/etc/pki/ca.crt"] rescue nil
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

:crying_cat_face:

ereslibre:
Should be green now, please review this.

lib/velum/kubernetes.rb
```diff
@@ -1,9 +1,36 @@
...
+                               arg:     ["ca", "x509.get_pem_entries"]
+      _, apiserver_crt = Velum::Salt.call action:      "cmd.run",
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         ["cat /etc/pki/minion.crt"]
+      _, apiserver_key = Velum::Salt.call action:      "cmd.run",
+                                          targets:     "roles:kube-master",
+                                          target_type: "grain",
+                                          arg:         ["cat /etc/pki/minion.key"]
+      host = Socket.gethostname
```
jimmykarily: Here we use the hostname of the contoller node as the kubernetes API host. Below, in the initialize method, we use "VELUM_KUBERNETES_HOST" for the same thing, or am I confused?

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -52,4 +54,33 @@
...
     end
   end
+
+  describe "GET /kubectl-config" do
+    it "gets redirected if not logged in" do
+      get :kubectl_config
+      expect(response.status).to eq 302
+    end
+
+    context "HTML rendering" do
```
jimmykarily: [Minor] I think it would be better if we used something like "format.yaml do" to return the correct mime type since we are not rendering HTML after all.

lib/velum/kubernetes.rb
ereslibre: This is wrong actually, because the velum host does not match the kubernetes master host (I left it to change it afterwards and forgot about it, so thank you). We aren't setting `VELUM_KUBERNETES_HOST` because this is an envvar and it's kind of strange to modify our manifests (when we first bootstrap we don't even know what the host will be), so I guess we will drop this envvar at some point.

jimmykarily: That's what I thought. So the host should be the ip/hostname of the minion with the k8s-master role. I'm tried to change to this up manually but it doesn't seem to work:

```
The connection to the server 10.17.3.111:443 was refused - did you specify the right host or port?
```
Can you give a try too? Maybe I'm doing something wrong.

jimmykarily:
This is a good candidate for an end to end test.

spec/controllers/dashboard_controller_spec.rb
ereslibre: We are actually setting `content_type: "text/yaml"` on the controller, it's the only mime type that we will return when asked for the kubectl config (at least for now). Moreover, we force the content disposition to be an attachment (so we force the browser to show the download dialog).

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -2,11 +2,14 @@
...
-  let(:minion) { create(:minion) }
+  let(:user)          { create(:user) }
+  let(:minion1)       { create(:minion) }
+  let(:minion2)       { create(:minion) }
+  let(:master_minion) { create(:master_minion) }
+  let(:worker_minion) { create(:worker_minion) }
 
   before do
-    Minion.create! [{ hostname: "master" }, { hostname: "minion0" }]
+    minion1 && minion2
```
mssola: Would it make sense to simply `let!(:minion1)` and `let(:minion2)` above ? Otherwise doing `minion1 && minion2` is confusing for someone that hasn't done too much rspec.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -31,7 +34,7 @@
 
     it "shows a simple monitoring when roles have already been assigned" do
       sign_in user
-      Minion.assign_roles!(roles: { master: Minion.first.id, minion: Minion.all[1..-1].map(&:id) })
+      master_minion && worker_minion
```
mssola: Same

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -40,7 +43,7 @@
   describe "GET / via JSON" do
     before do
       sign_in user
-      Minion.assign_roles!(roles: { master: Minion.first.id, minion: Minion.all[1..-1].map(&:id) })
+      master_minion && worker_minion
```
mssola: Same

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -52,4 +55,33 @@
...
+
+      it "renders the kubeconfig file if the orchestration did finish" do
+        VCR.use_cassette("kubeconfig/cluster_ready", record: :none) do
+          get :kubectl_config
+          expect(response.status).to eq 200
+        end
+      end
+    end
+  end
+
```
mssola: Remove empty line

jimmykarily:
Good job!

ereslibre:
Building for green again!

spec/controllers/dashboard_controller_spec.rb
ereslibre: How would be more readable? The idea is that we `let()` as before, so we don't forcefully create it (some test cases test that there are nothing yet created), and we force them to be created with the factory depending on the context of the test `minion1 && minion2` or `master_minion && worker_minion` (the idea is to remove complexity in the test using `Minion.assign_roles!`, because all we care about is that minions have roles assigned).

mssola: Maybe add a simple comment then :)

PR #82, opened 2017-03-21
Fix MinionPoller when no nodes have been observed

mssola:

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -28,14 +28,13 @@ MinionPoller = {
         }
         $(".nodes-container tbody").html(rendered);
 
-        var unassignedMinions = data.unassigned_minions;
+        var unassigned = data.unassigned_minions;
 
-        if (unassignedMinions.length > 0) {
-          $('#unassigned_count').html(unassignedMinions.length + " \
-            <strong>new</strong> nodes are available but have not been added to the cluster yet"
-          );
-        } else {
+        if (!unassigned || unassigned.length === 0) {
           $('#unassigned_count').html('');
+        } else {
+          $('#unassigned_count').html(data.unassigned_minions.length + " \
+            <strong>new</strong> nodes are available but have not been added to the cluster yet");
         }
       }
     });
--- a/spec/features/monitoring_feature_spec.rb
+++ b/spec/features/monitoring_feature_spec.rb
@@ -10,6 +10,15 @@ feature "Monitoring feature" do
     visit authenticated_root_path
   end
 
+  scenario "It works if there are no minions", js: true do
+    Minion.destroy_all
+    visit authenticated_root_path
+
+    using_wait_time 10 do
+      expect(page).not_to have_content("minion0.k8s.local")
+    end
+  end
+
   scenario "It updates the status of the minions automatically", js: true do
     # We poll every 5 seconds so the default Capybara wait time might not be enough
     using_wait_time 10 do


jimmykarily:
Good catch!

Maybe we can add a test [here](https://github.com/kubic-project/velum/blob/master/spec/features/monitoring_feature_spec.rb). Other than that 👍 .

mssola:
Sure thing. I'll add a test.

app/assets/javascripts/dashboard.js
```diff
@@ -28,14 +28,13 @@ MinionPoller = {
...
-        if (unassignedMinions.length > 0) {
-          $('#unassigned_count').html(unassignedMinions.length + " \
-            <strong>new</strong> nodes are available but have not been added to the cluster yet"
-          );
-        } else {
+        if (!unassigned || unassigned.length === 0) {
           $('#unassigned_count').html('');
+        } else {
+          $('#unassigned_count').html(data.unassigned_minions.length + " \
+            <strong>new</strong> nodes are available but have not been added to the cluster yet");
```
ereslibre: Maybe this message could lead to think that there is a way to add them to the cluster? Maybe we could explicitly say that this cannot be done at this point yet.

jimmykarily: In the (near?) future they will be able to add them so I don't think its too important to change the message for now. In any case, if we change it, it should also be changed [here](https://github.com/kubic-project/velum/blob/master/app/views/dashboard/index.html.slim#L8) (reminder).

ereslibre: Yes, but this text is going to be shipped with the Beta1 :) We can update it right after we release it :)

spec/features/monitoring_feature_spec.rb
```diff
@@ -10,6 +10,15 @@
     visit authenticated_root_path
   end
 
+  scenario "It works if there are no minions", js: true do
```
jimmykarily: I think test would be green even without the changes in the code. The problem exists when the message is printed once but then all the Minions get assigned somehow. In that case the messages was never deleted. To check the correct case you first need to visit the page, then (optionally) verify message existence, assign the Minions and verify message removal.

mssola: I'm not sure I'm following... I got this error when I had no minions available, so this test should already cover that, no ?

jimmykarily: This shouldn't have happened. Both [ruby](https://github.com/kubic-project/velum/blob/master/app/views/dashboard/index.html.slim#L5) and [javascript](https://github.com/kubic-project/velum/pull/82/files#diff-398e9dc91ab41e7d183e081c72f8f161L33) only print the message if there are unassigned minions. I thought the problem was when the message was printed once (because unassigned minions exist) and after assigning roles to the minions (not possible through the UI yet), the message wasn't cleared. The latter case is valid although not possible without messing with the console.

spec/features/monitoring_feature_spec.rb
```diff
@@ -10,6 +10,15 @@
...
+    Minion.destroy_all
+    visit authenticated_root_path
+
+    using_wait_time 10 do
+      expect(page).not_to have_content("minion0.k8s.local")
```
jimmykarily: Since we only care about JS errors not being raised, consider using a more generic assertion and changing the test description to something like:

"javascript doesn raise and error when there are no minions [fixed bug]"

MaximilianMeister:
@mssola can you rebase? let's get it merged

PI-Victor:
@mssola can you rebase this, so we can merge it if it's ok?

mssola:
Closing in favor #136.

PR #83, opened 2017-03-21
Generate application base secrets upon startup, and use previously generated base secrets if they exist.

HouzuoGuo:
The following comments from Rafael are addressed:
- "We have inconsistent quotes overall our manifests, but I think absolute paths are always unquoted. Still we would need to improve consistency overall."
- "Here I'd say it's better to use SecureRandom (you need to require it at the top). SecureRandom.hex 64 would do the trick and it's recommended for this kind of purposes."

One comment from Rafael is to be discussed: `test.rb`

"If the `RAILS_ENV` is `test`, we do want to set this envvar? Will there be anything to read at all?"

### DIFF
--- a/lib/velum/file_config.rb
+++ b/lib/velum/file_config.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+require "json"
+require "fileutils"
+require "securerandom"
+
+module Velum
+  # Handle IO operations and data serialisation for configuration files under a specific directory.
+  class FileConfig
+    def initialize(base_dir)
+      @base_dir = base_dir
+    end
+
+    # Return true only if the specified file name exists in the configuration base directory.
+    def exist?(file_name)
+      File.exist?(File.join(@base_dir, file_name))
+    end
+
+    # Deserialise JSON text from the specified file into a hash structure and return.
+    def read_hash(file_name)
+      JSON.parse(IO.read(File.join(@base_dir, file_name)))
+    end
+
+    # Serialise the hash structure into JSON text and overwrite the specified file with it.
+    def write_hash(file_name, hash)
+      FileUtils.mkdir_p(@base_dir)
+      IO.write(File.join(@base_dir, file_name), JSON.dump(hash))
+    end
+  end
+
+  # Manage persistent application-wide secret data.
+  class AppSecrets
+    def initialize(base_dir)
+      @files = FileConfig.new(base_dir)
+    end
+
+    # Generate a new random "secret_key_base" value and return.
+    def new_secret_key_base
+      SecureRandom.hex(64)
+    end
+
+    # Return persisted "secret_key_base", or save a new one if persistent value does not exist.
+    def generate_secret_key_base
+      name = "app_secrets.json"
+      return @files.read_hash(name)["base"] if @files.exist?(name)
+      new_base = new_secret_key_base
+      @files.write_hash(name, "base" => new_base)
+      new_base
+    end
+  end
+end
--- a/spec/lib/velum/file_config_spec.rb
+++ b/spec/lib/velum/file_config_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+require "tempfile"
+require 'fileutils'
+require "velum/file_config"
+
+describe Velum::FileConfig do
+  conf_dir = File.join(Dir.tmpdir, "velum_config_test")
+
+  before do
+    FileUtils::rm_r(conf_dir, :force => true)
+  end
+
+  after do
+    FileUtils::rm_r(conf_dir, :force => true)
+  end
+
+  it "reads and writes hash structures" do
+    conf = Velum::FileConfig.new(conf_dir)
+    expect(conf.exist?("hash")).to eq false
+
+    conf.write_hash("hash", "a"=>1)
+    expect(conf.exist?("hash")).to eq true
+    expect(conf.read_hash("hash")).to eq("a"=>1)
+  end
+end
+
+describe Velum::AppSecrets do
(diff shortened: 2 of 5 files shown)


lib/velum/file_config.rb
```diff
@@ -0,0 +1,54 @@
...
+
+module Velum
+  # Handle IO operations and data serialisation for configuration files under a specific directory.
+  class FileConfig
+    def initialize(base_dir)
+      @base_dir = base_dir
+    end
+
+    # Return true only if the specified file name exists in the configuration base directory.
+    def exists?(file_name)
```
MaximilianMeister: I'm not sure why we need to always pass the `file_name` to all instance methods instead of just making it an instance attribute.
```ruby
def initialize
  @secrets_path = Pathname.new("#{ENV.fetch("VELUM_SECRETS_DIR")}/app_secret.json")
end
```
have a look at the `pathname` gem. then you also don't have to reimplement the `exist?` method.
and then you can do all IO operations much easier, e.g
```ruby
unless @secrets_path.exist?
  @secrets_path.open("w") { |f| f.puts JSON.dump(base: SecureRandom.random_bytes(64).unpack('H*').first }
end

return JSON.parse(@secrets_path.read)[:base]
```

lib/velum/file_config.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    # Serialise the hash structure into JSON text and overwrite the specified file with it.
+    def write_hash(file_name, hash)
+      FileUtils::mkdir_p(@base_dir)
+      IO.write(File.join(@base_dir, file_name), JSON.dump(hash))
+    end
+  end
+
+  # Manage persistent application-wide secret data.
+  class AppSecrets
```
MaximilianMeister: as the app secrets are in the center of that feature why not call the library sth like `app_secrets.rb`? `file_config.rb` doesn't say much

lib/velum/file_config.rb
```diff
@@ -0,0 +1,54 @@
...
+    def initialize(base_dir)
+      @files = FileConfig.new(base_dir)
+    end
+
+    # Generate a new random "secret_key_base" value and return.
+    def new_secret_key_base()
+      SecureRandom.random_bytes(64).unpack('H*')[0]
```
jimmykarily: better? ```SecureRandrom.hex(64)```

spec/lib/velum/file_config_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "reads and writes hash structures" do
+    conf = Velum::FileConfig.new(conf_dir)
+    expect(conf.exists?("hash")).to eq false
+
+    conf.write_hash("hash", {"a"=>1})
+    expect(conf.exists?("hash")).to eq true
+    expect(conf.read_hash("hash")).to eq({"a"=>1})
+  end
+
+  it "generates application key in hex string" do
```
jimmykarily: Description doesn't really match what the test does. Maybe something like:

```generates a unique secret```. I would also wrap this in a "describe #new_secret_key_base" block ([Minor]).

**Comment by mssola** [comment]
I'm not entirely sure about all this. We could simply use kubernetes secrets that sets the key base environment variable. If you really want to generate these secrets on the fly, maybe you could consider a similar (and way simpler) approach like [gitlab does](https://github.com/gitlabhq/gitlabhq/blob/master/config/initializers/secret_token.rb) ? But I'm still not sure why we'd want to do this in the first place.

**Comment by ereslibre** [comment]
We still have to think about other kind of secrets. We have several (different) secrets to protect, and I'm talking in general here (not related to this PR necessarily).

- `salt-api` username/password (`saltapi:saltapi` ATM).
- `mysql` credentials
  - do not use `root`
  - generate random username/password and write it somewhere. Give permissions to it to only required hosts (I think `username_here@localhost` would be enough)
- Rails `SECRET_KEY_BASE`

Each of these settings need to have been autogenerated in a safe way for every installation.

We also need to play nice with updates, this basically removes the possibility of modifying the manifests (we would need to do tricky stuff to merge contents when updating them, I'd discard that possibility).

I'll write some options that I think we have:

1. Generate `salt-api` and `mysql` secrets on some script, `rails` secrets in the rails app: this splits the secret generation in two phases. In this case we are using an initializer to load this information into memory `Rails.application.secrets.secret_key_base`.

2. Write a script that generates all secrets (if not yet created), write them to disk (one file or several files) [only if the files didn't exist]. Our kubernetes manifest that brings velum up would in reality contain a wrapper that will source this files. This file (or files if you want to split contents, which I believe is good) would contain something like:

```
MYSQL_USERNAME=val1
MYSQL_PASSWORD=val2
RAILS_SECRET_KEY_BASE=val3
SALTAPI_USER=val4
SALTAPI_PASSWORD=val5
```

where `val1`..`val5` have been autogenerated by this very same script.

This script would not only write this (or these) files. In the case of mysql would ensure that the username and password exists, with the correct grants, as well as the saltapi user and password, and ultimately, would fire velum with a "rails server..." as we do right now to bring it up.

We would still use envvars for development & production and we wouldn't have differences between the development and production environments.

3. A variant of the 2) option is to write the contents to files and in our Velum source code do (this is pseudocode) `File.read("/some/settings")["RAILS_SECRET_KEY_BASE"] || ENV["RAILS_SECRET_KEY_BASE"]`. This has the downside of setup values coming from different directions, and could make debugging harder.

ereslibre:
This can be used along with https://github.com/kubic-project/caasp-container-manifests/pull/4 for our beta1 version, so we have `SECRET_KEY_BASE` autogenerated.

HouzuoGuo:
After a discussion with Rafael, we are now determined to take a very direct approach in this task.

PR #84, opened 2017-03-23
Fixed the production environment

mssola:
It contained several things from the Rails 5 -> 4 transition.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,10 @@ gem "jbuilder", "~> 2.5"
 gem "jquery-rails"
 gem "slim"
 gem "font-awesome-rails"
+
+# NOTE: this is no longer needed in Rails 5. See
+# https://github.com/heroku/rails_stdout_logging#rails-5 for instructions on how
+# to transition.
 gem "rails_stdout_logging", group: [:development, :staging, :production]
 
 gem "bcrypt", "~> 3.1.7"
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -24,7 +24,11 @@ Rails.application.configure do
   # config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
+  config.assets.compile = true
+
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = true
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.action_controller.asset_host = 'http://assets.example.com'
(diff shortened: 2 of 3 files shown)


config/environments/production.rb
```diff
@@ -24,7 +24,11 @@
   # config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
+  config.assets.compile = true
```
ereslibre: Wow, do we compile assets on production for every request? This adds a lot of overhead...

Gemfile
```diff
@@ -8,6 +8,10 @@ gem "jbuilder", "~> 2.5"
 gem "jquery-rails"
 gem "slim"
 gem "font-awesome-rails"
+
+# NOTE: this is no longer needed in Rails 5. See
+# https://github.com/heroku/rails_stdout_logging#rails-5 for instructions on how
+# to transition.
 gem "rails_stdout_logging", group: [:development, :staging, :production]
```
ereslibre: I just saw that we have no `staging` group. Thought we had it set up but it seems a leftover from a copy&paste from Pharos, no? -- sorry, didn't catch it when we actually included this.

kubernetes/manifest-templates/velum.yaml
```diff
@@ -14,6 +14,8 @@ spec:
       value: development
     - name: SECRET_KEY_BASE
       value: fc0ece3811c0b1e0118dd7d619b385955e3ed146b37315d67d1c2d201c42531da398941ef27aab42493fff23afcba540390a26db51b5889e96d850a14178aabdc1926c013936a3c41abdbeb696ce440685c5d7be3a6c548434699d526a9e37657a76a43ca975b824905fd949f7e8d50aa1196a58e7e79dcfb08e2e397ee4f7b7
+    - name: RAILS_SERVE_STATIC_FILES
```
davidcassany: I believe that makes sense to be part of the image configuration. This is way it lives together with other ruby and rails configuration variables. I am concerned about how this kind of information is splitted and sometimes duplicated in multiple places.

Thoughts?

config/environments/production.rb
mssola: This is only applied if the asset was not found. Maybe we don't need it, but I just copied what we do in Portus 😄

Gemfile
mssola: It shouldn't be a problem for now. But we should have a staging environment indeed.

config/environments/production.rb
jordimassaguerpla: we shouldn't because we don't have a js engine in production ... but I don't know if that is something required for puma to work...@miquel: is it?

ereslibre: Right, I didn't remember this was when an asset is not found (if `false` you raise an error, if `true` you pass through the pipeline). I'd say it's ok as it is. Thank you!

mssola: > we shouldn't because we don't have a js engine in production ... but I don't know if that is something required for puma to work...@miquel: is it?

That's correct for sass I'd say, since we don't include sass related stuff. For JS this should work, because there's no JS engine dependency. So, if the thing missing is stylesheet, then we'll get the "no sass gem could be found" (which is fine I'd say), and if it's JS then we'll hit a file in app/assets/javascripts.

@jordimassaguerpla  Either way we get the expected behavior.

PR #88, opened 2017-03-23
setup: don't allow empty initial configuration

MaximilianMeister:

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -11,6 +11,7 @@ class SetupController < ApplicationController
 
   skip_before_action :redirect_to_setup
   before_action :redirect_to_dashboard
+  before_action :check_empty_settings, only: :configure
 
   def configure
     status = {}
@@ -27,7 +28,7 @@ class SetupController < ApplicationController
       if status[:failed_pillar]
         msg = "Failed to apply configuration to #{status[:failed_pillar].join(", ")}"
         format.html do
-          flash[:error] = msg
+          flash[:alert] = msg
           redirect_to setup_worker_bootstrap_path
         end
         format.json { render json: msg, status: :unprocessable_entity }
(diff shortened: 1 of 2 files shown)


app/controllers/setup_controller.rb
```diff
@@ -93,4 +94,15 @@ def minion_not_found(exception)
       format.json { render json: exception.message, status: :not_found }
     end
   end
+
+  def check_empty_settings
+    return unless settings_params.values.any?(&:empty?)
```
jimmykarily: Not sure if it makes sense for such a simple case but if we need more validations like this one, we could introduce Form objects (e.g. [point 3 here](http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/)).

ereslibre: Yes, I agree. With form objects we will be able to mark wrong fields in an independient fashion, however this will be tricky because all of them are mapped to the `Pillar` model in reality, whose `value` semantics and validations highly depend on the `pillar` key (`country` can only accept one value from a list, an `email` needs to pass a regexp...). Maybe we could add some facade models that actually use `Pillar` as the real backend but they are only for making it easier to use form objects...

spec/controllers/setup_controller_spec.rb
```diff
@@ -163,14 +163,25 @@
   end
 
   describe "PUT /setup via HTML" do
+    let(:settings_params) do
+      {
+        company_name: "SUSE Linux GmbH",
+        company_unit: "Research and development",
+        email:        "containers@suse.de",
+        country:      "Germany",
```
ereslibre: Country is actually the Alpha-2 code of the country. DE in this case. :)

app/controllers/setup_controller.rb
jimmykarily: No need to make the ActiveRecord models (so storage isn't relevant). Form objects can be plain old simple Ruby classes that validate the various fields. I think there are some good gems out there to help. Anyway, lets see this at a later time if we get more of this.

spec/controllers/setup_controller_spec.rb
MaximilianMeister: thanks, fixed

MaximilianMeister:
> LGTM even though I'd advise to do this on the client side too.

@mssola addressed in https://github.com/kubic-project/velum/pull/92

mssola:
> @mssola addressed in #92

Not all the fields right ? You could use the [require](https://www.w3schools.com/tags/att_input_required.asp) HTML attribute 😉

PR #89, opened 2017-03-23
add init for containers

jordimassaguerpla:
add a wrapper script called init that will check if the database has
been configured and, in that case, it will configure it before starting
the rails server.

### DIFF
--- a/bin/check_db.rb
+++ b/bin/check_db.rb
@@ -1,0 +1,22 @@
+# This is a rails runner that will print the status of Portus' database
+# Possible outcomes:
+#
+#   * `DB_READY`: the database has been created and initialized
+#   * `DB_EMPTY`: the database has been created but has not been initialized
+#   * `DB_MISSING`: the database has not been created
+#   * `DB_DOWN`: cannot connect to the database
+
+def database_exists?
+  ActiveRecord::Base.connection
+  if ActiveRecord::Base.connection.table_exists? 'schema_migrations'
+    puts "DB_READY"
+  else
+    puts "DB_EMPTY"
+  end
+rescue ActiveRecord::NoDatabaseError
+  puts "DB_MISSING"
+rescue Mysql2::Error
+  puts "DB_DOWN"
+end
+
+database_exists?
--- a/bin/check_db.rb
+++ b/bin/check_db.rb
@@ -1,4 +1,4 @@
-# This is a rails runner that will print the status of Portus' database
+# This is a rails runner that will print the status of the database
 # Possible outcomes:
 #
 #   * `DB_READY`: the database has been created and initialized
--- a/bin/init
+++ b/bin/init
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# This script will ensure Portus' database is ready to be used. It will keep
+# waiting for the db to be usable, but the script will exit with an error
+# after a certain amount of failed attempts.
+#
+# The script will automatically import all the SSL certificates from
+# `/certificates` into the final system. This is needed to talk with the
+# registry API when this one is protected by TLS.
+#
+# Finally the script will start apache running Portus via mod_rails.
+
+set -e
+# check that BUNDLER_BIN is set
+if [ ! -z $BUNDLER_BIN ]; then
+    echo "BUNDLER_BIN var is not set. Please set that and try again" >&2
+    exit -1
+fi
+
+setup_database() {
+  set +e
+
+  TIMEOUT=90
+  COUNT=0
+  RETRY=1
+
+  export SKIP_MIGRATION=1
+
+  while [ $RETRY -ne 0 ]; do
+    case $(SKIP_MIGRATION=1 $BUNDLER_BIN exec rails r /check_db.rb | grep DB) in
+      "DB_DOWN")
+        if [ "$COUNT" -ge "$TIMEOUT" ]; then
+          printf " [FAIL]\n"
+          echo "Timeout reached, exiting with error"
+          exit 1
+        fi
+        echo "Waiting for mariadb to be ready in 5 seconds"
+        sleep 5
+        COUNT=$((COUNT+5))
+        ;;
+      "DB_EMPTY"|"DB_MISSING")
+        # create db, apply schema and seed
+        echo "Initializing database"
+        SKIP_MIGRATION=1 $BUNDLER_BIN exec rake db:create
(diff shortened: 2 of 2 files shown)


bin/init
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+# This script will ensure Portus' database is ready to be used. It will keep
```
MaximilianMeister: `s/Portus/velum`

bin/init
```diff
@@ -0,0 +1,63 @@
...
+
+set -e
+# check that BUNDLER_BIN is set
+if [ ! -z $BUNDLER_BIN ]; then
+    echo "BUNDLER_BIN var is not set. Please set that and try again" >&2
+    exit -1
+fi
+
+setup_database() {
+  set +e
```
MaximilianMeister: (i think) it should be enough to have that statement once (as you already have above)

bin/check_db.rb
```diff
@@ -0,0 +1,22 @@
+# This is a rails runner that will print the status of Portus' database
```
MaximilianMeister: `s/Portus/velum`

bin/init
```diff
@@ -0,0 +1,63 @@
...
+
+# This script will ensure Portus' database is ready to be used. It will keep
+# waiting for the db to be usable, but the script will exit with an error
+# after a certain amount of failed attempts.
+#
+# The script will automatically import all the SSL certificates from
+# `/certificates` into the final system. This is needed to talk with the
+# registry API when this one is protected by TLS.
+#
+# Finally the script will start apache running Portus via mod_rails.
```
MaximilianMeister: `s/Portus/velum`

bin/init
```diff
@@ -0,0 +1,63 @@
...
+        ;;
+    esac
+  done
+  set -e
+}
+
+cd /srv/velum
+setup_database
+$BUNDLER_BIN exec rake db:migrate
+$BUNDLER_BIN exec "rails s -b 0.0.0.0 -p 3000 Puma"
```
ereslibre: Here I'd use `rails s -b 0.0.0.0 -p $PORT Puma` and let's set this `PORT` envvar in our manifests once and for all...

ereslibre: Or maybe just use `80`, and we'll use 80 for all environments...

bin/init
jordimassaguerpla: above is "-e" ;)

bin/init
jimmykarily: The script will not import SSL certificates and won't run apache. I think the last 2 paragraphs need to be removed.

jordimassaguerpla:
If you like this one, please check https://github.com/kubic-project/caasp-container-manifests/compare/fix_init?expand=1

bin/init
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# This script will setup the database and then start the rails application
+
+set -e
+# check that BUNDLER_BIN is set
+if [ ! -z $BUNDLER_BIN ]; then
```
jimmykarily: Who sets this?

bin/init
```diff
@@ -0,0 +1,60 @@
...
+    echo "ERROR: BUNDLER_BIN var is not set. Please set that and try again" >&2
+    exit -1
+fi
```
ereslibre: I would remove this block completely.

bin/init
```diff
@@ -0,0 +1,60 @@
...
+  set +e
+
+  TIMEOUT=90
+  COUNT=0
+  RETRY=1
+
+  export SKIP_MIGRATION=1
+
+  while [ $RETRY -ne 0 ]; do
+    case $(SKIP_MIGRATION=1 $BUNDLER_BIN exec rails r /check_db.rb | grep DB) in
```
ereslibre: We can directly call to `bundle` instead of using `$BUNDLER_BIN`. We aren't setting it and since this runs in a containerized context we know that `bundle` is on the `PATH`.

bin/init
```diff
@@ -0,0 +1,60 @@
...
+        fi
+        echo "Waiting for mariadb to be ready in 5 seconds"
+        sleep 5
+        COUNT=$((COUNT+5))
+        ;;
+      "DB_EMPTY"|"DB_MISSING")
+        # create db, apply schema and seed
+        echo "Initializing database"
+        SKIP_MIGRATION=1 $BUNDLER_BIN exec rake db:create
+        SKIP_MIGRATION=1 $BUNDLER_BIN exec rake db:schema:load
```
ereslibre: Same on these two lines.

bin/init
jordimassaguerpla: The image should have this pointing to bundler.

bin/init
```diff
@@ -0,0 +1,60 @@
...
+        ;;
+    esac
+  done
+  set -e
+}
+
+cd /srv/velum
+setup_database
+$BUNDLER_BIN exec rake db:migrate
+$BUNDLER_BIN exec "rails s -b 0.0.0.0 -p $PORT Puma"
```
ereslibre: And same here. Sorry for my previous confusing comment. I'd also change `$PORT` to `80`.

bin/init
ereslibre: But this restricts the usage to our `sle` based docker images. I would preferably use `bundler` directly since we know it exists, or that isn't the case on our `sle` based docker images? The idea is to be able to use the same init script for both the `sle` docker images and `opensuse` images.

bin/init
```diff
@@ -0,0 +1,50 @@
...
+  set +e
+
+  TIMEOUT=90
+  COUNT=0
+  RETRY=1
+
+  export SKIP_MIGRATION=1
+
+  while [ $RETRY -ne 0 ]; do
+    case $(SKIP_MIGRATION=1 bundle exec rails r /check_db.rb | grep DB) in
```
jimmykarily: SKIP_MIGRATION is not needed here I think

bin/init
```diff
@@ -0,0 +1,48 @@
...
+          printf " [FAIL]\n"
+          echo "Timeout reached, exiting with error"
+          exit 1
+        fi
+        echo "Waiting for mariadb to be ready in 5 seconds"
+        sleep 5
+        COUNT=$((COUNT+5))
+        ;;
+      "DB_EMPTY"|"DB_MISSING")
+        # create db, apply schema and seed
```
ereslibre: `#create db and apply schema` :)

jordimassaguerpla:
rebased in https://github.com/kubic-project/velum/pull/91

PR #91, opened 2017-03-23
add init for containers

jordimassaguerpla:
add a wrapper script called init that will check if the database has
been configured and, in that case, it will configure it before starting
the rails server.

This script calls db:migrate because this is meant to be run in a container.
When velum is updated, it will be by launching a new container, thus the new
container should run db:migrate.

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/bin/check_db.rb
+++ b/bin/check_db.rb
@@ -1,0 +1,22 @@
+# This is a rails runner that will print the status of the database
+# Possible outcomes:
+#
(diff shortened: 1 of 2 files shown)


jordimassaguerpla:
This is a rebase from https://github.com/kubic-project/velum/pull/89 which has been accepted.

PR #92, opened 2017-03-24
setup: validate fields in the front end

MaximilianMeister:


app/helpers/setup_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# SetupHelper contains all the setup view helpers.
+module SetupHelper
+  # rubocop:disable Metrics/LineLength
+  def country_code_regex
+    "^(AF|AX|AL|DZ|AS|AD|AO|AI|AQ|AG|AR|AM|AW|AU|AT|AZ|BS|BH|BD|BB|BY|BE|BZ|BJ|BM|BT|BO|BQ|BA|BW|BV|BR|IO|BN|BG|BF|BI|KH|CM|CA|CV|KY|CF|TD|CL|CN|CX|CC|CO|KM|CG|CD|CK|CR|CI|HR|CU|CW|CY|CZ|DK|DJ|DM|DO|EC|EG|SV|GQ|ER|EE|ET|FK|FO|FJ|FI|FR|GF|PF|TF|GA|GM|GE|DE|GH|GI|GR|GL|GD|GP|GU|GT|GG|GN|GW|GY|HT|HM|VA|HN|HK|HU|IS|IN|ID|IR|IQ|IE|IM|IL|IT|JM|JP|JE|JO|KZ|KE|KI|KP|KR|KW|KG|LA|LV|LB|LS|LR|LY|LI|LT|LU|MO|MK|MG|MW|MY|MV|ML|MT|MH|MQ|MR|MU|YT|MX|FM|MD|MC|MN|ME|MS|MA|MZ|MM|NA|NR|NP|NL|NC|NZ|NI|NE|NG|NU|NF|MP|NO|OM|PK|PW|PS|PA|PG|PY|PE|PH|PN|PL|PT|PR|QA|RE|RO|RU|RW|BL|SH|KN|LC|MF|PM|VC|WS|SM|ST|SA|SN|RS|SC|SL|SG|SX|SK|SI|SB|SO|ZA|GS|SS|ES|LK|SD|SR|SJ|SZ|SE|CH|SY|TW|TJ|TZ|TH|TL|TG|TK|TO|TT|TN|TR|TM|TC|TV|UG|UA|AE|GB|US|UM|UY|UZ|VU|VE|VN|VG|VI|WF|EH|YE|ZM|ZW)$".freeze
```
MaximilianMeister: I tried to make this a real regex `/^(AF..)$/` but converting it back to a string in the view adds a regex option notation in the string to make it convertible again.. so this didn't work unfortunately

see https://ruby-doc.org/core-2.1.1/Regexp.html#method-i-to_s

jordimassaguerpla: Could you add a link to where you got this codes from? I guess this is an iso or something... . Just as a comment, or even better in the method name.

jordimassaguerpla: could we have this super long line on a yaml file or something? So that we don't have to disable the rubocop metrics here.

MaximilianMeister: good idea, done!

mssola:
You are only validating some of them though. You could use the [required](https://www.w3schools.com/tags/att_input_required.asp) HTML attribute to force that users actually insert some data.

PR #94, opened 2017-03-24
Use the same init script for development env as we do in production

jimmykarily:
Our bin/init script already waits for MariaDB to come up. We no longer
need our setup_envoronment script. Now exited containers are not
cleaned but they will, the next time the "start" script will be called.


kubernetes/manifest-templates/velum.yaml
```diff
@@ -14,6 +14,8 @@ spec:
       value: development
     - name: RAILS_SERVE_STATIC_FILES
       value: "true"
+    - name: VELUM_PORT
+      value: "3000"
```
jimmykarily: Just to spawn the development server to the usual port.

ereslibre:
We'd need to add this init script to the opensuse based docker images, no?

jimmykarily:
No that script is [part of velum](https://github.com/kubic-project/velum/blob/master/bin/init)

jimmykarily:
Rebase but the e2e test are failing now:

```
error: failed to get network 'e2e_testsnet'
error: Network not found: no network with matching name 'e2e_testsnet'
```

the error doesn't seem related to the changes.

MaximilianMeister:
> the error doesn't seem related to the changes.

that comes from https://github.com/kubic-project/e2e-tests/commit/75aad8c74883296ab74ef2f792caa713d266fbbe and it is not causing the tests to fail. it's a bit misleading therefore i've also submitted https://github.com/kubic-project/e2e-tests/pull/28

the real error is `"message": "Unable to find field \"user_email\""` in the rspec output, do you have any idea?

jimmykarily:
No problem. Tell me if you need my to have a look.

MaximilianMeister:
> No problem. Tell me if you need my to have a look.

ok thanks, i just rebased, lets see what happens in the next test

MaximilianMeister:
@jimmykarily the only problem here is that the init script change is not there in the container we fetch from `opensuse/velum:development` so the app gets started on port 80, that's why the `user_email` field can't be found (on port `3000`)

i think there's not much we can do about that here to get the tests green, so i think this is good to merge as it is.

PR #95, opened 2017-03-24
Set salt master configuration on the minion on a script

ereslibre:
The files directive is executed on the 2nd YaST stage (which it's
skipped in CaaSP), so the `/etc/salt/minion.d/master.conf` was never
created on the worker nodes.

Let's do the same, but running a script that will create this
configuration file in the 1st YaST stage.


jimmykarily:
Is it possible that we add a comment there to explain why this hack is needed? If it is common knowledge forget it.

PR #99, opened 2017-03-27
setup: enable select field for the country codes

MaximilianMeister:
like that we can drop the front-end validation, and it's more convenient
for the user

### DIFF
--- a/app/helpers/setup_helper.rb
+++ b/app/helpers/setup_helper.rb
@@ -4,8 +4,8 @@ require "yaml"
 # SetupHelper contains all the setup view helpers.
 module SetupHelper
   # https://www.iso.org/iso-3166-country-codes.html
-  # taken from https://coderwall.com/p/xww5mq/two-letter-country-code-regex
-  def country_code_regex
-    YAML.load_file(Rails.root.join("config", "country_codes.yml"))["iso"][3166]["regex"]
+  def country_codes
+    YAML.load_file(Rails.root.join("config", "country_codes.yml"))["iso"][3166]["codes"]
+        .to_a.map(&:reverse)
   end
 end
--- a/config/country_codes.yml
+++ b/config/country_codes.yml
@@ -1,5 +1,254 @@
 ---
 iso:
   3166:
-    regex:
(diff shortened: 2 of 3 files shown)


MaximilianMeister:
this still needs some design input, and e2e-tests might also need an update. so please don't merge yet

jimmykarily:
Maybe we can use a gem that also defines the mappings to names:
E.g. https://github.com/alexrabarts/iso_country_codes

or we can get a list with mappings. E.g.: http://data.okfn.org/data/core/country-list

config/country_codes.yml
```diff
@@ -1,5 +1,253 @@
...
+    - "MY"
+    - "MZ"
+    - "NA"
+    - "NC"
+    - "NE"
+    - "NF"
+    - "NG"
+    - "NI"
+    - "NL"
+    - "NO"
```
MaximilianMeister: because of norway, i had to quote the yaml values, as `- NO` gets interpreted to `false` ;-)

MaximilianMeister:
> Maybe we can use a gem that also defines the mappings to names:
> E.g. https://github.com/alexrabarts/iso_country_codes

looks good, but that would mean a new dependency, which i think is not worth it, as the list of countries might not change so quickly

> 
> or we can get a list with mappings. E.g.: http://data.okfn.org/data/core/country-list

you mean always fetch them from there?

jimmykarily:
No I mean one-off (as we do now) but with country names as well. It would be better if we showed country names to the user rather than ISO codes.

MaximilianMeister:
> No I mean one-off (as we do now) but with country names as well. It would be better if we showed country names to the user rather than ISO codes.

sure, good idea. that makes it a bit more readable

config/country_codes.yml
```diff
@@ -1,5 +1,253 @@
 ---
 iso:
```
jimmykarily: Please add a comment with the source of  this information in case we need to update this file in the future:

Either [this](http://data.okfn.org/data/core/country-list) or [this](https://raw.githubusercontent.com/datasets/country-list/master/data.csv) link.

jimmykarily:
Please add a blank option to the drop down because now "Afghanistan" is selected by default (which looks like a default)

MaximilianMeister:
> Please add a blank option to the drop down because now "Afghanistan" is selected by default (which looks like a default)

oh ok, haven't noticed as i already set a value...

app/helpers/setup_helper.rb
```diff
@@ -4,8 +4,11 @@
 # SetupHelper contains all the setup view helpers.
 module SetupHelper
   # https://www.iso.org/iso-3166-country-codes.html
-  # taken from https://coderwall.com/p/xww5mq/two-letter-country-code-regex
-  def country_code_regex
-    YAML.load_file(Rails.root.join("config", "country_codes.yml"))["iso"][3166]["regex"]
+  def country_codes
+    YAML.load_file(
+      Rails.root.join("config", "country_codes.yml")
+    )["iso"][3166]["codes"].collect do |c|
```
jimmykarily: I think if you remove the dashes in front of the country codes in the yaml file, you will get a Hash instead of an Array of (single element) Hashes. Then you can simply call `to_a` here instead of `collect` to get the same result:

```
 YAML.load_file(Rails.root.join("config", "country_codes.yml")["iso"][3166]["codes"].to_a
```

MaximilianMeister: fixed, just had to reverse the elements. but that was a good hint :+1:

jimmykarily:
The tests are probably failing due to this: https://github.com/kubic-project/salt/pull/26/files
They should be ok now. Lets wait and see.

MaximilianMeister:
> The tests are probably failing due to this: https://github.com/kubic-project/salt/pull/26/files
> They should be ok now. Lets wait and see.

that was it. tests are green now

PR #100, opened 2017-03-27
packaging: rename patch

MaximilianMeister:
this prevents it from being applied during spec file creation
it needs to be applied during the obs build

see packaging guidelines


jordimassaguerpla:
thanks.
lgtm

PR #102, opened 2017-03-27
discovery: don't allow empty bootstrap

### LINKED ISSUE #64: Bootstrap cluster with no minions raises an error
When clicking the "Bootstrap cluster" button and there are no Minions this error gets raised:

selection_002

We shouldn't let the user click on a button that will raise an error. I suggest that we disable the button showing a tooltip explaining the reason why it is disabled (no minions present).

MaximilianMeister:
fixes #64


MaximilianMeister:
all green! I am sorry for merging without approval, but we need to get the fix in today

PR #103, opened 2017-03-28
bootstrap: don't allow empty bootstrap in the back-end

MaximilianMeister:
extending #64


MaximilianMeister:
let's build the package

PR #104, opened 2017-03-28
Bug 1031352 - Do not expose mariadb to the world

robdaemon:
Switch to a UNIX socket for communication with MariaDB. The UNIX socket
is in a Kubernetes emptyDir so that it can be used among the other
containers in the pod.

Turned on set -e in the bash scripts, and stopped outputting the
db:setup to /dev/null as part of diagnosing startup problems.

You can't seem to use ConfigMaps in disconnected scenarios, so the
MariaDB configuration that disables networking is mounted from a
hostDir.


robdaemon:
This is my first PR against a repo here at SUSE, so any feedback would be much appreciated :)

robdaemon:
I'll take a look at the other manifests and see if I need to do similar work there. Thanks!!

robdaemon:
Will do. I will ping you on IRC tomorrow, I'm having some trouble fully running this in a development environment.

robdaemon:
Okay, Salt cannot connect to mysql in this configuration (and it's fairly silent unless you specifically look at the Salt master log). Sounds like we have a few options:

- use hostPath for the UNIX socket
- move the Salt master into the velum pod so that we can use the emptyDir volume in this PR

Using hostPath is really no different from binding on port 3306 from a security perspective, though.

config/database.yml
```diff
@@ -1,12 +1,13 @@
 default: &default
   adapter:  "mysql2"
 
-  host:     <%= ENV['VELUM_DB_HOST']     || "localhost" %>
-  port:     <%= ENV['VELUM_DB_PORT']     || 0           %>
-  username: <%= ENV['VELUM_DB_USERNAME'] || "root"      %>
-  password: <%= ENV['VELUM_DB_PASSWORD'] || ""          %>
```
jimmykarily: I'm not sure why we introduced these extra spaces in the first place but full free to remove them. I don't remember any style rule for this one and they look weird.

ereslibre: I believe they were added to align them to the socket line at the bottom.

kubernetes/start
```diff
@@ -29,7 +30,7 @@ while (( $interactive_mode == 1)) ; do
   esac
 done
 
-./cleanup
+./cleanup || true
```
flavio: Are you sure you don't want to fail if the cleanup doesn't work as expected?

config/database.yml
```diff
@@ -1,12 +1,13 @@
...
+  host:     <%= ENV['VELUM_DB_HOST']     || ""                            %>
+  port:     <%= ENV['VELUM_DB_PORT']     || 0                             %>
+  username: <%= ENV['VELUM_DB_USERNAME'] || "root"                        %>
+  password: <%= ENV['VELUM_DB_PASSWORD'] || ""                            %>
   encoding: utf8
-  pool:     <%= ENV['VELUM_DB_POOL']     || 5           %>
+  pool:     <%= ENV['VELUM_DB_POOL']     || 5                             %>
+  socket:   <%= ENV['VELUM_DB_SOCKET']   || "/var/run/mysqld/mysqld.sock" %>
```
flavio: `VELUM_DB_SOCKET` will still point to the socket even when `VELUM_DB_HOST` is set.  Isn't this going to cause issues to Rails?

Wouldn't it be better to have something like:

```ruby
<% if ENV['VELUM_DB_SOCKET'] %>
socket:   <%= ENV['VELUM_DB_SOCKET'] %>
<% endif %>
```

ereslibre:
My only doubt is how to proxy salt-master. It's clear that on the `public` manifest (the one using `hostNetwork: true`) we will proxy `velum` (trivial HTTP proxy). I am not that sure about how to proxy `salt-master` service, since it also needs to be reachable by other salt-minions.

ereslibre:
I answer to myself: the public manifest could include the velum proxy and the salt-master itself, because this manifest would be running with `hostNetwork: true`, the `salt-master` could still connect to the mysql instance running (despite is a different pod, the public one where `salt-master` is defined runs with `hostNetwork: true`).

flavio:
@ereslibre wrote:
> My only doubt is how to proxy salt-master. It's clear that on the public manifest (the one using hostNetwork: true) we will proxy velum (trivial HTTP proxy). I am not that sure about how to proxy salt-master service, since it also needs to be reachable by other salt-minions.

What do you mean with proxy these services? We have to expose both of the to the world, I see no problem with the current implementation

ereslibre:
Yes, nothing. If we specify (`salt-api`, `salt-minion-ca`, `mariadb`, `event_processor`) on the `private` manifest (`hostNetwork: false`) and (`velum`, `salt-master`) on the `public` manifest (`hostNetwork: true`), it should work. `Velum` and `salt-master` should be reachable from the outside, and both of them should be able to talk to the rest of the components, no?

flavio:
`velum` has to talk with the `salt-api` which AFAIK is not listening on a unix socket. So they must be inside of the same pod to share the network namespace. The only difference is: `salt-api` should listen only to `localhost`, not to `0.0.0.0`.

kubernetes/start
robdaemon: Maybe? The cleanup script needs some work then. It will currently fail if there are no containers to clean up.

config/database.yml
robdaemon: Good point. I'll adjust that.

robdaemon:
@ereslibre - I wasn't quite clear enough on what I meant about using hostPath - it's more secure than an open TCP port, but not as secure as an emptyDir. emptyDir volumes are kept in RAM instead of the filesystem.

I'll take the approach of splitting into public and private pods, allowing us to keep the UNIX socket in an emptyDir.

Now, if we were running inside a full Kubernetes cluster, I would make other decisions, but this seems like the best path forward for a disconnected kubelet.

robdaemon:
I believe I've addressed all the feedback. I tested by firing up minions via the terraform repo, and they appear in velum, and mysql is not exposed via TCP.

I didn't squash yet - I'm not familiar with the standard process in the team, so I erred on you being able to see the differences between the initial commit and this one easily. Can squash when changes are approved.

kubernetes/setup-environment
```diff
@@ -11,8 +11,16 @@ done
...
 log "MariaDB container is ready"
 log "Setting up database"
-docker exec -i $(docker ps | grep velum-dashboard | awk '{print $1}') rake db:setup
-docker exec -e RAILS_ENV=test -i $(docker ps | grep velum-dashboard | awk '{print $1}') rake db:setup
+until docker exec -i $(docker ps | grep velum-dashboard | awk '{print $1}') rake db:setup;
+do
+    log "Waiting for mariadb to accept connections"
+    sleep 5
+done
+until docker exec -e RAILS_ENV=test -i $(docker ps | grep velum-dashboard | awk '{print $1}') rake db:setup;
```
MaximilianMeister: I think here we can just have this one statement, without the loop, as we just want to create the test database for the unit testing in the development environment. and by that time the mariadb connection should already be up accepting connections.

MaximilianMeister:
> I didn't squash yet - I'm not familiar with the standard process in the team, so I erred on you being able to see the differences between the initial commit and this one easily. Can squash when changes are approved.

that's already a good practice, we are squashing the commits in the end usually.
thanks for the changes!

but somehow the e2e tests were timing out, so there must have gone sth wrong. HINT: you can run them also locally against your PR-branch with https://github.com/kubic-project/e2e-tests
just make sure to set `VELUM_BRANCH=pull/104/head`

kubernetes/salt/config/master.d/salt-master.conf
```diff
@@ -5,11 +5,11 @@ interface: 0.0.0.0
 event_return: mysql
 
 mysql:
-  host: '127.0.0.1'
+  # salt does not support specifying the UNIX socket location here - as a workaround,
```
ereslibre: https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.mysql.html says you can set `mysql.unix_socket: '<path_to_socket_here>'`. Didn't this work?

ereslibre: Another option is to just make mysql listen on `127.0.0.1` (we would need to change less settings, and still it wouldn't be reachable from the outside).

kubernetes/setup-environment
ereslibre: Maybe with the unix socket this is a bit different. There could be an execution path in which `/var/run/mysqld/mysql.sock` was created, but `mysql` wasn't listening yet there, might that be? I would say we keep the same strategy as before, listening on `127.0.0.1` instead of a unix socket, unless there are strong reasons to switch to a unix socket.

robdaemon: That's why the `until` loops exist - this script can detect the presence of the file before it's actually listening on it. I see @MaximilianMeister's point where the second one shouldn't be necessary.

kubernetes/salt/config/master.d/salt-master.conf
robdaemon: I looked at that - that appears to be the doc for manipulating a mysql instance via a salt script, whereas the pillar does not support that: https://github.com/saltstack/salt/blob/develop/salt/pillar/mysql.py#L102-L112

ereslibre: Sorry, you are absolutely correct.

ereslibre: Shouldn't then mysql listen rather on `127.0.0.1` as we had before instead of a unix socket? We wouldn't need the `MYSQL_UNIX_PORT` workaround.

robdaemon: It is a bit of an oversight in the pillar. Maybe I should submit a PR to address that 😄

kubernetes/start
ereslibre: This cleanup will basically `docker ps -a | grep k8s | awk '{print $1}' | xargs docker rm -vf`. I don't think it's a big deal whether it succeeds or not.

PI-Victor: if you don't want it to fail or complain and just pass silently if there's nothing to remove, you can also do
```
for i in `docker ps -a | grep -iv name | awk '{print $1}'`; do docker rm -vf $i; done
```
just mentioning because it's my favorite :)

with docker 1.13 it's nicer
```
vpalade:codeflavor$ docker system prune
WARNING! This will remove:
	- all stopped containers
	- all volumes not used by at least one container
	- all networks not used by at least one container
	- all dangling images
Are you sure you want to continue? [y/N] y
Deleted Volumes:
6f9ed8fcce60ad75f525baa42a0c3444d9be81294357cc37f37b5b89f947c644
b2788412fb1df308f60e7e877abeb418ea03c0d47437b36facb78e374c766402
bd3920f73bb15e7f05636248bf54c12271fa582dc7dfd6464020c682223f2437
e632d0d82b0fda0f30a204ce55b43f719661a7dd2f5aa81558052a198be25e23
203af67358fab87e30b2df15a50ea06f191fea2203d810349a55b5bc38851c33
d1948ab203a8af263d033f6acb4ddc22a00077d98264f21e28c81efea58e8f60
df5ce42dd760183573ae928b2c0e3c054fa0131b0526b9447dc702cb344d711c
e9c731350126800da8554f58a93c5c58ae4309deb04ccb17ecbda526cc7f3f73
ccdbaf01eaf6b16abf095adbeceba0174d02ca6b78a9359eaf8f5dfad92a07ac

Total reclaimed space: 1.904 GB
```

flavio: If the cleanup script fails because there's nothing to remove then we can use the `for` instruction proposed by @PI-Victor.

ereslibre:
I am trying the patch, I am getting `[ERROR   ] Could not store events - returner 'mysql' raised exception: MySQL returner could not connect to database: (2005, "Unknown MySQL server host 'salt' (-2)")` on the `salt-master` container. Will check further and report back.

robdaemon:
I do see that MySQL error once, since the slot master starts before MySQL is fully initialized. Are you seeing that error occur multiple times?

I'm still working on running the e2e tests. I was having trouble running them yesterday from home due to download speeds on the dist server in Nuremberg.

ereslibre:
Yes, I had to clean `kubernetes/tmp`, something strange happened. I can see minions appearing on the UI but the orchestration does not finish properly. I will try to have a look at it as soon as possible and report back.

MaximilianMeister:
could you still squash the commits, or do a force push to retrigger the e2e tests? thanks!

robdaemon:
@ereslibre switching to 127.0.0.1 would work, but would likely require moving more things into the same pod, so that they can all reach mysql. There are components in both public.yaml and private.yaml that need access to mysql.

ereslibre:
> @ereslibre switching to 127.0.0.1 would work, but would likely require moving more things into the same pod, so that they can all reach mysql. There are components in both public.yaml and private.yaml that need access to mysql.

Since `mariadb` is on the `private` pod, all services on the `private` pod will be able to access `mariadb` on `127.0.0.1`, and since the `public` pod is using `hostNetwork: true`, in theory all containers in this pod should also be able to reach `mariadb`, no?

robdaemon:
No, that shouldn't work - the private pod will bind to the container's internal IP address (hanging off a bridge) - so, since the host itself wouldn't be able to talk to the MariaDB on 127.0.0.1, neither would the public pod. This is one of the tricky things about not having kube-proxy.

robdaemon:
Oh, I think the failing e2e tests make sense now - the mismatch in this PR and the `terraform` project's deployment descriptors?

ereslibre:
> No, that shouldn't work - the private pod will bind to the container's internal IP address (hanging off a bridge) - so, since the host itself wouldn't be able to talk to the MariaDB on 127.0.0.1, neither would the public pod. This is one of the tricky things about not having kube-proxy.

Of course you are right.

robdaemon:
Hey, all the tests pass now! Although I think the e2e tests don't run anymore, @MaximilianMeister? I did manage to run them locally though.

ereslibre:
I don't understand the latest change, in which `salt-api` and `salt-minion-ca` have been added to a `public` pod. I was able to create the cluster with them being on the private pod and we don't need to expose the `salt-api` to the world (even when we generate strong secrets this service will only be called from `velum`), neither the `salt-minion-ca` needs to be exposed (this container generates certificates for all services, but AFAIK every request/response goes through the `salt-master`).

ereslibre:
Now for example we have on the `private` pod defined the host paths: `salt-minion-ca-grains`, `salt-minion-ca-config` and `salt-minion-ca-signing-policies`, but they are used in the `public` pod. Will this work? I haven't checked if there are other cross references.

ereslibre:
> I don't understand the latest change, in which salt-api and salt-minion-ca have been added to a public pod. I was able to create the cluster with them being on the private pod and we don't need to expose the salt-api to the world (even when we generate strong secrets this service will only be called from velum), neither the salt-minion-ca needs to be exposed (this container generates certificates for all services, but AFAIK every request/response goes through the salt-master).

Sorry, I see that `salt-api` is now listening on `127.0.0.1` despite being on the `public` pod, which is even better than listening on `0.0.0.0` on the `private` pod.

Why the `salt-minion-ca` needs to be on the `public` pod instead of the `private` one?

ereslibre:
> Now for example we have on the private pod defined the host paths: salt-minion-ca-grains, salt-minion-ca-config and salt-minion-ca-signing-policies, but they are used in the public pod. Will this work? I haven't checked if there are other cross references.

Those cross references are just leftovers on the `private` pod, since they are also on the `public` pod.

ereslibre:
I created https://github.com/kubic-project/velum/pull/109 because conversation would get messy otherwise. Everything on this PR is fine from my point of view, in this PR that I created I just added the event processor to the public pod, since it might need to contact with the salt-api in the very very near future (this is, to retrieve CPU, memory & other information from grains when new machines get registered).

This PR also contains some cleanup on the private pod.

I have tried a deployment and it worked fine.

robdaemon:
@ereslibre the salt-minion-ca talks to the salt master on localhost: https://github.com/kubic-project/velum/blob/master/kubernetes/salt/config/minion.d-ca/minion.conf#L2

I'll take a look at PR 109 now.

MaximilianMeister:
> Hey, all the tests pass now! Although I think the e2e tests don't run anymore, @MaximilianMeister? I did manage to run them locally though.

@robdaemon for some reason they were stuck, thanks for the hint. should be working again now after restarting concourse

ereslibre:
Should we merge this?

MaximilianMeister:
> Should we merge this?

if there was no change in the templates in the meantime, i'd say lets merge it. it's all green. but we should check if there was a change in the templates, as it can't be seen in this diff due to the name change of the files.

ereslibre:
It has my green light (I already approved it).

PR #106, opened 2017-03-31
Request the ca certificate directly to the CA machine (container)

ereslibre:
The ca certificate was retrieved from the mine. The mine gets wiped
when the admin node gets rebooted (the salt-master container gets
rebooted). We don't want to wait for the mine to be populated if we
actually have this information in the filesystem.

If an orchestration is run, this information will be automatically
populated on the mine (but we don't want to force that), so we
directly use salt to cat the ca certificate from the ca machine (it's
a container at the moment).

### DIFF
--- a/lib/velum/kubernetes.rb
+++ b/lib/velum/kubernetes.rb
@@ -11,9 +11,9 @@ module Velum
 
     # Returns the Kubernetes apiserver configuration. It returns a KubeConfig struct.
     def self.kubeconfig
-      _, ca = Velum::Salt.call action:  "mine.get",
-                               targets: "ca",
-                               arg:     ["ca", "x509.get_pem_entries"]
+      _, ca_crt = Velum::Salt.call action:  "cmd.run",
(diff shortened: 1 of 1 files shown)


lib/velum/kubernetes.rb
```diff
@@ -24,7 +24,7 @@ def self.kubeconfig
                                           arg:         "cat /etc/pki/minion.key"
       host = Minion.master.applied.pluck(:hostname).first
       # rubocop:disable Style/RescueModifier
-      ca_crt = ca["return"].first["ca"]["ca"]["/etc/pki/ca.crt"] rescue nil
+      ca_crt = ca_crt["return"].first.values.first rescue nil
```
jimmykarily: I would prefer if we used a different variable for this one. If some other code flow is introduced between the initial assignment (line 14 above) and this one, we might end up running only the first assignment and we end up with an attr_accessor which returns an unexpected value. This is and edge case of course but better safe than sorry :).

PR #108, opened 2017-03-31
Store and retrieve secret_key_base from disk file

HouzuoGuo:
As Rafael recommended, this time it uses an approach similar to gitlab's.

### DIFF
--- a/config/initializers/secrets.rb
+++ b/config/initializers/secrets.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+# Be sure to restart your server after modifying this file.
+
+require "pathname"
+require "velum/secrets"
+
+env_path = Pathname.new("#{ENV.fetch("VELUM_SECRETS_DIR")}/key_base.json")
+Rails.application.secrets.secret_key_base = Velum.read_create_secret_key_base(env_path)
--- a/kubernetes/manifest-templates/velum.yaml
+++ b/kubernetes/manifest-templates/velum.yaml
@@ -12,6 +12,8 @@ spec:
     env:
     - name: RAILS_ENV
       value: development
+    - name: VELUM_SECRETS_DIR
+      value: /tmp/velum-secrets
     - name: RAILS_SERVE_STATIC_FILES
       value: "true"
     - name: VELUM_DB_HOST
@@ -39,6 +41,8 @@ spec:
     volumeMounts:
       - mountPath: /usr/src/app
         name: velum-source-code
+      - mountPath: /tmp/velum-secrets
+        name: velum-secrets
     command: ["/bin/sh","-c"]
     args: ["bundle exec rails s -b 0.0.0.0 -p 3000 --pid /tmp/puma.pid Puma"]
   - name: velum-event-processor
@@ -46,6 +50,8 @@ spec:
     env:
     - name: WORKER_ID
       value: Worker_1
+    - name: VELUM_SECRETS_DIR
+      value: /tmp/velum-secrets
     - name: RAILS_ENV
       value: development
     - name: VELUM_DB_HOST
@@ -71,6 +77,8 @@ spec:
     volumeMounts:
       - mountPath: /usr/src/app
         name: velum-source-code
+      - mountPath: /tmp/velum-secrets
+        name: velum-secrets
     command: ["/bin/sh","-c"]
     args: ["bundle exec bin/rake salt:process"]
   - name: velum-mariadb
@@ -82,3 +90,6 @@ spec:
     - name: velum-source-code
       hostPath:
         path: ${project_dir}
+    - name: velum-secrets
+      hostPath:
+        path: /tmp/velum-secrets
--- a/lib/velum/secrets.rb
+++ b/lib/velum/secrets.rb
@@ -1,0 +1,19 @@
+require "securerandom"
+
+module Velum
+  # If there is a rails key base stored on disk, return the stored one.
+  # Otherwise, create a new key base from cryptographic random pool and store it.
+  def self.read_create_secret_key_base(pathname)
+    pathname.dirname.mkpath
+    key_base = nil
+    if pathname.exist?
+      content = pathname.read
+      content != "" && key_base = content
+    end
+    if key_base.nil?
+      key_base = SecureRandom.hex(64)
+      pathname.write(key_base)
+    end
+    key_base
+  end
+end
--- a/spec/lib/velum/secrets_spec.rb
+++ b/spec/lib/velum/secrets_spec.rb
@@ -1,0 +1,19 @@
+require "spec_helper"
+require "velum/secrets"
+
+describe "create_secret_key_base" do
+  key_base_dir = Pathname.new("/tmp/velum_secrets_spec")
+  key_base_path = key_base_dir.join("key_base")
+
+  after do
+    FileUtils::rm_r key_base_dir
+  end
+
+  it "makes new key_base" do
+    Velum.read_create_secret_key_base(key_base_path)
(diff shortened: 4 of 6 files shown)


mssola:
If the goal is to pass a secret key base, why don't we pass it through an environment variable as it's customary in container-based setups ?

ereslibre:
> If the goal is to pass a secret key base, why don't we pass it through an environment variable as it's customary in container-based setups ?

Because our manifests come from `caasp-container-manifests` package and we shouldn't modify them for each installation (it would be harder to manage updates and such). This way we ensure that each installation contains different application secrets without the need to have special requirements when updating `caasp-container-manifests` package.

config/initializers/secrets.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# Be sure to restart your server after modifying this file.
+
+require "pathname"
+require "velum/secrets"
+
+env_path = Pathname.new("#{ENV.fetch("VELUM_SECRETS_DIR")}/key_base.json")
```
ereslibre: Cannot this be expressed as: `env_path = File.join ENV["VELUM_SECRETS_DIR"], "key_base.json"` ?

kubernetes/manifest-templates/velum.yaml
```diff
@@ -82,3 +90,6 @@ spec:
     - name: velum-source-code
       hostPath:
         path: ${project_dir}
+    - name: velum-secrets
+      hostPath:
+        path: /tmp/velum-secrets
```
ereslibre: It's OK to mount secrets on `/tmp` on the containers, but on the host we want this path to survive reboots, so our secrets are saved across restarts (otherwise we can lose sessions right now, or other secrets in the future)

lib/velum/secrets.rb
```diff
@@ -0,0 +1,20 @@
...
+    key_base = nil
+    if pathname.exist?
+      content = IO.read(pathname)
+      if content != ""
+        key_base = content
+      end
+    end
+    if key_base == nil
+      key_base = SecureRandom.hex(64)
+      IO.write(pathname, key_base)
```
ereslibre: I would use `File` API: `File.open(pathname, "w") { |f| f.write SecureRandom.hex(64) }`

lib/velum/secrets.rb
```diff
@@ -0,0 +1,20 @@
+require 'securerandom'
+
+module Velum
+  # If there is a rails key base stored on disk, return the stored one.
+  # Otherwise, create a new key base from cryptographic random pool and store it.
+  def self.create_secret_key_base(pathname)
+    key_base = nil
+    if pathname.exist?
+      content = IO.read(pathname)
```
ereslibre: I would use `File` API: `File.read pathname`

config/initializers/secrets.rb
```diff
@@ -0,0 +1,8 @@
...
+Rails.application.secrets.secret_key_base = Velum.create_secret_key_base(env_path)
```
ereslibre: Maybe I'd rename `create_secret_key_base` to `read_or_create_secret_key_base`. Just to give it a name closer to what it does.

lib/velum/secrets.rb
```diff
@@ -0,0 +1,20 @@
...
+        key_base = content
+      end
+    end
+    if key_base == nil
+      key_base = SecureRandom.hex(64)
+      IO.write(pathname, key_base)
+    end
+    key_base
+  end
+end
```
ereslibre: Overall this is not very "the ruby way". Rewrite hint:

```
def self.create_secret_key_base(pathname)
  if File.exist? pathname
    File.read pathname
  else
    SecureRandom.hex(64).tap do |secret|
      File.open(pathname, "w") { |f| f.write secret }
    end
  end
end
```

Disclaimer: not tested.

mssola:
> Because our manifests come from caasp-container-manifests package and we shouldn't modify them for each installation (it would be harder to manage updates and such). This way we ensure that each installation contains different application secrets without the need to have special requirements when updating caasp-container-manifests package.

Got it, thanks.

lib/velum/secrets.rb
```diff
@@ -0,0 +1,20 @@
+require 'securerandom'
+
+module Velum
```
mssola: I'd create either a `Secrets` module or class inside of the `Velum` module. That's what a ruby developer would expect I think.

config/initializers/secrets.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# Be sure to restart your server after modifying this file.
+
+require "pathname"
```
mssola: If you then follow @ereslibre's suggestion, then remove this since it won't be needed anymore 😉

config/initializers/secrets.rb
MaximilianMeister: i think a `pathname` object makes more sense here, as it is a handy preparation for all kinds of file operations

lib/velum/secrets.rb
MaximilianMeister: I would use `pathname.read`

lib/velum/secrets.rb
MaximilianMeister: `pathname.open("w") { |f| f.write SecureRandom.hex(64) }`

MaximilianMeister: or even just `pathname.write(SecureRandom.hex(64))` i think that could work too

MaximilianMeister:
e2e-tests are also green after a retrigger :+1: (they were a bit broken today for other reasons)

lib/velum/secrets.rb
ereslibre: I agree with your comment if we pass the `Pathname` around. I have never used `Pathname` since I think is a bit misleading in front of `File.read` (that for me is clear -> read this file), and `Pathname` semantics are a bit more blurry -- but I agree with your comments regarding `pathname` if we pass it around and use its instance methods.

MaximilianMeister: I know for sure that you've already used it ;-) e.g. `Rails.root` is a `Pathname` object

ereslibre: Hehe, yes, for handling paths, but not for reading or writing file contents :)

lib/velum/secrets.rb
HouzuoGuo: OOP is how I approached it in the very first revision of the PR from three weeks ago, but many decided against OOP approach, well, that happened.

spec/lib/velum/secrets_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "spec_helper"
+require "velum/secrets"
+
+describe "create_secret_key_base" do
+  key_base_dir = File.join("/tmp", "velum_secrets_spec")
```
MaximilianMeister: If you're already using `Pathname` below you could also use it here, and replace the `IO.read` ocurrences below

lib/velum/secrets.rb
```diff
@@ -0,0 +1,19 @@
+require "securerandom"
+
+module Velum
+  # If there is a rails key base stored on disk, return the stored one.
+  # Otherwise, create a new key base from cryptographic random pool and store it.
+  def self.read_create_secret_key_base(pathname)
+    FileUtils.mkdir_p pathname.dirname
```
MaximilianMeister: `pathname.dirname.mkpath` should work too

HouzuoGuo:
I have ultimately failed to please rubocop, after having taken hours to write merely dozens of lines of code I may only conclude that ruby is not the right language for me. Well, see you later.

PR #109, opened 2017-04-04
Update manifests

ereslibre:
Move the event processor to the `public` pod since it will very soon
require to connect to the `salt-api` service in order to fetch grains
from recently registered machines in order to retrieve number of CPUs,
memory, hard disk space...
    
Also cleanup the private manifest that did contain some host paths that
were used by containers that used to be on this private pod.

### DIFF
--- a/kubernetes/manifest-templates/private.yaml
+++ b/kubernetes/manifest-templates/private.yaml
@@ -7,40 +7,6 @@ metadata:
 spec:
   hostNetwork: false
   containers:
-  - name: velum-event-processor
-    image: opensuse/velum:development
-    env:
-    - name: WORKER_ID
-      value: Worker_1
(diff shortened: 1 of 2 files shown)


ereslibre:
If this gets merged at all, it is meant to be merged to `rar_mysql_unix_socket` branch, and then get merged with https://github.com/kubic-project/velum/pull/104 altogether. But feel free to ignore this PR and implement it directly on https://github.com/kubic-project/velum/pull/104, I just wanted to point out that we will be needing the event processor to access the `salt-api` in the short term, and the volume cleanup.

robdaemon:
Okay, going to merge this in, since it merges into my branch, and then when people are ready, they can merge my PR :)

PR #110, opened 2017-04-04
[WIP] Store and retrieve secret_key_base from disk file

ereslibre:
Take it from where @HouzuoGuo left it.

### DIFF
--- a/lib/velum/secrets.rb
+++ b/lib/velum/secrets.rb
@@ -3,17 +3,19 @@ require "securerandom"
 module Velum
   # If there is a rails key base stored on disk, return the stored one.
   # Otherwise, create a new key base from cryptographic random pool and store it.
-  def self.read_create_secret_key_base(pathname)
-    pathname.dirname.mkpath
-    key_base = nil
-    if pathname.exist?
-      content = pathname.read
-      content != "" && key_base = content
+  class Secrets
+    def self.read_or_create_secret_key_base(pathname)
+      pathname.dirname.mkpath
+      key_base = nil
+      if pathname.exist?
+        content = pathname.read
+        content != "" && key_base = content
+      end
+      if key_base.nil?
+        key_base = SecureRandom.hex(64)
+        pathname.write(key_base)
+      end
+      key_base
     end
-    if key_base.nil?
-      key_base = SecureRandom.hex(64)
-      pathname.write(key_base)
-    end
-    key_base
   end
 end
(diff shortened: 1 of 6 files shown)


kubernetes/manifest-templates/velum.yaml
```diff
@@ -82,3 +90,6 @@ spec:
     - name: velum-source-code
       hostPath:
         path: ${project_dir}
+    - name: velum-secrets
+      hostPath:
+        path: /tmp/velum-secrets
```
jimmykarily: I think `/tmp` dir is cleanup using various mechanisms. If the host decides to cleanup its /tmp dir or the guest decides to do the same, then bye bye secrets. If that happens, all active sessions will become invalid, no? Do we want to persist the secrets across reboots and whatever? If yes then maybe `/tmp` is not the best choice (both on host and guest).

jimmykarily: I now see, we only set this to `/tmp` on dev and test. It shouldn't be a problem then.

lib/velum/secrets.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Velum
+  # If there is a rails key base stored on disk, return the stored one.
+  # Otherwise, create a new key base from cryptographic random pool and store it.
+  class Secrets
+    def self.read_or_create_secret_key_base(pathname)
+      pathname.dirname.mkpath
+      key_base = nil
+      if pathname.exist?
+        content = pathname.read
```
jimmykarily: Maybe add a `.strip` just in case there is some whitespace in the file? Just to make sure the next line works.

jimmykarily:
Is this done? The e2e tests are breaking but doesn't seem related to the changes in this PR. Lets get them green before merge though.

kubernetes/manifest-templates/velum.yaml
MaximilianMeister: nevertheless, as it is rails app specific, can't we always use the rails tmp dir here? then we wouldn't need an environment variable nor a volume mount.

MaximilianMeister: this also needs a rebase now after merging #104

ereslibre:
I will fix this ASAP, so we can remove the WIP tag.

kubernetes/manifest-templates/velum.yaml
MaximilianMeister: as discussed on IRC, the rails app is only mounted persistently in the dev environment, not in production. so we still need to find a folder.

ereslibre:
@MaximilianMeister is taking over this task on https://github.com/kubic-project/velum/tree/secret_key_base, so closing this one.

PR #112, opened 2017-04-10
fix bsc#1031682

MaximilianMeister:
run a mini server during the time that velum is waiting for the db
this avoids the awkward situation that nothing seems to happen directly
after the installation

https://bugzilla.suse.com/show_bug.cgi?id=1031682

database_not_ready

### DIFF
--- a/bin/init
+++ b/bin/init
@@ -11,6 +11,11 @@ setup_database() {
   COUNT=0
   RETRY=1
 
+  # let the user know that the UI is about to start
+  # https://bugzilla.suse.com/show_bug.cgi?id=1031682
+  ruby -run -e httpd public/503_database_not_ready.html -p 80 --request-timeout=$TIMEOUT --bind-address=0.0.0.0 &
+  export TEMPORARY_SERVER_PID=$!
+
   while [ $RETRY -ne 0 ]; do
     case $(bundle exec rails r bin/check_db.rb | grep DB) in
       "DB_DOWN")
@@ -35,6 +40,9 @@ setup_database() {
         ;;
       "DB_READY")
         echo "Database ready"
+        # UI is ready, so we can kill the temporary server
+        # https://bugzilla.suse.com/show_bug.cgi?id=1031682
+        kill -9 $TEMPORARY_SERVER_PID
         break
         ;;
     esac
--- a/public/503_database_not_ready.html
+++ b/public/503_database_not_ready.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>The database is not running (503)</title>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <style>
+  body {
+    background-color: #EFEFEF;
+    color: #2E2F30;
+    text-align: center;
+    font-family: arial, sans-serif;
+    margin: 0;
+  }
+
+  div.dialog {
+    width: 95%;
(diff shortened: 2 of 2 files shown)


MaximilianMeister:
### NOTE

If you want to test this in the development environment, you need to adapt the velum manifest to call the init script like [here](https://github.com/kubic-project/caasp-container-manifests/blob/master/velum.yaml#L47)

MaximilianMeister:
btw my first thought was to add a `rescue_from` mysql error to the `application_controller` but i couldn't even get to that point, as the error was raised before the `application_controller`, that's why I hooked into the init script, instead of trying to go down the rails stack and add some custom code to redirect

jordimassaguerpla:
IIRC, the container was being restarted because the database was missing ... will this fix fix that?

MaximilianMeister:
> IIRC, the container was being restarted because the database was missing ... will this fix fix that?

there wasn't anything mentioned in the bugreport, so i haven't checked that. but if the container would be restarting, the init script would stop to execute, which is not the case here.

MaximilianMeister:
> So we also need to write some instructions of the kind Please, visit http://<hostname_or_ip_here> in order to boostrap the cluster in the console itself.

thanks for the hint. not sure, as i am always using the devenv. is this something else for velum? or is this something that needs to be adapted in the installer?

ereslibre:
@MaximilianMeister yeah, nothing to be done on this side. I just wanted to point out that for bsc#1031682 to be closed we still need to fix this at installation level (maybe we can do something on the activation script if YaST doesn't do it).

bin/init
```diff
@@ -11,6 +11,11 @@ setup_database() {
   COUNT=0
   RETRY=1
 
+  # let the user know that the UI is about to start
+  # https://bugzilla.suse.com/show_bug.cgi?id=1031682
+  ruby -run -e httpd public/503_database_not_ready.html -p 80 --request-timeout=$TIMEOUT &
```
MaximilianMeister: I just realised that we need to bind to all interfaces here...

MaximilianMeister:
> The only thing that does not convince me about this is that the path on the browser is /503_database_not_ready.html. If we could just show this same warning but on / I would like it even more.
> 
> Is this possible?

the screenshot is a bad example, as i navigated there by hand... it's actually `/`

ereslibre:
I'm ok as it is right now, but maybe we can add a `<meta http-equiv="refresh" content="5">`. This would even avoid manual refreshes (and there are no forms where you can lose information or such).

MaximilianMeister:
> I'm ok as it is right now, but maybe we can add a <meta http-equiv="refresh" content="5">. This would even avoid manual refreshes (and there are no forms where you can lose information or such).

that sounds like a good idea, haven't thought about that. but i'm not sure if it's a huge benefit, imagine we want to redirect from a controller during runtime because of some database issue, then we would be at that url, and the refresh would not be really useful. because of possible other cases i wrote "If you just finished the installation... then, should the error persist ..." not sure if that makes sense now, as we're just using it for that particular case...

but we might add some `rescue_from MySQL::Error` at some point in the application controller and redirect to that page...

all in all IMO refreshing by hand at that point isn't a big deal, as it is only a window of 1min once in a lifetime (hopefully) but if you insist I'll change it!

ereslibre:
It's ok @MaximilianMeister, we can improve it later if needed.

PR #114, opened 2017-04-11
Fix kubelet arguments for version >= 1.6.1

ereslibre:


PI-Victor:
ah, so that's what i was doing wrong. i was passing --kubeconfig instead. it didn't make sense to me to pass manifests to kubeconfig but i didn't give it a second look. testing now.

mjura:
@ereslibre : Should we fix also kubernetes-kubelet package and /etc/kubernetes/kubelet file?

ereslibre:
@mjura I don't think we need to change that, because by default it still is `/etc/kubernetes/manifests`, and I think the only place where we were overriding this manifest location is here in velum.

PI-Victor:
tested and works. LGTM.

PR #115, opened 2017-04-11
Generate unique minion ID for Minions, (bsc#1031623)

mjura:
Generate unique minion ID for Minions if DHCP does not provide hostname

### DIFF
--- a/app/views/dashboard/autoyast.xml.erb
+++ b/app/views/dashboard/autoyast.xml.erb
@@ -9,7 +9,9 @@
         <interpreter>shell</interpreter>
         <source><![CDATA[
         #!/bin/sh
-        echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
+        MINION_CFG="/etc/salt/minion.d/master.conf"
+        echo "id: <%= SecureRandom.hex %>" > $MINION_CFG
+        echo "master: <%= @controller_node %>" >> $MINION_CFG
         ]]>
         </source>
       </script>


app/views/dashboard/autoyast.xml.erb
```diff
@@ -9,7 +9,12 @@
         <interpreter>shell</interpreter>
         <source><![CDATA[
         #!/bin/sh
-        echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
+        MINION_ID=$(grep id /tmp/master-id | sed 's/id: //')
+        if [ -z "$MINION_ID" ]; then
+            UUID=$(uuidgen -r)
+            echo "id: $UUID" > /etc/salt/minion.d/master.conf      
+            echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
```
ereslibre: You need to append, not overwrite. Use `>>` instead of `>` or the previous `id` redirection will be void.

app/views/dashboard/autoyast.xml.erb
```diff
@@ -9,7 +9,12 @@
...
+        fi
```
ereslibre: This could be replaced completely by the following snippet, which is simpler, if we are talking about completely random `minion-ids` because it's generated on the rails application instead of the machine being installed:

```
        #!/bin/sh
        echo "id: <%= SecureRandom.hex %>" > /etc/salt/minion.d/master.conf
        echo "master: <%= @controller_node %>" >> /etc/salt/minion.d/master.conf
        ]]>
```

However I think we should try to get this ID from `/etc/machine-id` if that's possible at all. At least it would be good to explore this possibility.

This merge has some other implications in our salt states that we need to address. Maybe we should create a dedicated branch for this change (for example, on [salt states](https://github.com/kubic-project/salt/search?utf8=%E2%9C%93&q=grains&type=) we take as reference `grains['ip4_interfaces']['eth0'][0]` and I am not sure with this change if `grains['ip4_interfaces']['eth0'][0]` will be the hostname or the minion_id - probably the minion_id, so we nee to adapt some parts of our salt states)

app/views/dashboard/autoyast.xml.erb
mjura: I fixed this

app/views/dashboard/autoyast.xml.erb
```diff
@@ -9,7 +9,12 @@
         <interpreter>shell</interpreter>
         <source><![CDATA[
         #!/bin/sh
-        echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
+        MINION_CFG="/etc/salt/minion.d/master.conf"
+        [ -f $MINION_CFG ] && MINION_ID=$(grep id $MINION_CFG | sed 's/id: //')
```
ereslibre: If the autoyast profile is requested, do we really need to check this? This will only be requested and interpreted in an unattended installation right? So we may just assume this is a clean installation and do the `echoes` without having to check anything else, right?

mjura: Ok, as you want. I am fan of validating everything but you are right that this will be run once only during installation.

app/views/dashboard/autoyast.xml.erb
mjura: This still has to be tested.

thkukuk:
I don't like the fix for one simple reason: it only works with autoyast (and even here I'm not sure), but not with an installation from DVD or VMX images (which means also not with "golden" images from the customer). 
How does the SUSE Manager team did solve this?
I think a better way is to create the minion id based on /etc/machine_id during the first start of salt-minion.

ereslibre:
This has been already fixed at packaging level. We need to find out how we are going to fix this for the opensuse case using terraform (it shouldn't be very tricky).

PR #116, opened 2017-04-12
Improve bootstrap documentation

ereslibre:
Fixes bsc#1031685

### DIFF
--- a/app/views/setup/worker_bootstrap.html.slim
+++ b/app/views/setup/worker_bootstrap.html.slim
@@ -14,19 +14,19 @@ p
   |  This process leverages AutoYaST and is (almost) fully automated.
   |  In case you are not familiar with it, you can find more information about AutoYaST in the&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html" official documentation.
-  |  The automatic installation gets invoked by adding autoyast=#{autoyast_url} to the kernel parameter list.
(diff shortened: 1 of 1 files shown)


app/views/setup/worker_bootstrap.html.slim
```diff
@@ -14,19 +14,19 @@ p
...
   |  Bootstrap all the nodes you want to make part of this platform by adding the following boot parameter
-  | autoyast=#{autoyast_url}
+  |  <strong>autoyast=#{autoyast_url}</strong>
 
 h2 Tips
 p
   | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#bootmedium.pxe" here
-  | . Where #{autoyast_url} is the real, generated path to the AutoYaST profile served by the dashboard.
+  | . Where <strong>#{autoyast_url}</strong> is the real, generated path to the AutoYaST profile served by the dashboard.
```
MaximilianMeister: would it make sense to make this a hyperlink as well?

ereslibre: I wouldn't add an hyperlink here, since we don't really want the customer/user to visit it, no?

MaximilianMeister: not necessarily, although he could do that to see it upfront, but i guess that's not really necessary as you said

PR #119, opened 2017-04-18
enhancements to the initialization and cleanup

MaximilianMeister:


kubernetes/cleanup
```diff
@@ -4,3 +4,6 @@ log()   { echo ">>> $1" ; }
 
 log "Cleaning up old containers"
 docker ps -a | grep k8s | awk '{print $1}' | xargs docker rm -vf
+
+log "Removing any existing server process id's"
+rm -rf ../tmp/pids/server.pid
```
ereslibre: For being honest I preferred how we did this before: https://github.com/kubic-project/velum/commit/10954daeea7b5f613b4c828ab1c881ea78ec8aa0#diff-f1f577cbb090fc9738404e4d85e9c1ddL63

We were modifying the command to be executed so it wrote the PID inside the container (no mounts), so the PID always stayed inside the container itself with `--pid /tmp/puma.pid`. I think we should do the same [but here](https://github.com/kubic-project/velum/blob/master/bin/init#L56), and this cleanup wouldn't be necessary at all.

bin/init
```diff
@@ -50,7 +52,6 @@ setup_database() {
   set -e
 }
 
-: ${VELUM_PORT:=80}
 setup_database
 bundle exec rake db:migrate
-bundle exec "rails s -b 0.0.0.0 -p $VELUM_PORT Puma"
+bundle exec "rails s -b 0.0.0.0 -p $VELUM_PORT Puma" --pid tmp/pids/server.pid
```
ereslibre: The `--pid` argument goes with the `rails server` command: `bundle exec "rails s -b 0.0.0.0 -p $VELUM_PORT --pid tmp/pids/server.pid Puma"`

MaximilianMeister: oops, thanks

PR #121, opened 2017-04-18
Do not assume minion_id is hostname/fqdn

ereslibre:
With this change we:

* Rename hostname to fqdn. Feels more correct.
* Do not assume on the salt part that the minion_id is the hostname/fqdn
* Explicitly save minion_id as a column, as well as fqdn as another column
* We will use minion_id internally to talk to salt, but show hostname, and
  probably other properties [that we can cache in the very near future as
  columns] like number of cores, memory, disk...

## Depends on
https://github.com/kubic-project/salt/pull/40

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -84,7 +84,7 @@ MinionPoller = {
       <tr> \
         <th>" + minion.id +  "</th>\
         <td class='text-center'>" + statusHtml +  "</td>\
-        <td>" + minion.hostname +  "</td>\
+        <td>" + minion.fqdn +  "</td>\
         <td>" + (minion.role || '') +  "</td>\
         <td class='text-center'>" + masterHtml + "</td>\
       </tr>";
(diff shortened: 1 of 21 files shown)


MaximilianMeister:
unfortunately the e2e-tests with this patch, plus a fix (https://github.com/kubic-project/e2e-tests/pull/31) of the hostname check that failed here in this run still fail at the cluster info. the nodes are still not visible in the kubernetes api.
```
{
    "metadata": {
        "selfLink": "/api/v1/nodes",
        "resourceVersion": "76"
    },
    "items": []
}
```

do you think there is still something missing here for that PR or is it a separate issue? i am not 100% sure if it's related

ereslibre:
I'd try to reuse our existing migration now that we can to keep things more clean the maximum time we can.

Salt also needs to be updated (so this PR doesn't go alone) for the e2e tests to work: https://github.com/kubic-project/salt/pull/40.

However that PR (https://github.com/kubic-project/salt/pull/40) is a WIP and I will finish it this morning. The problem is that we need to blindly merge one or the other for the e2e tests to work for the other one.

MaximilianMeister:
> The problem is that we need to blindly merge one or the other for the e2e tests to work for the other one.

@ereslibre ok, just saw the salt PR as well. at any time i can run the e2e-tests with both PR's, but it won't be visible in the status this PR, but i guess that's ok. so just tell me when you think you're done with the salt part, and i'll run the tests again with both PR's

PR #122, opened 2017-04-18
Added NTP configuration for nodes in autoyast.xml

inercia:
Implement the changes suggested by @thkukuk in the `autoyast.xml` template for configuring NTP in the minions.

(see [this Trello card](https://trello.com/c/Orzhm7rZ))

### DIFF
--- a/app/views/dashboard/autoyast.xml.erb
+++ b/app/views/dashboard/autoyast.xml.erb
@@ -13,6 +13,17 @@
         ]]>
         </source>
       </script>
+      <script>
+        <chrooted config:type="boolean">true</chrooted>
+        <filename>configure_timesyncd.sh</filename>
+        <interpreter>shell</interpreter>
+        <source><![CDATA[
+        #!/bin/sh
+        sed -i -e 's|#NTP=.*|NTP=<%= @controller_node %>|g' /etc/systemd/timesyncd.conf
+        timedatectl set-ntp true
+        ]]>
+        </source>
+      </script>
     </chroot-scripts>
   </scripts>
   <bootloader>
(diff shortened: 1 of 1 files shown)


inercia:
BTW, this will add even more divergences between CaaSP and OpenSUSE.

inercia:
> That was fast. We will ask soon @controller_node to the customer, since I believe this is one of our weaknesses. Depending on how you visit velum, @controller_node will contain one thing or the other. We should explicitly ask for a reachable name/ip for the controller node on the setup phase and use that instead of guess that the rest of the machines will reach to @controller_node the same way the operator did to bootstrap the cluster.

Yeah, we should set a reasonable default value for `@controller_node`  (as you said, it makes sense to use the name or IP used for reaching the Dashboard), but that could be easily unreachable from the rest of the cluster in many cases, so the user should have the option to specify a different IP/address...

I've just filed a new issue (#123) with these concerns...

app/views/dashboard/autoyast.xml.erb
```diff
@@ -13,6 +13,17 @@
         ]]>
         </source>
       </script>
+      <script>
+        <chrooted config:type="boolean">true</chrooted>
+        <filename>configure_timesyncd.sh</filename>
+        <interpreter>shell</interpreter>
+        <source><![CDATA[
+        #!/bin/sh
+        sed -i -e 's|#NTP=.*|NTP=<%= @controller_node %>|g' /etc/systemd/timesyncd.conf
```
MaximilianMeister: MINOR: you could omit the `-e` here and it would work as well

inercia: I will not change not even a single char of what Throsten suggested [here](https://trello.com/c/Orzhm7rZ) :-D

PR #124, opened 2017-04-18
Cleaning folders that can be problematic

robdaemon:
I was having a lot of trouble starting a cluster because I ended up with
two SaltStack files that were supposed to be UNIX sockets, but were
regular files instead.

Making the cleanup script clear out the manifests/ and tmp/ folders

Also making a change to the cleanup script suggested in #104

### DIFF
--- a/kubernetes/cleanup
+++ b/kubernetes/cleanup
@@ -2,5 +2,11 @@
 
 log()   { echo ">>> $1" ; }
 
+[[ -d manifest-templates ]] || { echo >&2 "Please run this script from within the kubernetes folder"; exit 1; }
+
 log "Cleaning up old containers"
-docker ps -a | grep k8s | awk '{print $1}' | xargs docker rm -vf
+for i in `docker ps -a | grep k8s | awk '{print $1}'`; do docker rm -vf $i; done
+
+log "Cleaning the tmp and manifests folders"
+sudo rm -rf tmp/* manifests/* 
+git checkout tmp/ manifests/*
--- a/kubernetes/start
+++ b/kubernetes/start
@@ -30,7 +30,7 @@ while (( $interactive_mode == 1)) ; do
   esac
 done
 
-./cleanup || true
+./cleanup
 
 default_interface=$(awk '$2 == 00000000 { print $1 }' /proc/net/route)
 ip_address=$(ip addr show $default_interface | awk '$1 == "inet" {print $2}' | cut -f1 -d/)


kubernetes/cleanup
```diff
@@ -2,5 +2,11 @@
 
 log()   { echo ">>> $1" ; }
 
+[[ -d manifest-templates ]] || { echo >&2 "Please run this script from within the kubernetes folder"; exit 1; }
+
 log "Cleaning up old containers"
-docker ps -a | grep k8s | awk '{print $1}' | xargs docker rm -vf
+for i in `docker ps -a | grep -iv name | grep k8s | awk '{print $1}'`; do docker rm -vf $i; done
```
ereslibre: Why grep-skip with `name`?

kubernetes/cleanup
```diff
@@ -2,5 +2,11 @@
...
+
+log "Cleaning the tmp and manifests folders"
+sudo rm -rf tmp/* manifests/* 
+git checkout tmp/ manifests/*
```
ereslibre: I think there's no need to checkout `manifests` since the globbing won't match `.gitkeep`

kubernetes/start
```diff
@@ -30,7 +30,7 @@ while (( $interactive_mode == 1)) ; do
   esac
 done
 
-./cleanup || true
+./cleanup
```
ereslibre: I have seen this `|| true` appearing and disappearing for a while now :)

kubernetes/cleanup
PI-Victor: it eliminated the header of the `docker ps -a output`

```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                    PORTS               NAMES
```

kubernetes/start
PI-Victor: this shouldn't be removed, it will cause the whole script to fail if docker didn't find any images/containers to remove.

kubernetes/cleanup
ereslibre: But we are already grepping with `k8s`...

kubernetes/start
PI-Victor: mainly this part https://github.com/kubic-project/velum/blob/master/kubernetes/cleanup#L6

PI-Victor: bah! i didn't sleep enough, you've added it to the commits, my bad. nvm! lgtm.

kubernetes/cleanup
PI-Victor: ah, yeah. you're right. 👍

kubernetes/cleanup
```diff
@@ -2,5 +2,11 @@
 
 log()   { echo ">>> $1" ; }
 
+[[ -d manifest-templates ]] || { echo >&2 "Please run this script from within the kubernetes folder"; exit 1; }
```
MaximilianMeister: we could also just cd into the folder, i guess it would be a better default behaviour

PI-Victor: agreed. pushd/popd would work better

ereslibre: This script is meant to be run by `start` script or manually from the same folder, I don't see why we would need to try to make it smarter than we need for our development purposes (we already document that `start` needs to be executed from within the `velum/kubernetes` folder)... I think we can just keep the script "dumb" and simple...

PI-Victor: @ereslibre KISS purist!

robdaemon: I've ran this script standalone several times. I was following the same patterns as the start script here. 😀

PR #127, opened 2017-04-21
Explicitly ask for the dashboard location

ereslibre:
It will come with a reasonable default, but the person bootstrapping
the cluster will have the chance to set a different value so the
rest of the machines that will form part of the cluster will use this
reference during the bootstrapping process.

This modifies the workflow of the bootstrapping. To avoid problems, now
it's required for the customer/user to fill in all the first step
information in velum (registering first) for the autoyast profile to be
served (since it will contain a reference to the salt-master that will be
set on the rest of the machines). This way, the reference to the salt-master
on the autoyast profile, as well as the reference to the administration node
during the bootstrapping phase will be the one the customer/user set,
intead of being inferred from how the customer/user visited velum.

Until this first step of the setup is completed (dashboard reachable
location by the rest of the cluster for autoyast and salt) it won't be
possible to bring up new nodes (the autoyast profile will not be served, so
they will just raise an error in the YaST UI saying they can't retrieve
the autoyast configuration, since a 503 error will be returned).

Fixes bsc#1032461
Fixes bsc#1031605

### DIFF
--- a/app/views/setup/welcome.html.slim
+++ b/app/views/setup/welcome.html.slim
@@ -8,6 +8,9 @@
 h1 Initial CaaSP Configuration
 = form_for :settings, url: setup_path, method: :put do |f|
   .form-group
+    = f.label :dashboard, "Dashboard location"
+    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control"
+  .form-group
     = f.label :company_name, "Company Name"
     = f.text_field :company_name, value: Pillar.value(pillar: :company_name), class: "form-control", required: true
   .form-group
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -36,19 +36,23 @@ class DashboardController < ApplicationController
   # fields that customer uses) and also skips the redirection to the setup process (when a worker
   # asks for the autoyast profile we will either serve it, or return an error).
   def autoyast
-    @controller_node = request.host
-    begin
-      suse_connect_config = Rails.cache.fetch("SUSEConnect_config") do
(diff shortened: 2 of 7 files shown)


ereslibre:
I am first asking review from @ajaeger and @eotchi for the concept. If that validates we could go on with a technical review.

ajaeger:
I'm not sure I see the full picture here, but this looks sane at first and let's us move forward.

app/views/setup/welcome.html.slim
```diff
@@ -8,6 +8,9 @@
 h1 Initial CaaSP Configuration
 = form_for :settings, url: setup_path, method: :put do |f|
   .form-group
+    = f.label :dashboard, "Dashboard location"
+    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control"
```
MaximilianMeister: is there any chance we can add a note about the default `request.host` and it's implications? could be a flash message, or a tooltip, or sth like a `<span>` above the form field

ereslibre: Its implication highly depends on the contents of `request.host`, while `localhost` will always fail, any other resolvable name or reachable ip address by the cluster will work fine. What I mean: using as dashboard `request.host` isn't bad per-se, it highly depends on the `request.host` actually (e.g. `localhost` will always be a bad idea, for example).

inercia: Do you explicitly check that the `dashboard_ref != "localhost"`? As you said, that would always fail...

ereslibre: I was thinking in doing that in a separate PR...

PR #129, opened 2017-04-25
Remove dashboard pillar seed

ereslibre:


ereslibre:
@mssola I agree. We can seed depending on `Rails.env`. This way on development we don't need to always create an account and is easier to smoke test, and in production we seed with production data (empty for now, but might exist on the future).

My take would be to change `seeds.rb` and only seed if `Rails.env.development?` for now.

PR #130, opened 2017-04-25
Add seeding depending on environment

ereslibre:
Create separate functions so it's very easy to visually identify what
gets seeded when.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,4 +8,17 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -8,4 +8,17 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-User.create email: "test@test.com", password: "password"
+case Rails.env
+when "development", "test"
+  seed_development
```
MaximilianMeister: I think the method needs to be defined before that point, as it's just evaluated from top to bottom i think

ereslibre: You are right, thank you!

PR #132, opened 2017-04-25
Prevent exception in console during initial bootstrap

kiall:
As the data structure changes between bootstrap and monitoring, we
have to account for both in the code. This change will prevent the
exception without affecting behaviour.

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -35,7 +35,7 @@ MinionPoller = {
           $("#bootstrap").prop('disabled', false);
         }
 
-        var unassignedMinions = data.unassigned_minions;
+        var unassignedMinions = data.unassigned_minions || [];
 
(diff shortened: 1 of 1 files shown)


PI-Victor:
seems that there's already a PR to address this https://github.com/kubic-project/velum/pull/82/files

kiall:
@PI-Victor that PR won't fix this issue, as it still includes:

    var unassigned = data.unassigned_minions;

which will cause the exception, as `data` does not have an `unassigned_minions` property.

kiall:
Oh, maybe it will.. (It's been a few years since I touched JS!). Let me look again.

PI-Victor:
i think it does through https://github.com/kubic-project/velum/pull/82/files#diff-398e9dc91ab41e7d183e081c72f8f161R33

PI-Victor:
i don't oppose this change in any way, though :) lgtm.

PR #133, opened 2017-04-25
Internally support selection of multiple masters

kiall:
This change allows for the selection of multiple masters when bootstrapping
the cluster. Note that, this commit does not expose this functionality in
the UI. Allowing users to choose multiple masters will come further down the
line once HA deployments are working etc.

A second commit (which will land on a different branch) will expose this in the
UI for dev/test purposes.

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -5,7 +5,7 @@ require "velum/salt_minion"
 # Minion represents the minions that have been registered in this application.
 class Minion < ApplicationRecord
   # Raised when Minion doesn't exist
-  class NonExistingMinion < StandardError; end
+  class NonExistingNode < StandardError; end
 
   scope :assigned_role, -> { where.not role: nil }
   scope :unassigned_role, -> { where role: nil }
@@ -19,27 +19,21 @@ class Minion < ApplicationRecord
   # Example:
   #   Minion.assign_roles(
   #     roles: {
-  #       master: 1,
+  #       master: [1],
   #       minion: [2, 3]
   #     },
   #     default_role: :dns
   #   )
   def self.assign_roles!(roles: {}, default_role: :minion)
-    requested_master = roles[:master].to_i
-    requested_minions =
-      roles[:minion] || Minion.unassigned_role.pluck(:id) - [requested_master]
-    requested_minions = requested_minions.map(&:to_i)
+    # Lookup selected masters and minions
+    masters = Minion.select_role_members(roles: roles, role: :master)
+    minions = Minion.select_role_members(roles: roles, role: :minion)
 
-    if !requested_master.blank? && !Minion.exists?(id: requested_master)
-      raise NonExistingMinion, "Failed to process non existing minion id: #{requested_master}"
-    end
-    master = Minion.find(requested_master)
-    # choose requested minions or all other than master
-    minions = Minion.where(id: requested_minions)
-
-    # assign master if requested
+    # assign roles to each master and minion
     {}.tap do |ret|
-      ret[master.minion_id] = master.assign_role(:master) if master
+      masters.find_each do |master|
+        ret[master.minion_id] = master.assign_role(:master)
+      end
 
       minions.find_each do |minion|
         ret[minion.minion_id] = minion.assign_role(:minion)
@@ -52,6 +46,21 @@ class Minion < ApplicationRecord
         end
       end
     end
+  end
+
+  # Prepares an ActiveRecord relation which will return all the members
+  # assigned to given role. Additionally, ensures all supplied node IDs
+  # exist within the database at the time of calling.
+  def self.select_role_members(roles: {}, role: nil)
+    node_ids = roles.key?(role) ? roles[role].map(&:to_i) : []
+
+    node_ids.each do |node|
+      unless Minion.exists?(id: node)
+        raise NonExistingNode, "Failed to process non existing node id: #{node}"
+      end
+    end
+
+    Minion.where(id: node_ids)
   end
 
   # rubocop:disable SkipsModelValidations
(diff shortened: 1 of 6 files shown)


kiall:
Hey @flavio - we discussed landing some changes which would be needed for HA into master, even if it wasn't going to be exposed to users soon. I'm thinking this kind of change qualifies, thoughts?

app/models/minion.rb
```diff
@@ -54,6 +63,18 @@ def self.assign_roles!(roles: {}, default_role: :minion)
     end
   end
 
+  def self.select_masters(roles)
+    roles.key?(:master) ? roles[:master].map(&:to_i) : []
+  end
+
+  def self.select_minions(roles, requested_masters)
```
ereslibre: Same here.

app/models/minion.rb
```diff
@@ -54,6 +63,18 @@ def self.assign_roles!(roles: {}, default_role: :minion)
     end
   end
 
+  def self.select_masters(roles)
```
ereslibre: We have been using keyword args, I think it would be good to stick with them for easier extensibility.

app/models/minion.rb
```diff
@@ -19,27 +19,36 @@ class NonExistingMinion < StandardError; end
...
-      roles[:minion] || Minion.unassigned_role.pluck(:id) - [requested_master]
-    requested_minions = requested_minions.map(&:to_i)
+    requested_masters = select_masters(roles)
+    requested_minions = select_minions(roles, requested_masters)
 
-    if !requested_master.blank? && !Minion.exists?(id: requested_master)
-      raise NonExistingMinion, "Failed to process non existing minion id: #{requested_master}"
+    requested_masters.each do |requested_master|
+      unless Minion.exists?(id: requested_master)
+        raise NonExistingMinion, "Failed to process non existing master id: #{requested_master}"
```
mssola: Maybe it makes sense to simply have an exception class called `NonExistingNode`, which covers both masters and workers ? Just saying because it feels a bit weird to raise a `NonExistingMinion` for a master 😉

app/models/minion.rb
```diff
@@ -19,27 +19,36 @@ class NonExistingMinion < StandardError; end
...
-    requested_master = roles[:master].to_i
-    requested_minions =
-      roles[:minion] || Minion.unassigned_role.pluck(:id) - [requested_master]
-    requested_minions = requested_minions.map(&:to_i)
+    requested_masters = select_masters(roles)
+    requested_minions = select_minions(roles, requested_masters)
 
-    if !requested_master.blank? && !Minion.exists?(id: requested_master)
-      raise NonExistingMinion, "Failed to process non existing minion id: #{requested_master}"
+    requested_masters.each do |requested_master|
```
mssola: Note that this piece of code is utterly similar to the one for masters. Moreover, wouldn't it make more sense of the `select_*` methods already covered this ?

app/models/minion.rb
```diff
@@ -19,27 +19,36 @@ class NonExistingMinion < StandardError; end
...
+    requested_minions.each do |requested_minion|
+      unless Minion.exists?(id: requested_minion)
+        raise NonExistingMinion, "Failed to process non existing minion id: #{requested_minion}"
+      end
     end
-    master = Minion.find(requested_master)
-    # choose requested minions or all other than master
+
+    # choose selected masters and minions
+    masters = Minion.where(id: requested_masters)
```
mssola: Again, we have more code duplication: let's merge things into a specific method or something.

app/models/minion.rb
kiall: Sure, will update.

app/models/minion.rb
kiall: Sure, will update.

mssola: This method and the one below can be perfectly merged into a single method like:

```ruby
def self.select_role(ary, key, others: nil)
  if ary.key?(key)
    ary[key].map(&:to_i)
  else
    Minion.unassigned_role.pluck(:id) - others
  end
end
```

And I'm sure it can be further tuned if we merge things up in the main function 😉

mssola: Oh, use keyword arguments as @ereslibre suggested of course 😉

app/models/minion.rb
mssola: `NonExistingRole` maybe ?

kiall: I agree, but it was only ever raised for masters (that I could see at a glace!), so I didn't want to change it unnecessarily. Will update.

app/models/minion.rb
kiall: Would it ever be desirable to select 2 minions explicitly, and have the rest end up as masters? I'm thinking `Minion.unassigned_role.pluck(:id)` can just be removed entirely, allowing the "assign default role if there is any minion left with no role" code to handle this instead.

app/models/minion.rb
kiall: Makes sense, will update.

app/models/minion.rb
mssola: That would make sense for me as it is right now. I don't know if in the future we'd have other roles though. So if you go for this solution, write in a documentation comment this assumption 😉

kiall: If we want this to scale, we'll absolutely need additional roles - or likely - more flexibility in deployment topology. For example (what follows is a wild guess on numbers) - for >30 nodes, etcd may need to be running on dedicated machines.. for >200 nodes, multiple etcd clusters may be needed, each running on their own machines (for kube core stuff, and the second for kube's event storage). Though, this is a long way off and likely shouldn't be given too much thought just yet.

kiall:
Part 2 of this, which actually enables selection of multiple masters by users, can't be merged to GA and lives here for now: https://github.com/kubic-project/velum/commit/d3fc36443d0c22ae7cadeed208c4b7a31546386a

PR #134, opened 2017-04-25
Add /var/log/pods/* to cleanup script

grahamhayes:
Re running the dev environment can fail to start from time to time if the old log files for the pods are still in place.

### DIFF
--- a/kubernetes/cleanup
+++ b/kubernetes/cleanup
@@ -8,5 +8,6 @@ log "Cleaning up old containers"
 for i in `docker ps -a | grep k8s | awk '{print $1}'`; do docker rm -vf $i; done
 
 log "Cleaning the tmp and manifests folders"
-sudo rm -rf tmp/* manifests/* 
+sudo rm -rf tmp/* manifests/*
+sudo rm -rf /var/log/pods/*
 git checkout tmp/ manifests/*


kubernetes/cleanup
```diff
@@ -8,5 +8,6 @@ log "Cleaning up old containers"
 for i in `docker ps -a | grep k8s | awk '{print $1}'`; do docker rm -vf $i; done
 
 log "Cleaning the tmp and manifests folders"
-sudo rm -rf tmp/* manifests/* 
+sudo rm -rf tmp/* manifests/*
+sudo rm -rf /var/log/pods/*
```
ereslibre: Another option is to also redirect this logging somewhere else when we start the kubelet.

PI-Victor: i was looking at `systemd-exec` or `systemd-run` for this script so that you can add on-the-fly arguments to your `kubelet` and wanted to bring it up, but it seemed overkill to me. this way you can redirect the logs to syslog i think. i think, not 100% sure :). do you guys think it would be overkill?

PI-Victor: https://www.freedesktop.org/software/systemd/man/systemd-run.html
nvm about `systemd-exec` that's the execution env.

PR #135, opened 2017-04-26
Make the development environment as close as possible to production

ereslibre:
This task is almost completed. The rails application cannot be started because `bundle` isn't found, because it is a symlink to `/srv/velum/vendor/bundle/ruby/2.1.0/bin/bundler.ruby2.1`, and for development purposes we are mounting velum's source code in `/srv/velum`. I'd like to avoid mounting only specific folders and/or files within the velum project, because this can cause confusion in the future and is hard to track.

Once that this is fixed (I am checking it out), and once that the automation is in place to automatically push containers to `docker-testing-registry.suse.de` the development environment should work fine. In the meanwhile that this automation is in place I already pushed all required sle12 images (`velum`, `salt-master`, `salt-minion`, `salt-api`, `mariadb`) to `docker-testing-registry.suse.de` with their tags, so unless the `Gemfile` or `Gemfile.lock` changes you can still test it (remember to add `docker-testing-registry.suse.de` as insecure registry to your docker options in `/etc/sysconfig/docker`.

Status:
- [x] Remove development manifests
- [x] Remove specific pillar substitutions
- [x] Point to the manifests using CONTAINER_MANIFESTS_DIR envvar along with the start script
- [x] Script that takes production manifests and does the required changes and substitutions to convert them to development manifests
- [x] Make velum gem vendoring work despite source code is mounted in `/srv/velum`
- [x] Tests can be executed and pass
- [x] Fix and add documentation

PR #137, opened 2017-04-26
[WIP] add readiness/liveness probe for mariadb instance.

PI-Victor:
* fix an issue with `skip-networking` in mariadb config.
i think i need to lower the `initialDelaySeconds` to 20 or something.

### DIFF
--- a/kubernetes/mariadb/skip-networking.cnf
+++ b/kubernetes/mariadb/skip-networking.cnf
@@ -1,1 +1,2 @@
+[mysqld]
 skip-networking
--- a/kubernetes/manifest-templates/private.yaml
+++ b/kubernetes/manifest-templates/private.yaml
@@ -10,6 +10,24 @@ spec:
   - name: velum-mariadb
     imagePullPolicy: Always
     image: library/mariadb:10.0.29
+    readinessProbe:
+      exec:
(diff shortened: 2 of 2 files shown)


kubernetes/mariadb/skip-networking.cnf
```diff
@@ -1 +1,2 @@
-skip-networking
\ No newline at end of file
+[mysqld]
+skip-networking
```
ereslibre: Please, consider applying this change also on the production manifests (https://github.com/kubic-project/caasp-container-manifests)

ereslibre: Sorry, this was meant to be for the whole commit, not only this bit. Take into account that very soon this manifests are going to be removed and we will only use the production ones (with some automatic tweaking): https://github.com/kubic-project/velum/pull/135

PI-Victor:
@ereslibre i know they will be removed in favor of the prod ones. the problem with this PR probe is that it doesn't solve any of the issues that i was trying to cover. it's fine to have, just not really what i wanted. i'm closing this for now and will start a new PR in the prod manifests repo directly.
i started it here because it was easier to test with the velum setup.

PR #140, opened 2017-04-28
discovery: only send json responses to AJAX calls

mssola:
When pressing the back/forward buttons in Chrome it sometimes returned a
JSON response for the HTML version. To avoid this, let's be explicit on
this and only allow JSON responses for requests ending with ".json"
(rails convention).

bsc#103174

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -8,6 +8,7 @@ MinionPoller = {
     $.ajax({
       url: $('.nodes-container').data('url'),
       dataType: "json",
+      cache: false,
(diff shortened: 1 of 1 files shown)


ereslibre:
This looks like an issue with caching on the client side. While separating the requests in two types fixes the issue because chrome won't reuse the json cached as the html view and show that, another possibility that I think fits better is to fix this on the client side alone by adding `cache: false` to the jquery ajax request that we launch from `dashboard.js`. Would that make sense?

mssola:
@ereslibre good idea, I didn't realize of that. The solution is way simpler now 😄

PR #144, opened 2017-05-05
Propagate ssh public key from dashboard to workers, bsc#1030876

mjura:


MaximilianMeister:
does this PR depend on https://github.com/kubic-project/caasp-container-manifests/pull/40 ?

app/controllers/dashboard_controller.rb
```diff
@@ -43,6 +43,8 @@ def autoyast
              Velum::SUSEConnect::MissingCredentialsException
         @do_registration = false
       end
+      ssh_key_file = "/var/lib/misc/ssh-public-key/id_rsa.pub"
+      @ssh_public_key = File.file?(ssh_key_file) ? File.read(ssh_key_file) : ""
```
ereslibre: `@ssh_public_key = File.file?(ssh_key_file) ? File.read(ssh_key_file) : nil` might be a better fit. We can also check this `nil`/emptyness on the autoyast profile... see below.

app/views/dashboard/autoyast.xml.erb
```diff
@@ -24,6 +24,18 @@
...
+        <filename>set_authorized_keys.sh</filename>
+        <interpreter>shell</interpreter>
+        <source><![CDATA[
+        #!/bin/sh
+        mkdir -p /root/.ssh
+        chmod 600 /root/.ssh
+        echo "<%= @ssh_public_key %>" > /root/.ssh/authorized_keys
+        ]]>
+        </source>
+      </script>
```
ereslibre: I think we can avoid writing this at all if we haven't a ssh key, so we could do something like this:

```
<% unless @ssh_public_key.blank? %>
      <script>
        <chrooted config:type="boolean">true</chrooted>
        <filename>set_authorized_keys.sh</filename>
        <source><![CDATA[
        #!/bin/sh
        mkdir -p /root/.ssh
        chmod 600 /root/.ssh
        echo "<%= @ssh_public_key %>" > /root/.ssh/authorized_keys
        ]]>
        </source>
      </script>
<% end %>
```

ereslibre:
This still concerns me a bit about HA, since it is yet another file in disk that is only available to one administration dashboard machine... We will need to figure this out along with the application secrets...

LGTM :+1: addressing the changes (so we don't write an empty key if that happens, and that we only really care about having read the contents of the file).

app/views/dashboard/autoyast.xml.erb
mjura: ok, done

app/controllers/dashboard_controller.rb
mjura: Done

mjura:
@ereslibre: ok, I've addressed your comments, thank you for suggestions

PR #146, opened 2017-05-08
Gemfile: fix ENV

MaximilianMeister:
we are setting IGNORE_ASSETS=no from the spec file
that's why it was working, but it is not correct

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem "kubeclient", "~> 2.3.0"
 #
 # IGNORE_ASSETS=yes bundle list
 
-unless ENV["VELUM_IGNORE_ASSETS"] == "yes"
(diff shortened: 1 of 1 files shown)


ereslibre:
Hmm... if we change this, then https://build.suse.de/package/view_file/Devel:CASP:1.0:ControllerNode/sles12sp2-velum-docker-image/sles12sp2-velum-docker-image.kiwi?expand=1 would need updating too (the envvar)

MaximilianMeister:
> Hmm... if we change this, then https://build.suse.de/package/view_file/Devel:CASP:1.0:ControllerNode/sles12sp2-velum-docker-image/sles12sp2-velum-docker-image.kiwi?expand=1 would need updating too (the envvar)

good point, I'll send a SR
thanks for catching this! i was looking for a place where this was actually ignored, but i haven't looked there

PR #148, opened 2017-05-08
WIP: Ensure that gem dependencies versions are the same in development/test and production

ereslibre:
By removing code conditionals on the Gemfile we ensure that a restriction
that (dis)appears in production don't shift dependencies versions. Is one
of the good points of bundler, that we can use groups to avoid installing
unnecessary gems in certain environments, but the versions of the gems
are the same both in development/test and production.

This makes bug easier to reproduce if they are related to dependencies.
We can still use groups to only install the groups that we want, or
discard groups that we won't need (--without development test in production
for instance). But we are sure that the installed gems are always the same
versions in all environments.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,52 +21,30 @@ gem "gravatar_image_tag"
 gem "devise"
 gem "kubeclient", "~> 2.3.0"
 
-# Assets group.
-#
-# Do not set it or set it to no when precompiling the assets.
-#
-# IGNORE_ASSETS="no" RAILS_ENV=production bundle exec rake assets:precompile
-#
-# Set IGNORE_ASSETS to YES when creating the Gemfile.lock for
-# production after having precompiled the assets
-# run:
-#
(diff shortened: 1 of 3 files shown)


MaximilianMeister:
This needs some input from @jordimassaguerpla as well. It seems though that we have this kind of way, as it was before this patch for some reason. see https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/Gemfile#L65 or Portus

a hint and some more info is in https://github.com/SUSE/Portus/commit/2fd175256d4e25d70c4a5bac09d0e6878ca92907

mssola:
Sounds good to me in principle, but I'd like to hear from @jordimassaguerpla 😃

ereslibre:
I will handle this in a different way. We should discuss the implications of having a `Gemfile.lock` that changes depending on the environment. However I think that it might not be as troublesome as in other environments (here a gem change would trigger a dependency fail on buildtime on OBS if a gem version changes and we weren't predicting it).

PR #149, opened 2017-05-08
packaging: build velum as noarch

MaximilianMeister:
there's nothing architecture-specific about the velum package
and we can't build for all arches if we deposit files under /usr/share
which is reserved for architecture independent files


MaximilianMeister:
we'll go for a revert to `/srv/velum/` instead

PR #151, opened 2017-05-09
[WIP] Move to init containers

PI-Victor:
This change will take advantage of kubernetes init containers for setting up velum
What this means is that 
a) no application containers are started before the checks inside the init containers are done
b) once the database is up, in the second init container, we execute `bundle exec rake db:create && bundle exec rake db:schema:load && bundle exec rake db:migrate` before we start the `velum-event-processor`. This solves the current issue with the even-processor crashing and restarting because it can't reach the database.  
We can now ditch the `init` script together with showing the page. I think we can also ditch any other checks that we do.

This is here, and not in the caasp-manifests repo because it's meant to be a test for now, since i'm not sure what else it would impact.
It's a work in progress because, for the life of me, i can't understand why the check at https://github.com/kubic-project/velum/compare/master...PI-Victor:move_to_init_container?expand=1#diff-f1f577cbb090fc9738404e4d85e9c1ddR19 fails with 
```shell
https://github.com/kubic-project/velum/compare/master...PI-Victor:move_to_init_container?expand=1#diff-f1f577cbb090fc9738404e4d85e9c1ddR19
```
unless i put a different command before the execution of the `until`.
I didn't add the removal of the init script and other checks because i'm not sure if we use it for anything else.

### DIFF
--- a/kubernetes/manifest-templates/public.yaml
+++ b/kubernetes/manifest-templates/public.yaml
@@ -16,7 +16,7 @@ spec:
     - name: MARIADB_SOCKET
       value: /var/run/mysqld/mysqld.sock
     command: ["sh", "-c"]
-    args: ["until mysqladmin -p\"${MYSQL_ROOT_PASSWORD}\" -S \"${MARIADB_SOCKET}\" ping; do echo 'Database not online yet, waiting...' && sleep 3; done"]
+    args: ["ls && until mysqladmin -p\"${MYSQL_ROOT_PASSWORD}\" -S \"${MARIADB_SOCKET}\" ping; do echo 'Database not online yet, waiting...' && sleep 3; done"]
     volumeMounts:
       - mountPath: /var/run/mysqld
         name: mariadb-unix-socket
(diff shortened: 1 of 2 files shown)


PI-Victor:
debugging them can be done with 
`docker logs -f k8s_init-rake-setup_velum-public-127.0.0.1`
`docker logs -f k8s_init-mariadb_velum-public-127.0.0.1`

inercia:
I guess we wouldn't need [this setup](https://github.com/kubic-project/terraform/blob/master/provision/dashboard.sh#L244) in our Terraform code once we have this code in the `caasp-containers-manifests` repo, right?

PI-Victor:
@ereslibre  thanks, so... i'll probably just close this off and open it in the caasp-manifests repo directly then.
@inercia no, i don't think we would need that anymore. also with a bit of tinkering, maybe we can eliminate this whole waiting for containers to start. https://github.com/kubic-project/terraform/blob/master/provision/dashboard.sh#L232-L247. need to see why the tests are failing, i see the even process was crashing.

inercia:
> @inercia no, i don't think we would need that anymore. also with a bit of tinkering, maybe we can eliminate this whole waiting for containers to start. https://github.com/kubic-project/terraform/blob/master/provision/dashboard.sh#L232-L247. 

That would be great, that would make both environments look much more similar.

PI-Victor:
awesome! closing this for now.

PR #160, opened 2017-05-18
Add setting for the main apiserver hostname

ereslibre:
Fixes bsc#1039437

### DIFF
--- a/app/views/setup/welcome.html.slim
+++ b/app/views/setup/welcome.html.slim
@@ -9,7 +9,10 @@ h1 Initial CaaSP Configuration
 = form_for :settings, url: setup_path, method: :put do |f|
   .form-group
     = f.label :dashboard, "Dashboard location"
-    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control"
+    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control", required: true
+  .form-group
+    = f.label :apiserver, "External Kubernetes API server FQDN"
+    = f.text_field :apiserver, value: Pillar.value(pillar: :apiserver), class: "form-control", required: true
   .form-group
     = f.label :company_name, "Company Name"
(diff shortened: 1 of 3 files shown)


lib/velum/kubernetes.rb
```diff
@@ -22,7 +22,7 @@ def self.kubeconfig
                                           targets:     "roles:kube-master",
                                           target_type: "grain",
                                           arg:         "cat /etc/pki/minion.key"
-      host = Minion.master.applied.pluck(:fqdn).first
+      host = Pillar.value(pillar: :apiserver) || Minion.master.applied.pluck(:fqdn).first
```
kiall: It looks like we've added the new field as required, so the pillar value should always exist. Should we remove the `|| Minion.master.applied.pluck(:fqdn).first`?

app/views/setup/welcome.html.slim
```diff
@@ -9,7 +9,10 @@ h1 Initial CaaSP Configuration
 = form_for :settings, url: setup_path, method: :put do |f|
   .form-group
     = f.label :dashboard, "Dashboard location"
-    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control"
+    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control", required: true
+  .form-group
+    = f.label :apiserver, "External Kubernetes API server hostname"
```
kiall: FQDN is likely better than hostname, I know, I know - this name was my suggestion, my bad! A hostname is `foo`, a FQDN is `foo.mycompany,com`.

ereslibre: Yeah... but technically anything the customer set as resolvable on his/her side will work (whether it's a hostname or a fqdn). I am OK with showing fqdn here though.

lib/velum/kubernetes.rb
ereslibre: Yes, I will remove it, it was to be extra safe here, but it doesn't make sense if we require it already.

PR #165, opened 2017-05-26
setup: Block non-alphabetical characters

mssola:
I've also fixed a redirect.

See bsc#1031475

### DIFF
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -33,7 +33,7 @@
 #   - https://github.com/kubic-project/salt/blob/master/config/master.d/returner.conf
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
-  validates :value, presence: true
+  validates :value, presence: true, alphabetic: true
 
   scope :global, -> { where minion_id: nil }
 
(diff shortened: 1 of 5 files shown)


app/validators/alphabetic_validator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    record.errors[attribute] << "contains invalid characters"
+  end
+
+  protected
+
+  # Returns true if the given value belongs to a protected key.
+  def protected?(pillar)
+    keys = Pillar.all_pillars
+    PROTECTED_VALUES.any? { |v| keys[v] == pillar }
```
ereslibre: Can this be simplified as `PROTECTED_VALUES.include? pillar` ?

app/models/pillar.rb
```diff
@@ -54,5 +54,22 @@ def all_pillars
...
+    # encountered errors.
+    def apply(pillars)
+      errors = []
+
+      Pillar.all_pillars.each do |key, pillar_key|
+        pillar = Pillar.find_or_initialize_by pillar: pillar_key
+        pillar.value = pillars[key]
+        next if pillar.save
+
+        exp = pillar.errors.messages.empty? ? "" : ": #{pillar.errors.messages[:value].first}"
```
ereslibre: The test can be done with `pillar.errors.empty?`

ereslibre: In the near future we can even add `pillar.errors.add(pillars[key], ...` to add errors to their specific virtual pillar attribute, and we could highlight specific fields on the forms, if required.

app/validators/alphabetic_validator.rb
mssola: Not really. That's what I tried at first, but bear in mind that the `PROTECTED_VALUES` constant only contain the key for which we identify the different pillar keys, which are the ones to be compared then after. A bit confusing, but we cannot simplify it :P

PR #166, opened 2017-05-29
Fixed footer overlapping elements

vitoravelino:
If a page didn't fit users screen height, the footer overlaps any
element since its position is absolute to .container element.

An example of that was the "Next" button on initial configuration form.

See bsc#1041440


mssola:
Thanks a lot 👏

PR #167, opened 2017-05-30
Split js code into different files

vitoravelino:
This refactoring doesn't introduce many things. Most of the code
was just separated into different files due to its reponsibilities.

Also, an EventBus object was created to make possible multiple
object communicate with each other without problems.

The purpose of this is to allow a healthier way to maintain/implement new features until we move to a modern library.


vitoravelino:
I didn't rewrite anything, just moved some pieces around to become easier to work on it until we move to a more robust organization.

Weird, tests didn't pass on travis. I'll re-check... it was passing here...

PR #168, opened 2017-05-31
signup: added proper validation message for the password fields

mssola:
See bsc#1041905

### DIFF
--- a/app/views/devise/registrations/new.html.slim
+++ b/app/views/devise/registrations/new.html.slim
@@ -25,9 +25,9 @@
       .form-group
         = f.email_field :email, autofocus: true, class: 'form-control', placeholder: 'Introduce your Email', required: true
       .form-group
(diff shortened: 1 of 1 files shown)


vitoravelino:
I'm not a fan of writing js direct in views templates but since we didn't decide yet how things are gonna be on the front-end and this is really simple, I'm good with this.

mssola:
> I'm not a fan of writing js direct in views templates but since we didn't decide yet how things are gonna be on the front-end and this is really simple, I'm good with this.

Agreed. We should change this whenever we move to a more modern frontend 👍 

> It looks a bit hard to manage, do you think it makes sense to use a view helper here?

This will be refactored away whenever we modernize the frontend after the 1.0 release, and it will most surely be handled purely from the JS side, so I don't think it's worth it.

PR #169, opened 2017-05-31
Don't allow users to download the kubeconfig when bootstrapping

mssola:
Unless all the nodes are up and ready, don't allow users to download the
kubeconfig file.

See bsc#1041806


mssola:
We probably shouldn't be so strict (not allowing to download this file when we've already bootstrapped but then a node fails on an update). What do you think ?

ereslibre:
I don't feel comfortable fixing this on the UI side. We already had this check on the backend side and it was working.

I could reproduce this by deploying the cluster. Then went to the kubernetes-master for example, `mv /etc/pki/minion.crt /etc/pki/minion.crt.bak` and then run a rails session:

```
require "velum/kubernetes"
Velum::Kubernetes.kubeconfig
```

And see the result:

```
=> #<struct Velum::Kubernetes::KubeConfig host="master.ereslibre.io", ca_crt="valid_ca_cert", client_crt="cat: /etc/pki/minion.crt: No such file or directory", client_key="valid_key">
```

That is completely wrong, and `nil` should have been returned on the `client_crt` field. We already have the check for only allowing to download the kubeconfig file when all certificates are ready (even before the orchestration finished, when certs are ready you will be able to download the kubeconfig file). We do this check here: https://github.com/kubic-project/velum/blob/432ec7784e8cb6260edc353f68df5af203ae0a76/app/controllers/dashboard_controller.rb#L67-L73

At some moment we started to return the `client_crt` contents `cat: /etc/pki/minion.crt: No such file or directory` as if it was the certificate itself... what is obviously wrong...

Summing up, I don't think we need another check UI-wide, but even if we add it, we should fix the backend problem.

ereslibre:
So my take here would be to revisit https://github.com/kubic-project/velum/blob/432ec7784e8cb6260edc353f68df5af203ae0a76/lib/velum/kubernetes.rb#L14-L24 and see what happens when they don't exist, and return `nil` on those cases, so the controller will find out and will show the flash message.

vitoravelino:
@ereslibre Even if we have a problem on the back-end that should be fixed, it would be nice to let the  the user know that he's only capable of downloading the kubeconfig if all the nodes are bootstrapped.

ereslibre:
> @ereslibre Even if we have a problem on the back-end that should be fixed, it would be nice to let the the user know that he's only capable of downloading the kubeconfig if all the nodes are bootstrapped.

Agreed, it's nice to have this visual feedback on the UI side, but I'd love to see the backend part fixed now. To be honest, I don't know in which moment this stopped working properly.

vitoravelino:
Totally agree on fixing the back-end part also. :+1:

mssola:
Current checklist:

- [x] Don't allow this from the frontend.
- [ ] Fix backend, so this is not allowed.
- [ ] Provide feature tests

lib/velum/salt.rb
```diff
@@ -34,5 +34,20 @@ def self.orchestrate
...
+      _, data = Velum::Salt.call action:      "cmd.run",
+                                 targets:     targets,
+                                 target_type: target_type,
+                                 arg:         "cat #{file}"
+
+      data["return"].map do |el|
+        val = el.values.first
+
+        # TODO: improve error handling...
+        val.include?("No such file or directory") ? nil : val
```
mssola: @ereslibre do you know any better way to do this ? It looks like salt-api is simply returning the stdout and that's it. No retcode, or job id, no nothing 😕 I've also tried other functions like `cp.get_file_str` but I've got no luck...

mssola:
@ereslibre @vitoravelino you can review it again 😋

app/assets/javascripts/dashboard.js
```diff
@@ -11,6 +11,8 @@ MinionPoller = {
       cache: false,
       success: function(data) {
         var rendered = "";
+        var all_applied = 1;
```
vitoravelino: :point_up:

app/assets/javascripts/dashboard.js
```diff
@@ -54,6 +61,16 @@ MinionPoller = {
...
 
+  enable_kubeconfig: function(enable) {
+    if (enable) {
+      $("#download-kubeconfig").unbind('click')
+      $("#download-kubeconfig").removeAttr("disabled");
+    } else {
+      $("#download-kubeconfig").bind('click', function(e) { e.preventDefault(); });
+      $("#download-kubeconfig").attr("disabled", true);
+    }
+  },
```
vitoravelino: :point_up:

lib/velum/salt.rb
ereslibre: I am going to do a fast checking, but if this is what you got I think this is the only way... Just double checking just in case...

app/assets/javascripts/dashboard.js
mssola: > Use .prop('disabled', !enabled) instead of attr and removeAttr

Weird, it didn't work for me, that's why I did it 😕 

> Is the bind and unbind really necessary?

Even if disabled, I could still click it and then I got the download dialog... So it looks like it...

mssola: > I would suggest to rename enable to enabled

Done 😉

ereslibre:
There's the option to also use `cp.get_file_str` (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.cp.html#salt.modules.cp.get_file_str), but still... even with a non-existent file it returns a `200` http result and we would need to parse the response...

Are we missing anything when using the salt api? This seems all so strange to me...

app/assets/javascripts/dashboard.js
vitoravelino: > Weird, it didn't work for me, that's why I did it :confused:

Just realized that it's an anchor using `.btn-link`. That's why it didn't work. You can just use `.attr('disabled', !enabled)` and jQuery will automatically remove if `false`.

> Even if disabled, I could still click it and then I got the download dialog... So it looks like it...

I would suggest a global event listener for all elements with `disabled` attribute like

```js
$('body').on('click', '[disabled]', function(e) {
  e.preventDefault();
});
```

It will make this method cleaner.

mssola: Done, it works beautifully, thanks 👍

mssola:
@vitoravelino tests are green again. You can review again 😉

PR #171, opened 2017-05-31
spec: added newer phantomjs version from gem

vitoravelino:


ereslibre:
Isn't this going to be a problem when packaging Velum? We don't have this gem packaged. @jordimassaguerpla?

jordimassaguerpla:
@ereslibre good point, but it this case, since this is in the "unless PACKAGING" clause, we don't need this as an RPM.

jordimassaguerpla:
lgtm

vitoravelino:
I'm only doing this way because this was the cleanest way I've found to use the latest version of phantomjs in Travis. I already experienced some errors on Portus because of not using the latest version. Eventually we will need this too.

PR #172, opened 2017-06-01
UI: show salt ID instead of rails' internal ID [bsc#1041789]

flavio:
Change the UI to show the salt ID (which now is set to be equal to the machine-id) instead of showing Rails' internal ID.

This is required to fix bsc#1041789

Note well: I'm still working on setting up the devenv to test this code, but I didn't want to hold the code on my machine.

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -88,7 +88,7 @@ MinionPoller = {
 
     return "\
       <tr> \
-        <th>" + minion.id +  "</th>\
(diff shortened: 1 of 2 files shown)


ereslibre:
Just verified it. Works as expected when discovering and also during the orchestration and when it finishes.

ereslibre:
Just checked `rubocop` and `rspec`. All is fine, Travis' errors are artifacts.

Thank you! Merging...

PS: As a side note, I think if the minion-id was shown with a fixed-width font it would look even better, but we're good to go :+1:

PR #173, opened 2017-06-01
Allow selection of nodes to bootstrap cluster

vitoravelino:
Previously, all the nodes found were used to be part of the cluster.
Now, user needs to select which ones they want to be used for it.

See bsc#1042193

- [X] Implement UI
- [x] Features' tests
- [ ] e2e test

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,4 +1,6 @@
 MinionPoller = {
+  selectedNodes: [],
+
   poll: function() {
     this.request();
   },
@@ -214,19 +216,29 @@ MinionPoller = {
 
   renderDiscovery: function(minion) {
     var masterHtml;
-    var checked;
+    var masterChecked = '';
+    var minionHtml;
+    var minionChecked = '';
 
     if (MinionPoller.selectedMasters && MinionPoller.selectedMasters.indexOf(minion.id) != -1) {
-      checked = "checked";
-    } else {
-      checked = '';
+      masterChecked = 'checked';
+      minionChecked += 'disabled="disabled" checked ';
     }
+
     masterHtml = '<input name="roles[master][]" id="roles_master_' + minion.id +
-      '" value="' + minion.id + '" type="radio" ' + checked + '>';
+      '" value="' + minion.id + '" type="radio" ' + masterChecked + '>';
+
+    if (MinionPoller.selectedNodes && MinionPoller.selectedNodes.indexOf(minion.id) != -1) {
+      minionChecked += 'checked';
+    }
+
+    minionHtml = '<input name="roles[worker][]" id="roles_minion_' + minion.id +
+      '" value="' + minion.id + '" type="checkbox" ' + minionChecked + '>';
 
     return "\
       <tr> \
-        <th>" + minion.minion_id +  "</th>\
+        <td>" + minionHtml +  "</td>\
+        <td>" + minion.minion_id +  "</td>\
         <td>" + minion.fqdn +  "</td>\
         <td class='text-center'>" + masterHtml + "</td>\
       </tr>";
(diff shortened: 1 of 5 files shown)


flavio:
@ereslibre could you take a look at the PR too?

@vitoravelino can you merge it once you have 2 approvals?

flavio:
Also, has someone looked into why the e2e tests are failing right now? Is that something related with this PR?

flavio:
[Bug 1042193 - Allow removal or exclusion of cluster nodes for bootstrapping the cluster](https://trello.com/c/9iB5QLRP/80-bug-1042193-allow-removal-or-exclusion-of-cluster-nodes-for-bootstrapping-the-cluster)

ereslibre:
I can find a pattern that can lead to unexpected behavior:

Click `master` on the radio button in `node1`, unselect the `node1` checkbox. How would this behave?

I guess it would make sense to completely disable the current `master` selected node, so it's not possible to uncheck it at all.

ereslibre:
As another comment: if I click sequentially nodes using the `master` radio button, each row is kept selected. Would this be a desired behavior?

vitoravelino:
> I guess it would make sense to completely disable the current master selected node, so it's not possible to uncheck it at all.

Agreed.

> As another comment: if I click sequentially nodes using the master radio button, each row is kept selected. Would this be a desired behavior?

I don't know. If I deselect when changing the master, it will deselect each one I had selected previously. Since we will provide a solution to add unassigned nodes, I guess it makes more sense to leave deselected.

@flavio @eotchi your input is appreciated on this.

ereslibre:
> I don't know. If I deselect when changing the master, it will deselect each one I had selected previously. Since we will provide a solution to add unassigned nodes, I guess it makes more sense to leave deselected.

As an idea, maybe when the radio button when selecting the `master` automatically checks the checkbox (and possibly disables it), it could also save the previous value of the checkbox, so when another `master` is selected, the previous checkbox is returned to its original state (it could have been checked or unchecked). Apart from restoring it to its original state (when the master selection changes to another node), it should make the checkbox selectable again (if we are disabling it for the current master selection).

vitoravelino:
@ereslibre Changed the behavior to what you suggested. I guess that's the way to go. Thanks! :)

mssola:
Things are bit buggy at least from my side (using Firefox):

- If I try to deselect a master, it takes a bit of time to actually deselect.
- By playing around I got inconsistent states. For instance, try to first select a master, then check the checkbox on the left for this node, unselect master, select again. Go on and play with this logic and you'll end up in situations like: I check the checkbox on the left, then everything is checked (even the master), but then it automatically unselects again by itself. It's really weird, and I don't have a clear way to reproduce it. Simply play with selecting/unselecting the different elements and you'll get there 😕

vitoravelino:
Probably broke when I rebased. Will fix. Today I'll prioritize this PR since there are a few PRs depending on this.

ereslibre:
It looks like the polling cleans up the selections, as far as I could see.

vitoravelino:
Yeah, I actually messed up on the `update-ui` branch because I didn't realize for a moment that the code was being shared with 2 pages with different logic and that would break one of them. Now I'm missing #167. We also need to start testing js itself and don't rely only on feature specs. Anyway, I'll open another PR to fix it.

ereslibre:
The orchestration not working seemed like an artifact (not enough memory) in my laptop. With the workstation selecting a subset of all possible minions I got a green deployment with only the selected workers (left one out of the cluster that was discovered).

When the UI hiccups are fixed, this will be a +1 from me, please confirm when we can retry.

PI-Victor:
just a friendly reminder that this PR also affects a P1 bug and it should be a priority.

ereslibre:
This PR needs (for a clean e2e tests run):

- https://github.com/kubic-project/e2e-tests/pull/59/
- https://github.com/kubic-project/salt/pull/122

ereslibre:
Also, we should add another test case checking only a subset of the available minions, in order to properly test this.

ereslibre:
I can confirm that after running 3 times the following command in my local e2e tests (with [this change applied](https://github.com/kubic-project/e2e-tests/pull/59)): `VERBOSE=1 SALT_BRANCH=pull/122/head VELUM_BRANCH=pull/173/head rspec spec/**/*`, all deployments were a success:

```
Finished in 4 minutes 54.3 seconds (files took 0.47399 seconds to load)
1 example, 0 failures
```

PR #177, opened 2017-06-02
Added Docker HTTP proxy settings

mssola:
See kubic-project/salt#66
See bsc#1036627

### DIFF
--- a/app/views/setup/welcome.html.slim
+++ b/app/views/setup/welcome.html.slim
@@ -14,6 +14,15 @@ h1 Initial CaaSP Configuration
     = f.label :apiserver, "External Kubernetes API server FQDN"
     = f.text_field :apiserver, value: Pillar.value(pillar: :apiserver), class: "form-control", required: true
   .form-group
+    = f.label :http_proxy, "HTTP proxy for Docker"
+    = f.url_field :http_proxy, value: Pillar.value(pillar: :http_proxy), class: "form-control"
+  .form-group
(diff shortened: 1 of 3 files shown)


mssola:
Note that I had to massage a bit the code, because it assumed that all pillar values are to be set (which is not the case for the optional proxy values...)

app/views/setup/welcome.html.slim
```diff
@@ -14,6 +14,15 @@ h1 Initial CaaSP Configuration
...
     = f.text_field :apiserver, value: Pillar.value(pillar: :apiserver), class: "form-control", required: true
   .form-group
+    = f.label :http_proxy, "HTTP proxy for Docker"
+    = f.url_field :http_proxy, value: Pillar.value(pillar: :http_proxy), class: "form-control"
+  .form-group
+    = f.label :https_proxy, "HTTPS proxy for Docker"
+    = f.url_field :https_proxy, value: Pillar.value(pillar: :https_proxy), class: "form-control"
+  .form-group
+    = f.label :no, "No-proxy settings for Docker"
+    = f.text_field :no_proxy, value: Pillar.value(pillar: :no_proxy), class: "form-control"
```
ereslibre: I think we can explain somewhere that this accepts a list of values separated by commas

mssola:
@ereslibre I've opened #179 exactly for that. This is something that can be done later I think.

ereslibre:
> I've opened #179 exactly for that. This is something that can be done later I think.

LGTM, merging.

PR #180, opened 2017-06-07
Add "update workers" button to velum.

grahamhayes:
* Adds a button that will trigger the update orchestration
  run, when there is nodes that can be updated.
* Updates the status column to show status of updates.
* Updated the Salt Event listener to update status of minions
  marked as "failed" if another highstate succeeds

screenshot from 2017-06-13 13-00-48

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -31,6 +31,8 @@ MinionPoller = {
       success: function(data) {
         var rendered = "";
         var allApplied = true;
+        var updateAvailable = false;
+        var updateAvailableNodeCount = 0;
 
         // In discovery, the minions to be rendered are unassigned, while on the
         // dashboard we don't want to render unassigned minions but we still
@@ -61,15 +63,25 @@ MinionPoller = {
           if (minions[i].highstate != "applied") {
             allApplied = false;
           }
+          if (minions[i].update_status == 1) {
+            updateAvailable = true;
+            updateAvailableNodeCount++;
+          }
         }
         $(".nodes-container tbody").html(rendered);
 
         MinionPoller.handleAdminUpdate(data.admin || {});
 
+        if (updateAvailable && allApplied) {
+          $("#update-all-nodes").attr('disabled', false);
+        }
+
         // disable bootstrap button if there are no minions
         $("#bootstrap").prop('disabled', minions.length === 0);
 
         MinionPoller.enable_kubeconfig(minions.length > 0 && allApplied);
+
+        $('#out_dated_nodes').text(updateAvailableNodeCount)
 
         $('.assigned-count').text(minions.length);
         $('.master-count').text(MinionPoller.selectedMasters.length);
@@ -156,10 +168,35 @@ MinionPoller = {
     masterHtml = '<input name="roles[master][]" id="roles_master_' + minion.id +
       '" value="' + minion.id + '" type="radio" disabled="" ' + checked + '>';
 
+    statusText = ''
+
+    switch(minion.update_status) {
+      case 1:
+        switch (minion.highstate) {
+          case "applied":
+            statusHtml = '<i class="fa fa-arrow-circle-up text-info fa-2x" aria-hidden="true"></i>';
+            statusText = 'Update Available'  
+            break;
+          case "failed":
+            statusHtml = '<i class="fa fa-arrow-circle-up text-warning fa-2x" aria-hidden="true"></i>';
+            statusText = 'Update Failed - Retryable'
+            break;
+          case "pending":
+            statusText = 'Update in progress'
+            break;
+        }
+        break;
+      case 2:
+        statusHtml = '<i class="fa fa-arrow-circle-up text-danger fa-2x" aria-hidden="true"></i>';
+        statusText = 'Update Failed'
+        break;
+    }
+
     return "\
       <tr> \
+        <td>" + statusHtml +  "</td>\
+        <td>" + statusText +  "</td>\
         <th>" + minion.minion_id +  "</th>\
-        <td class='text-center'>" + statusHtml +  "</td>\
         <td>" + minion.fqdn +  "</td>\
         <td>" + (minion.role || '') +  "</td>\
         <td class='text-center'>" + masterHtml + "</td>\
--- a/app/controllers/salt_controller.rb
+++ b/app/controllers/salt_controller.rb
@@ -1,0 +1,15 @@
+require "velum/salt"
+# SaltController holds methods for triggering updates of nodes
+class SaltController < ApplicationController
+  skip_before_action :redirect_to_setup
+
+  def update
+    Minion.mark_pending_update
+    Velum::Salt.update_orchestration
+
+    respond_to do |format|
+      format.html { redirect_to root_path }
+      format.json { head :ok }
+    end
+  end
+end
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -101,6 +101,18 @@ class Minion < ApplicationRecord
   end
   # rubocop:enable SkipsModelValidations
 
+  # Updates all nodes with a grain of `tx_update_reboot_needed: True` with a 
+  # highstate = pending, and persists it to the database
+  def self.mark_pending_update
+    needed, failed = ::Velum::Salt.update_status(targets: "*", cached: true)
+    minions = Minion.assigned_role
+    minions.each do |minion|
+      if Minion.computed_status(minion.minion_id, needed, failed) == Minion.statuses[:update_needed]
+        Minion.find_by(minion_id: minion.minion_id).update(highstate: Minion.highstates[:pending])
+      end
+    end
+  end
+
   # Returns the proxy for the salt minion
   def salt
     @salt ||= Velum::SaltMinion.new minion_id: minion_id
--- a/app/models/salt_handler/minion_highstate.rb
+++ b/app/models/salt_handler/minion_highstate.rb
@@ -25,10 +25,11 @@ class SaltHandler::MinionHighstate
 
     # After applying a highstate once, it is applied again and again.
     # Salt probably tries to make sure it stays applied. We don't need to process
-    # those events. We only need to process a highstate if one is pending.
+    # those events. We only need to process a highstate if one is pending, or has
+    # failed in the past.
     minion = Minion.find_by(
       minion_id: minion_id,
-      highstate: Minion.highstates[:pending]
+      highstate: [Minion.highstates[:pending], Minion.highstates[:failed]]
     )
     return false unless minion
 
--- a/app/views/dashboard/index.html.slim
+++ b/app/views/dashboard/index.html.slim
@@ -22,10 +22,14 @@ h1 Cluster Status
             dd.unassigned-count
         .col-md-6.right-column
           dl.side-by-side
-            dt Node updates
-            dd ?
+            dt Updates
+            dd Manual
             dt # of nodes w/ outdated software
-            dd.outdated-count ?
+            dd.outdated-count
+              span#out_dated_nodes 0
+              = link_to update_path, id: "update-all-nodes", class: "btn btn-xs btn-default pull-right", disabled: true, method: :post do
+                  i.fa.fa-arrow-circle-up
+                  | Update All Nodes
     .panel-footer.admin-outdated-notification.hidden
       .message
         i.fa.fw.fa-exclamation-circle
@@ -56,15 +60,17 @@ h1 Cluster Status
           thead
             tr
               th
-                | Id
-              th.text-center
+                | &nbsp;
+              th
                 | Status
+              th
+                | ID
               th
                 | Hostname
               th
                 | Role
-              th.text-center
-                | Master
+              th
+                | &nbsp;
           tbody
 
 = render "dashboard/update_admin_modal"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -17,6 +17,7 @@ Rails.application.routes.draw do
   get "/autoyast", to: "dashboard#autoyast"
   get "/kubectl-config", to: "dashboard#kubectl_config"
   get "/_health", to: "health#index"
+  post "/update", to: "salt#update"
 
   namespace :setup do
     get "/", action: :welcome
--- a/lib/velum/salt.rb
+++ b/lib/velum/salt.rb
@@ -50,6 +50,15 @@ module Velum
       [res, JSON.parse(res.body)]
     end
 
+    # Call the update orchestration
+    def self.update_orchestration
+      res = perform_request(endpoint: "/run", method: "post",
+                            data: { client: "runner_async",
+                                    fun:    "state.orchestrate",
+                                    mods:   "orch.update" })
+      [res, JSON.parse(res.body)]
+    end
+
     # Returns the contents of the given file.
     def self.read_file(targets: "*", target_type: "glob", file: nil)
       _, data = Velum::Salt.call action:      "cmd.run",
--- a/spec/controllers/salt_controller_spec.rb
+++ b/spec/controllers/salt_controller_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe SaltController, type: :controller do
+  let(:user)                 { create(:user) }
+  let(:master_minion)        { create(:master_minion) }
+  let(:worker_minion)        { create(:worker_minion) }
+  let(:stubbed) do
+    [
+      [{ "admin" => "",   master_minion.minion_id => true, worker_minion.minion_id => true }],
+      [{ "admin" => true, master_minion.minion_id => true, worker_minion.minion_id => ""   }]
+    ]
+  end
+  before do
+    sign_in user
+    setup_stubbed_update_status!(stubbed: stubbed)
+    allow(::Velum::Salt).to receive(:call).and_return(true)
+  end
+
+  describe "POST /update" do
+    it "gets an 302 response" do
+      VCR.use_cassette("salt/update_orch", record: :none) do
+        post :update
+        expect(response.status).to eq 302
+      end
+    end
+  end
+end
--- a/spec/vcr_cassettes/salt/update_orch.yml
+++ b/spec/vcr_cassettes/salt/update_orch.yml
@@ -1,0 +1,48 @@
+---
+http_interactions:
+- request:
+    method: post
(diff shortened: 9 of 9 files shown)


app/controllers/concerns/discovery.rb
```diff
@@ -1,18 +1,47 @@
...
+
+    minions = perform_request(endpoint: "/minions", method: "get")
+
+    #logger.info JSON.parse(minions.body)["return"]
+
+    ret = JSON.parse(minions.body)["return"]
+
+    for minion in ret do
+      minion.each do | key, value |
+        if value["tx_update_reboot_needed"]
```
ereslibre: This `if-else` branch can be simplified as `tx_update_reboot_needed[key] = value["tx_update_reboot_needed"]`

app/controllers/concerns/discovery.rb
```diff
@@ -1,18 +1,47 @@
...
+          tx_update_reboot_needed[key] = true        
+        else
+          tx_update_reboot_needed[key] = false
+        end
+        if value["tx_update_failed"]
```
ereslibre: Same happens here, we can simplify it.

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    Velum::Salt.update_orchestration
+
+    Minion.where(minion_id: nil).find_each
+    respond_to do |format|
+    	format.html { redirect_to root_path }
+    	format.json { head :ok }
+    end
+	end
+end
```
ereslibre: Indentation seems to be broken on the whole file.

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      minion.each do | key, value |
+        if value["tx_update_reboot_needed"]
+        	Minion.where(minion_id: key).update_all( highstate: Minion.highstates[:pending] )
+        end
+      end
+    end
+
+    Velum::Salt.update_orchestration
+
+    Minion.where(minion_id: nil).find_each
```
ereslibre: What's the purpose of this?

app/assets/javascripts/dashboard.js
```diff
@@ -97,10 +122,32 @@ MinionPoller = {
...
       '" value="' + minion.id + '" type="radio" disabled="" ' + checked + '>';
 
+    statusText = ''
+
+    if (tx_update_reboot_nodes && minion.highstate == "applied") {
+      statusHtml = '<i class="fa fa-arrow-circle-up text-info fa-2x" aria-hidden="true"></i>';
+      statusText = 'Update Available'
+    }
+
+    if (tx_update_reboot_nodes && minion.highstate == "failed") {
```
ereslibre: Do we want to take this as `Update Available` and silently hide the fact  that the highstate failed?

app/controllers/concerns/discovery.rb
```diff
@@ -1,18 +1,47 @@
...
 
   # Responds with either an HTML or JSON version of the available minions.
   def discovery
     @assigned_minions = Minion.assigned_role
     @unassigned_minions = Minion.unassigned_role
 
+    tx_update_reboot_needed = Hash.new
+    tx_update_failed = Hash.new
+
+    minions = perform_request(endpoint: "/minions", method: "get")
```
ereslibre: I am not sure about calling all the time here to the `salt-api`. This will the request coming from the user slower. We could add a cron job that updates the status of the minions on the background and we would only ask for this column on the database (since we are missing a resque-similar tool for background jobs, that's why a cron would be the best solution IMHO).

However this can be improved at a later time, is not that important.

app/controllers/concerns/discovery.rb
```diff
@@ -1,18 +1,47 @@
...
+
+    #logger.info JSON.parse(minions.body)["return"]
```
ereslibre: We can remove this I guess.

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+class SaltController < ApplicationController
+	include Velum::SaltApi
+ 	skip_before_action :redirect_to_setup
+
+	def update
+
+    minions = perform_request(endpoint: "/minions", method: "get")
+    ret = JSON.parse(minions.body)["return"]
+
+    for minion in ret do
```
ereslibre: We usually use the `ret.each do |minion|` syntax, but affter using `Salt.minions` you won't need to do this I think.

app/controllers/concerns/discovery.rb
ereslibre: This can be removed to use `Salt.minions` that is already implemented: https://github.com/kubic-project/velum/blob/master/lib/velum/salt.rb#L24

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
+require "velum/salt"
+require "velum/salt_api"
+# HealthController holds methods for checking the readiness of the application.
+class SaltController < ApplicationController
+	include Velum::SaltApi
+ 	skip_before_action :redirect_to_setup
+
+	def update
+
+    minions = perform_request(endpoint: "/minions", method: "get")
```
ereslibre: This can be removed using `Salt.minions` that is already implemented: https://github.com/kubic-project/velum/blob/master/lib/velum/salt.rb#L24

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    ret = JSON.parse(minions.body)["return"]
+
+    for minion in ret do
+      minion.each do | key, value |
+        if value["tx_update_reboot_needed"]
```
ereslibre: Could we rename the arguments? `minion.each do |minion_id, grains|`?

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        	Minion.where(minion_id: key).update_all( highstate: Minion.highstates[:pending] )
```
ereslibre: Would `Minion.where(minion_id: key).update_all(highstate: :pending)` work?

app/models/salt_handler/minion_highstate.rb
```diff
@@ -25,10 +25,11 @@ def process_event
 
     # After applying a highstate once, it is applied again and again.
     # Salt probably tries to make sure it stays applied. We don't need to process
-    # those events. We only need to process a highstate if one is pending.
+    # those events. We only need to process a highstate if one is pending, or has 
+    # failed in the past.
     minion = Minion.find_by(
       minion_id: minion_id,
-      highstate: Minion.highstates[:pending]
+      highstate: [Minion.highstates[:pending], Minion.highstates[:failed]]
```
ereslibre: I also think we can directly use `highstate: [:pending, :failed]` here.

app/views/dashboard/index.html.slim
```diff
@@ -7,15 +7,20 @@ h1 Cluster Status
         strong  new
         | nodes are available but have not been added to the cluster yet
 = link_to "Download kubectl config", kubectl_config_path, id: "download-kubeconfig", class: "btn btn-default btn-table pull-right", disabled: true
+= link_to update_path, id: "update-nodes", class: "btn btn-default btn-table pull-right", disabled: true, method: :post do
+    span.glyphicon.sidebar-icon.glyphicon-circle-arrow-up
+    span   Update Nodes
```
ereslibre: Indentation looks weird here.

app/controllers/concerns/discovery.rb
ereslibre: Also, renaming the arguments to `minion_id`, `grains` would be great, so it's something more intuitive.

app/controllers/salt_controller.rb
grahamhayes: done

app/controllers/salt_controller.rb
grahamhayes: Done

app/controllers/salt_controller.rb
grahamhayes: I am not sure - `Minion.highstates` is an `enum` - I have no idea how ruby deals with them

app/controllers/salt_controller.rb
grahamhayes: leftover code :)

app/assets/javascripts/dashboard.js
grahamhayes: I am not sure - I am open to suggestions on this

app/views/dashboard/index.html.slim
grahamhayes: The spans are in the button link - I *think* they should be indented?

app/assets/javascripts/dashboard.js
kiall: Maybe use "Update Failed (Retryable)" or something like that?

app/controllers/salt_controller.rb
kiall: Mixing tabs and spaces @grahamhayes? I'm pretty sure that against the law (programmer law..)

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,25 @@
+require "velum/salt"
+require "velum/salt_api"
+# HealthController holds methods for checking the readiness of the application.
```
kiall: Comment is out of date

grahamhayes:
updated-view

New screenshot

grahamhayes:
Tested in conjunction with kubic-project/salt#105 (needed for a reboot to actually happen)

-  Added a grain to a node (manually by editing `/etc/salt/grains`).
- Ran `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt '*' saltutil.refresh_grains`
- Opened Velum.
- Clicked "Update all Nodes"
- Watched node to see if it rebooted
- Checked velum to see if it showed no nodes needing an update

mssola:
As @ereslibre pointed out on IRC, after #173 we might have unassigned minions on the DB, so we have to make sure that updates are not being updated when you press that button.

app/models/salt_handler/minion_highstate.rb
ereslibre: This check can be removed, no? We only care about finding the `Minion` and setting it the last highstate result (so `Minion.find_by minion_id: minion_id` would be enough. (I know we want to move away from this, but I mean for 1.0 logic)

ereslibre: NVM, will merge https://github.com/kubic-project/velum/pull/189 after this and rebasing it... No need to do it here.

ereslibre:
Rubocop is complaining about a too long example. Also, it looks like that this PR alters how the table is being built, not matching contents with columns. This is current Velum `master`:

screenshot from 2017-06-13 15-50-25

As opposed to how it looks like with this PR:

screenshot from 2017-06-13 15-48-40

grahamhayes:
@ereslibre good catch, that was lost in rebase hell :) - Fixed
update

ereslibre:
@grahamhayes Is the latest screenshot the last one? Or did you upload another? I'd say it still looks broken :)

app/assets/javascripts/dashboard.js
```diff
@@ -156,10 +168,35 @@ MinionPoller = {
...
+          case "failed":
+            statusHtml = '<i class="fa fa-arrow-circle-up text-warning fa-2x" aria-hidden="true"></i>';
+            statusText = 'Update Failed - Retryable'
+            break;
+          case "pending":
+            statusText = 'Update in progress'
+            break;
+        }
+        break;
+      case 2:
```
vitoravelino: If `update_status` is 2 then it's not retryable?

app/views/dashboard/index.html.slim
```diff
@@ -56,15 +62,15 @@ h1 Cluster Status
...
+                | &nbsp;
+              th
                 | Status
               th
-                | Hostname
+                | ID
               th
-                | Role
+                | Hostname
               th.text-center
```
vitoravelino: You removed `Master` column here. Also, only master column needs `.text-center`, you can remove from this one.

app/assets/javascripts/dashboard.js
```diff
@@ -156,10 +168,35 @@ MinionPoller = {
     masterHtml = '<input name="roles[master][]" id="roles_master_' + minion.id +
       '" value="' + minion.id + '" type="radio" disabled="" ' + checked + '>';
 
+    statusText = ''
```
vitoravelino: Status text will be empty if no update? Isn't it a bit weird to have an empty "Status" column? I would guess since it's being used to show the text "update available", we could use for other states too.

app/views/dashboard/index.html.slim
grahamhayes: In the mockups "Role" and "master" were combined

app/assets/javascripts/dashboard.js
grahamhayes: Thats what was shown in the mockups - so I went with what was laid out there.

app/assets/javascripts/dashboard.js
grahamhayes: No - it means the `transactional-update` has failed - which needs manual intervention

vitoravelino: Oh. Does this apply to the admin node too?

app/views/dashboard/index.html.slim
vitoravelino: You are right... Do we need the radio input then? The role is already there.

grahamhayes: We can take it out - but lets do that in a separate commit

app/assets/javascripts/dashboard.js
grahamhayes: Yes

app/views/dashboard/index.html.slim
```diff
@@ -45,6 +49,8 @@ h1 Cluster Status
       = link_to kubectl_config_path, id: "download-kubeconfig", class: "btn btn-xs btn-default pull-right", disabled: true do
         i.fa.fa-download.fa-fw
         | kubectl config
+      a.btn.btn-default.btn-xs.update-admin-btn.pull-right.hidden data-toggle="modal" data-target=".update-admin-modal"
```
vitoravelino: You can remove this. Probably kept during rebases.

app/assets/javascripts/dashboard.js
vitoravelino: Ouch! Then the UI related to the admin node notification needs to be updated. Sorry. :(

Can I fix that in this branch? You can squash the commit before merging.

app/views/dashboard/index.html.slim
grahamhayes: Done

app/assets/javascripts/dashboard.js
vitoravelino: Actually, let's confirm with @mssola. He was the one who did the `admin_status`.

grahamhayes:
Travis looks random fails - only 2 of the ruby versions are broken - everything else is good to go

mssola:
Related to my third point: 

lala

mssola:
For my first point, I think that it's done because we prioritize failed updates and needed updates to the top right ? If so, ignore my first point 😉

app/assets/javascripts/dashboard.js
mssola: @vitoravelino this is a separate issue to be fixed in another PR I think. The admin node will have the same range of values. If an update has failed, imho we should be able to click the reset button. Moreover, I'd put a message somewhere telling the user that the last update of the admin node failed.

mssola: That being said, the message should also instruct the administrator that the failure has to be manually handled. I think that we should get back to either @flavio or @eotchi on this, but as I said, I think this should be addressed in a separate PR.

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,31 @@
+require "velum/salt"
+# HealthController holds methods for checking the readiness of the application.
+class SaltController < ApplicationController
+  skip_before_action :redirect_to_setup
+
+  def update
+    minions = assigned_with_status
+    minions.each do |minion|
+      if minion.update_status == Minion.statuses[:update_needed]
```
mssola: As a general rule, avoid fat controller methods. In this case, I'd move this logic into an instance method of the `Minion` model. This method would be called in the `assigned_with_status` method when we loop through all the minions (we don't want to loop twice if we can merge both operations).

grahamhayes:
@mssola 

> The order of the rows changes when a node has been updated. It's a minor thing, but it's still weird and unexpected that this happens.

That came in from another patchset - I am not sure how useful it is, but if we want to remove it we can make that change in a separate PR.

> Failed updates are not considered

They are - there are two types of failures.

1. `transactional-update` failed to apply. This requires manual intervention, or waiting for another `transactional-update` run to happen. We cannot run an update on a node if `tx_update_failed: True` is set.
2. The highstate failed to apply. This is possibly retryible - and the "Update" button will show in this case.

> Why should we have the text "Update Failed" if we already have the icon ?

I was working off of the mockups provided - they had this text, so I included it.

> Plus, the "Role" column is not aligned correctly.

Yup - my bad, will update that

app/controllers/salt_controller.rb
grahamhayes: So move the `Minion.find_by(minion_id: minion.minion_id).update(highstate: Minion.highstates[:pending])` line to a method on the model ? Or the whole `assigned_with_status` as well?

grahamhayes: Updated in patchset

app/models/minion.rb
```diff
@@ -101,6 +101,16 @@ def assign_role(new_role)
   end
   # rubocop:enable SkipsModelValidations
 
+  def self.mark_pending_update
```
mssola: Could you add documentation ? 😇

grahamhayes: Done

app/controllers/salt_controller.rb
```diff
@@ -0,0 +1,15 @@
+require "velum/salt"
+# HealthController holds methods for checking the readiness of the application.
```
mssola: Leftover ?

grahamhayes: Yup - fixed

grahamhayes:
e2e tests ran and passed locally.

grahamhayes:
@ereslibre and @mssola had approved - I needed to retrigger CI, so that caused the +As to go away.

mssola:
@grahamhayes can this branch be deleted, or are you using it for something else ?

PR #182, opened 2017-06-07
Improve cluster configuration page [bsc#1040367]

flavio:
I know the UX of the cluster configuration page is still ongoing, but I think we can make a significant improvement with a few lines of code.

This is **not meant** to be the final look of the page, it's just something that we can can get out quickly before 1.0.

This is just something that I hacked together in between today's meeting thanks to my mad UI skills (yes, I'm kidding!)

This also helps with [bsc#1040367](https://bugzilla.suse.com/show_bug.cgi?id=1040367).

## Before

full-page-before

## After

The different options have been grouped together based on their "topic":

full-page-after

### Contextual help

For cryptic settings we could provide some contextual help, you can see that in action here:

tooltip

# Open questions

  * [x] proxy configuration: I took a look at the salt code and these settings are applied only to the docker engine. The proxy settings are not going to be extended to the node too (the autoyast profile is not changed -> zypper on the new nodes won't use it). Am I right?
   * [x] no-proxy: this is a white space separated list. Am I right?
   * [x] External Kubernetes API server FQDN: this is a tough one to document. I tried my best, but I'm open to feedback. Maybe @kiall has something better :)
   * [x] Generic settings: I couldn't come with a better name for this group of options
   * [x] openQA tests: I guess these changes are going to break them. Isn't it?

### DIFF
--- a/app/views/setup/welcome.html.slim
+++ b/app/views/setup/welcome.html.slim
@@ -1,47 +1,68 @@
 = content_for :body_class, "welcome"
-.alert.alert-info.alert-dismissible role="alert"
-  button.close[type="button" data-dismiss="alert" aria-label="Close"]
-    span[aria-hidden="true"]
-      | &times;
-  i.fa.fa-4x.pull-left aria-hidden="true"
-  span In order to secure transactions and communication between your cluster nodes, it is required to generate certificates. Please enter the details for your certificate authority (CA).
 h1 Initial CaaSP Configuration
 = form_for :settings, url: setup_path, method: :put do |f|
-  .form-group
-    = f.label :dashboard, "Dashboard location"
-    = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control", required: true
-  .form-group
-    = f.label :apiserver, "External Kubernetes API server FQDN"
-    = f.text_field :apiserver, value: Pillar.value(pillar: :apiserver), class: "form-control", required: true
-  .form-group
-    = f.label :http_proxy, "HTTP proxy for Docker"
-    = f.url_field :http_proxy, value: Pillar.value(pillar: :http_proxy), class: "form-control"
-  .form-group
-    = f.label :https_proxy, "HTTPS proxy for Docker"
-    = f.url_field :https_proxy, value: Pillar.value(pillar: :https_proxy), class: "form-control"
-  .form-group
-    = f.label :no, "No-proxy settings for Docker"
-    = f.text_field :no_proxy, value: Pillar.value(pillar: :no_proxy), class: "form-control"
-  .form-group
-    = f.label :company_name, "Company Name"
-    = f.text_field :company_name, value: Pillar.value(pillar: :company_name), class: "form-control", required: true
-  .form-group
-    = f.label :company_unit, "Company Unit"
-    = f.text_field :company_unit, value: Pillar.value(pillar: :company_unit), class: "form-control", required: true
-  .form-group
-    = f.label :email, "E-Mail"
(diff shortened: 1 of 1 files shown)


flavio:
ping @eotchi, @vitoravelino, @ereslibre, @mssola.

Unfortunately I cannot remember Cierra's handle...

vitoravelino:
Looks great, @flavio! :clap:

ereslibre:
> This is just something that I hacked together in between today's meeting thanks to my mad UI skills (yes, I'm kidding!)

:rofl: 

> What do you think?

I love it. Yeah, I know this is not meant to be the *final* version, but is a huge improvement for what we have now.

A clear +1 from me.

flavio:
I updated the code. This is how the page will look like:

screenshot-localhost 3000 2017-06-08 08-32-40

Open questions:

  * proxy configuration: I took a look at the salt code and these settings are applied only to the docker engine. The proxy settings are not going to be extended to the node too (the autoyast profile is not changed -> `zypper` on the new nodes won't use it). Am I right?
  * `no-proxy`: this is a white space separated list. Am I right?
  * `External Kubernetes API server FQDN`: this is a tough one to document. I tried my best, but I'm open to feedback. Maybe @kiall has something better :)
  * `Generic settings`: I couldn't come with a better name for this group of options

app/views/setup/welcome.html.slim
```diff
@@ -1,47 +1,67 @@
...
+          = f.text_field :dashboard, value: (Pillar.value(pillar: :dashboard) || request.host), class: "form-control", required: true
+          span class="input-group-addon"
+            a data-toggle="tooltip" data-placement="left" title="Hostname or IP of the node running this web interface"
+              i class='glyphicon glyphicon-info-sign'
+      .form-group
+        = f.label :apiserver, "External Kubernetes API server FQDN"
+        .input-group
+          = f.text_field :apiserver, value: Pillar.value(pillar: :apiserver), class: "form-control", required: true
+          span class="input-group-addon"
+            a data-toggle="tooltip" data-placement="left" title="FQDN pointing to the node acting as Kubernetes master"
```
flavio: @kiall: this is the "help" message for the Kube external FQDN

eotchi: What about some additional context (assuming it fits decently), on the lines of:

` Fully qualified domain name used to reach the cluster from the outside. In a simple, single master deployment this will be the FQDN of the node you are about to select as master. For advanced options refer to the documentation.`

kravciak:
I think it would be useful to let user know that proxy settings are optional

app/views/setup/welcome.html.slim
flavio: Thanks, I'm going to use this one.

kiall: +1

flavio:
> I think it would be useful to let user know that proxy settings are optional

Done

screenshot-localhost 3000 2017-06-08 14-08-51

kiall:
Oh - "proxy configuration: I took a look at the salt code and these settings are applied only to the docker engine. The proxy settings are not going to be extended to the node too (the autoyast profile is not changed -> zypper on the new nodes won't use it). Am I right?"

Ehh, that's likely a problem. It's likely we will need to extend the proxy settings to zypper too?

flavio:
> Ehh, that's likely a problem. It's likely we will need to extend the proxy settings to zypper too?

Yes, I was going to file a dedicated bug for that. Luckily the fix should be easy: we just have to extend the autoyast file we are serving with these information

flavio:
I created [bsc#1043538](https://bugzilla.suse.com/show_bug.cgi?id=1043538) to track adding http proxy to the entire node at provisioning time.

PR #183, opened 2017-06-07
WIP: bootstrap: enforce at least 3 nodes for cluster

vitoravelino:
The etcd configuration used in CaaSP 1.0 *requires*
three total nodes to be present in the cluster
(can be 1 master and two minions) - without this,
etcd does not elect a leader, and bootstrap will fail.

See bsc#1040477

**ATENTION**: To avoid extra work and only because this bugfix was really simple, I've checked out this branch based on another branch (#173) instead of master. So, only worry to review this after merging https://github.com/kubic-project/velum/pull/173.

The only changes were on `setup_controller.rb` and `setup_controller_spec.rb` (e1a6f5b60f6).

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -67,7 +67,9 @@ class SetupController < ApplicationController
   end
 
   def update_nodes_params
-    params.require(:roles)
+    roles_params = params.require(:roles)
+    roles_params[:worker] = (roles_params[:worker] || []) - roles_params[:master]
+    roles_params
   end
 
   def failed_assigned_nodes(assigned)
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -7,6 +7,9 @@ require "velum/salt"
 # welcoming, setting certain general settings, master selection, discovery and
 # bootstrapping
 class SetupController < ApplicationController
+  MINIMUM_MASTER_NODES = 1
+  MINIMUM_WORKER_NODES = 2
+
   include Discovery
 
   rescue_from Minion::NonExistingNode, with: :node_not_found
@@ -15,6 +18,7 @@ class SetupController < ApplicationController
   before_action :redirect_to_dashboard
   before_action :check_empty_settings, only: :configure
   before_action :check_empty_bootstrap, only: :bootstrap
+  before_action :check_minimum_nodes_bootstrap, only: :bootstrap
 
   def configure
     res = Pillar.apply(settings_params)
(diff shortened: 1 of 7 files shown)


app/controllers/setup_controller.rb
```diff
@@ -105,4 +111,17 @@ def check_empty_bootstrap
       format.json { render json: msg, status: :unprocessable_entity }
     end
   end
+
+  def check_minimum_nodes_bootstrap
```
mssola: Add documentation please.

app/controllers/setup_controller.rb
```diff
@@ -105,4 +111,17 @@ def check_empty_bootstrap
...
+    return if roles_params[:master].length >= MINIMUM_MASTER_NODES &&
+      roles_params[:worker].length >= MINIMUM_WORKER_NODES
+
+    respond_to do |format|
+      msg = "Please select a master and at least two workers"
+      format.html do
```
mssola: This can be a one-liner like the `format.json` below no ?

mssola:
Rubocop is complaining 😀

vitoravelino:
Done! :)

vitoravelino:
Added WIP again because I messed up with the commits when rebasing the other branch and I'll fix this when the other branch is merged. One thing at a time. :)

ereslibre:
I am not so sure about this, I don't think we can enforce 3 minimum nodes at least. AFAIK etcd should elect a leader even with 2 nodes.

@flavio, @kiall, thoughts on this?

kiall:
I think it's 100% valid to enforce 3 nodes - 1 or 2 nodes is basically a broken cluster since quorum is needed :)

flavio:
Wait a second, a cluster with 2 nodes is still possible. There's going to be just a etcd master so the quorum will be reached. The cluster won't be production ready at all, but it's good for people doing a quick evaluation of CaaSP (see [this](https://bugzilla.suse.com/show_bug.cgi?id=1040477#c6) and [this](https://bugzilla.suse.com/show_bug.cgi?id=1040477#c7) comments).

I would say:

  * we allow the user to bootstrap a cluster made by 2 nodes:
    * The deployment button is clickable
    * A show a big fat warning is shown when clicking the deploy button. A confirmation dialog is shown, this one explains how dangerous the final operation is (@eotchi and I  can provide a text later for that).
    * A warning message is shown all the time until the cluster has 3 nodes assigned to it. The idea is: keep warning about that until the user adds a new node to the cluster.
  * 3+ nodes: no warning at all, just go ahead

What do the others think about that?

ereslibre:
> What do the others think about that?

I completely agree, and those were mi thoughts the other day. Is a really weak cluster, but quorum will be possible, and etcd should work just fine.

I am OK showing a warning dialog, but letting the cluster be created so it's possible to create such a cluster for testing reasons, for example.

vitoravelino:
It wasn't clear for me if we should display a warning or avoid the creation of the cluster. I took the risk and got the wrong idea. No problem. I guess we can close this and I'll work on the other solution. No worries.

PR #181, opened 2017-06-07
WIP: Initial work on the update of the dashboard

mssola:
Updating comment so Github picks the "to do" list properly:

- [x] Detect needed updates
- [x] Detect failed updates
- [x] Polish the `TODO` note
- [ ] Add endpoint for updating node and reboot (through Salt API)
- [ ] Tests (~~detecting updates~~, update & reboot)
- [X] Implement UI layout changes
- [X] Manage notification if update is available
- [X] Feature tests if update is available
- [ ] Handle "Reboot to update" user request 
- [ ] Reboot to update feature test

PR #185, opened 2017-06-09
Proper update detection and reboot of the admin node

mssola:
- [x] Initial code
- [x] UI: I've taken the same code as in PR #181. @vitoravelino: let's scratch what you did (since it's going to #180), and let's simply add the update & reboot button (dialog) when the `admin` value contains an update (the backend does the same as in #181).
- [x] Update tests
- [x] More tests (coverage < 100%)

### DIFF
--- a/app/models/salt_handler/minion_start.rb
+++ b/app/models/salt_handler/minion_start.rb
@@ -20,12 +20,14 @@ class SaltHandler::MinionStart
   def process_event
     parsed_data = salt_event.parsed_data
 
-    # Ignore the ca minion. It shouldn't be used as part of the k8s cluster.
-    return false if parsed_data["id"] == "ca"
+    # Ignore the ca and the dashboard minions. It shouldn't be used as part of
+    # the k8s cluster.
+    id = parsed_data["id"]
+    return false if id == "ca" || id == "admin"
 
-    minion_info = Velum::Salt.minions[parsed_data["id"]]
+    minion_info = Velum::Salt.minions[id]
 
     # false if a minion with this minion_id or fqdn already exists (uniqueness validation)
-    Minion.new(minion_id: parsed_data["id"], fqdn: minion_info["fqdn"]).save
+    Minion.new(minion_id: id, fqdn: minion_info["fqdn"]).save
   end
 end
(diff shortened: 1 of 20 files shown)


mssola:
Do not review yet 😇

app/views/dashboard/_update_admin_modal.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+        <h4 class="modal-title" id="modal-label">
+          The admin node needs to reboot in order to apply the software update
+        </h4>
+      </div>
+      <div class="modal-body">
+       <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Tempore maiores repellat quia architecto, sunt commodi earum eos, sint dolor quaerat voluptatum! Earum obcaecati distinctio, at explicabo vel similique minus nam.</p>
```
vitoravelino: @flavio I guess you could help with the detailed message of admin node reboot?

mssola:
> My biggest concerns is targeting all minions with '*' when we might have unassigned minions (but still registered against the salt master)

As explained on IRC, this is something we don't have to fear for this PR, since we are explicitly only tackling assigned minions. That being said, this will surely be a problem in #180, so I'll also add a comment there.

ereslibre:
https://github.com/kubic-project/salt/blob/master/salt/orch/kubernetes.sls#L3 will need to be updated, so we also disable `rebootmgr` on the admin node.

app/models/salt_handler/minion_start.rb
```diff
@@ -20,12 +20,14 @@ def initialize(salt_event)
   def process_event
     parsed_data = salt_event.parsed_data
 
-    # Ignore the ca minion. It shouldn't be used as part of the k8s cluster.
-    return false if parsed_data["id"] == "ca"
+    # Ignore the ca and the dashboard minions. It shouldn't be used as part of
+    # the k8s cluster.
+    id = parsed_data["id"]
+    return false if id == "ca" || id == "admin"
```
ereslibre: This could be simplified for future values as `return false if ["ca", "admin"].include? id`

mssola:
> https://github.com/kubic-project/salt/blob/master/salt/orch/kubernetes.sls#L3 will need to be updated, so we also disable rebootmgr on the admin node.

Good catch!

PR #188, opened 2017-06-12
registration: allow only one admin user

vrothberg:
The administration dashboard allows to create as many admin users as
desired, which is a potential security threat as Velum does not have any
user management. Besides, an admin cannot see how many other users
exist.

Avoid creating more than one admin user by not showing the
`Create an account` button when a user already exists. `users/sign_up`
will further redirect to the `root_path` and print an alert.

Notice that this is only an intermediate solution until multiuser
support and RBAC is implemented. Until then, a password reset must be
done by connecting to the machine (manually or via a script).

See bsc#1040831

Signed-off-by: Valentin Rothberg <vrothberg@suse.com>

See the screenshot below. The button is not displayed anymore and an 
alert is printed when forcefully visiting the registration path.
screenshot from 2017-06-12 12-56-49

### DIFF
--- a/app/controllers/auth/registrations_controller.rb
+++ b/app/controllers/auth/registrations_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 # Auth::RegistrationsController manages user signup/removal.
 class Auth::RegistrationsController < Devise::RegistrationsController
+  before_action :restrict_admin_users
+
   layout "authentication", except: :edit
 
   skip_before_action :redirect_to_setup
@@ -9,5 +11,12 @@ class Auth::RegistrationsController < Devise::RegistrationsController
   def new
     @have_users = User.any?
     super
+  end
+
+  # Restrict the amount of admin users to 1, see bsc#1040831
+  def restrict_admin_users
+    return if User.count.zero?
+    flash[:alert] = "Admin user already exists."
+    redirect_to root_path
   end
 end
--- a/app/views/devise/sessions/new.html.slim
+++ b/app/views/devise/sessions/new.html.slim
@@ -8,9 +8,10 @@
       p SUSE Container as a Service Platform allows you to provision, manage, and scale container-based applications.
       p It automates your tedious management tasks allowing you to focus on development and writing apps to meet business goals.
     p
-      | Don't have an account? 
-      a.btn.btn-default[href="/users/sign_up"]
-        | Create an account
+      - if User.count.zero?
+        | Don't have an account?
+        a.btn.btn-default[href="/users/sign_up"]
+          | Create an account
   .col-sm-4.col-sm-offset-1
     h2.Raleway-font
       | Log In
--- a/spec/features/auth/signup_feature_spec.rb
+++ b/spec/features/auth/signup_feature_spec.rb
@@ -2,51 +2,51 @@
 require "rails_helper"
 
 feature "Signup feature" do
-  before do
-    visit new_user_registration_path
-  end
-
   let(:user) { build(:user) }
 
-  scenario "I am able to signup from login page" do
+  # rubocop:disable RSpec/ExampleLength
+  # rubocop:disable RSpec/MultipleExpectations
+
+  # XXX: the following tests depend on another and thus cannot be split into
+  # multiple scenarios. This must be fixed as soon as multiuser support is
+  # implemented.
+
+  scenario "Account creation tests", js: true do
+    # account creation reachable
     visit new_user_session_path
     click_link("Create an account")
     expect(current_path).to eq new_user_registration_path
-  end
 
-  scenario "I am able to go back to the login page" do
-    user.save
-    visit new_user_registration_path
+    # wrong email format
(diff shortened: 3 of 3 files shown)


app/controllers/auth/registrations_controller.rb
```diff
@@ -1,13 +1,22 @@
...
 class Auth::RegistrationsController < Devise::RegistrationsController
+  before_action :restrict_admin
+
   layout "authentication", except: :edit
 
   skip_before_action :redirect_to_setup
 
   # Re-implemented so the template has some auxiliary instance variables.
   def new
+    binding.pry
```
MaximilianMeister: you forgot to remove the debugger ;-)

vrothberg: Thanks for catching this!

app/controllers/auth/registrations_controller.rb
```diff
@@ -1,13 +1,22 @@
...
     @have_users = User.any?
     super
   end
+
+  def restrict_admin
+    return if User.count == 0
+    flash[:alert] = "Admin user already exists."
+    redirect_to root_path
+  end
```
ereslibre: We could use a regular flow here since complexity is very low, why not:

```
if User.count > 0
  flash[:alert] = "Admin user already exists."
  redirect_to root_path
end
```

vitoravelino: :+1:

vitoravelino:
Maybe we could go further and hide the "Create admin account" button if user already exists? Just wondering...

app/controllers/auth/registrations_controller.rb
vrothberg: Thanks for reviewing. I can change that, no problem.

vrothberg:
@vitoravelino Yes, that's what I just wanted to do now. If not, we just confuse users.

vitoravelino:
> @vitoravelino Yes, that's what I just wanted to do now. If not, we just confuse users.

So, all you need to do is move the condition to a helper method and use it on the view.

vrothberg:
> We might also want to add specific tests for this, so coverage don't drop.

Those tests also have to be moved into one scenario as the the `Create an account` button will not be displayed as soon as a user has been created. I will update the commit soon.

app/controllers/auth/registrations_controller.rb
mssola: rubocop will complain if you do this change 😄

vrothberg:
Now addressing the rubocop issues...

vrothberg:
I restarted the builds several times and the builds are failing randomly. Sometimes all fail, some random builds fail on random Ruby version. I opened an issue as this must be fixed to have a reliable CI, see #192.

Reviews are welcome.

app/controllers/auth/registrations_controller.rb
```diff
@@ -10,4 +12,10 @@ def new
     @have_users = User.any?
     super
   end
+
+  def restrict_admin
```
mssola: Could you add documentation? You don't have to be too verbose, just an explanation of what it does.

vrothberg: Done [x]

vrothberg:
Any further comments / objections or shall we merge?

app/views/devise/sessions/new.html.slim
```diff
@@ -8,9 +8,10 @@
       p SUSE Container as a Service Platform allows you to provision, manage, and scale container-based applications.
       p It automates your tedious management tasks allowing you to focus on development and writing apps to meet business goals.
     p
-      | Don't have an account? 
-      a.btn.btn-default[href="/users/sign_up"]
-        | Create an account
+      - if User.count == 0
```
ereslibre: Just for the sake of consistency, could this check be `User.count.zero?`?

app/controllers/auth/registrations_controller.rb
```diff
@@ -10,4 +12,11 @@ def new
     @have_users = User.any?
     super
   end
+
+  # Restrict the amount of admin users to 1, see bsc#1040831
+  def restrict_admin_users
+    return false if User.count.zero?
+    flash[:alert] = "Admin user already exists."
+    redirect_to root_path
```
ereslibre: Could we use a regular `if` block here?

```ruby
if User.count.zero?
  flash[:alert] = "Admin user already exists."
  redirect_to root_path
end
```

vrothberg: Sure. I'll fix address both comments and update PR soon. Thanks!

vrothberg: app/controllers/auth/registrations_controller.rb:18:5: C: Use a guard clause instead of wrapping the code inside a conditional expression.
    unless User.count.zero?

:-(

ereslibre: Thank you for checking @vrothberg, gosh :man_facepalming:

vrothberg:
\o/ What a historic moment, @ereslibre. My first code goes into velum :-)

PR #189, opened 2017-06-12
When setting the highstate resolution, consider all highstates

ereslibre:
With the current logic, it could happen that during the same
orchestration run the first highstate succeeds (setting the minion
status to :applied as succeeded), while the second highstate fails
(and the :applied status as succeeded will be kept).

With this change we consider the latest highstate run is the most
recent, and whatever it reports is what the final status of the
minion will be.

### DIFF
--- a/app/models/salt_handler/minion_highstate.rb
+++ b/app/models/salt_handler/minion_highstate.rb
@@ -23,14 +23,7 @@ class SaltHandler::MinionHighstate
     minion_id = salt_event.tag.match(TAG_MATCHER)[1]
     return false unless minion_id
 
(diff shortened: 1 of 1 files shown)


ereslibre:
It is a green run with `rubocop` and `rspec` locally. Seems that Travis failures are artifacts.

ereslibre:
Let's wait until https://github.com/kubic-project/velum/pull/180 is merged to merge this one. I will check that everything is fine again after rebasing it on top of https://github.com/kubic-project/velum/pull/180.

ereslibre:
Tests are passing, @grahamhayes gave his +1 and @mssola did too. Merging this one.

PR #190, opened 2017-06-12
Check that result isn't `nil` before checking result contents

ereslibre:
We are checking that the referenced file exists by performing a
rudimentary check. However, the result might be completely empty
under certain circumstances, making our check fail (`include?` called
on `nil` will fail).

This fixes the 500 error, but we still need to find out why the salt
side of things is failing at all to retrieve the contents of the
file.

Fixes: bsc#1043843

### DIFF
--- a/lib/velum/salt.rb
+++ b/lib/velum/salt.rb
@@ -46,7 +46,7 @@ module Velum
         val = el.values.first
 
         # TODO: improve error handling...
-        val.include?("No such file or directory") ? nil : val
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
[Bug 1043843 - Downloading kubeconfig fails during master node reboot](https://trello.com/c/vUtgJeJz/131-bug-1043843-downloading-kubeconfig-fails-during-master-node-reboot)

PR #193, opened 2017-06-13
add a model for certificates and registries

MaximilianMeister:
fixes bsc#1039876

https://bugzilla.suse.com/show_bug.cgi?id=1039876

## Depends on
https://github.com/kubic-project/salt/pull/313
https://github.com/kubic-project/salt/pull/129

registryv2


MaximilianMeister:
[Bug 1039876 - docker cannot pull images from insecure registry](https://trello.com/c/PyAkd0il/57-bug-1039876-docker-cannot-pull-images-from-insecure-registry)

ereslibre:
I know this is still not decided, but in case of going in this direction, would it make sense to accept a list instead of a single insecure registry? We could accept several comma-separated.

mssola:
> I know this is still not decided, but in case of going in this direction, would it make sense to accept a list instead of a single insecure registry? We could accept several comma-separated.

👍

app/models/pillar.rb
```diff
@@ -65,7 +66,7 @@ def apply(pillars)
 
       Pillar.all_pillars.each do |key, pillar_key|
         # The following pillar keys can be blank, so ignore them if they are.
-        next if [:http_proxy, :https_proxy, :no_proxy].include?(key) && pillars[key].blank?
+        next if [:http_proxy, :https_proxy, :no_proxy, :insecure_registry].include?(key) && pillars[key].blank?
```
mssola: If this list is getting larger, maybe it makes sense to have it in a constant or something like that on the top of the class. Or maybe on another method.

MaximilianMeister:
> I know this is still not decided, but in case of going in this direction, would it make sense to accept a list instead of a single insecure registry? We could accept several comma-separated.

yes that would make sense indeed, it can be adapted once we have some feedback

app/models/pillar.rb
MaximilianMeister: ok

app/views/setup/welcome.html.slim
```diff
@@ -18,6 +18,13 @@ h1 Initial CaaSP Configuration
           span class="input-group-addon"
             a data-toggle="tooltip" data-placement="left" title="Fully qualified domain name used to reach the cluster from the outside. In a simple, single master deployment this will be the FQDN of the node you are about to select as master. For advanced options refer to the documentation."
               i class='glyphicon glyphicon-info-sign'
+      .form-group
+        = f.label :insecure_registry, "External insecure Docker registry (optional)"
+        .input-group
+          = f.text_field :insecure_registry, value: Pillar.value(pillar: :insecure_registry), class: "form-control", required: true
+          span class="input-group-addon"
+            a data-toggle="tooltip" data-placement="left" title="Local insecure registries are highly discouraged and should be avoided. However in some setups it might be necessary. You can add an insecure registry here."
```
vrothberg: There's a comma "," missing after "[...]. However".

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,32 @@ $(function() {
       $(this).text('Show');
     }
   });
-});
\ No newline at end of file
+
+  $(document).on('click', '.js-toggle-registry-settings-btn', function() {
```
vitoravelino: You can remove this and add `.js-toggle-overlay-settings-btn` above separating it with a comma.

app/controllers/setup_controller.rb
```diff
@@ -31,13 +31,35 @@ def welcome
     @services_cidr = Pillar.value(pillar: :services_cidr) || "172.24.0.0/16"
     @api_cluster_ip = Pillar.value(pillar: :api_cluster_ip) || "172.24.0.1"
     @dns_cluster_ip = Pillar.value(pillar: :dns_cluster_ip) || "172.24.0.2"
+    @insecure_registries = DockerRegistry.all || []
   end
   # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
   def configure
+    settings_params["insecure_registries"].each do |registry|
+      next if !registry["docker_registry_name"] || registry["docker_registry_name"].empty?
```
vitoravelino: Use `blank?` instead?

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,32 @@ $(function() {
...
+
+    if (collapsed) {
+      $(this).text('Hide');
+    } else {
+      $(this).text('Show');
+    }
+  });
+
+  $(document).on('click', '.btn-add-registry', function() {
+    var registryForm = $('.registry-form'),
```
vitoravelino: Prefix variable with $ to indicate that it's a jQuery object.

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,32 @@ $(function() {
...
+        currentEntry = $(registryForm).children('.registry-entry:first'),
```
vitoravelino: You can simply `$registryForm.find`

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,32 @@ $(function() {
...
+        newEntry = $(currentEntry.clone()).appendTo(registryForm);
+
+    newEntry.find('#settings_insecure_registries__docker_registry_name').val('').prop('disabled', false);
+    newEntry.find('#settings_insecure_registries__docker_registry_mirror').prop('checked', false);
+    newEntry.find('#settings_insecure_registries__docker_registry_certificate').val('');
+    registryForm.find('.registry-entry:not(:last) .btn-add-registry')
+        .removeClass('btn-add-registry').addClass('btn-remove-registry')
+        .removeClass('btn-success').addClass('btn-danger')
+        .html('<span class="glyphicon glyphicon-minus"></span>');
+  }).on('click', '.btn-remove-registry', function(e) {
```
vitoravelino: Write this as another block, makes it easier to read.

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,32 @@ $(function() {
...
+    console.log($(this).parents('.row .registry-entry:first'));
+    $(this).parents('.row .registry-entry:first').remove();
```
vitoravelino: Use '.closest' when looking for a parent node.

vitoravelino: You can probably get rid of the `:first` pseudoclass too.

app/views/setup/_registry.html.slim
```diff
@@ -0,0 +1,15 @@
+= f.fields_for "insecure_registries[]", [] do |field|
+  .row.registry-entry
+    .form-group.col-sm-3
+      = field.text_field :docker_registry_name, placeholder: "Registry name", type: "text", value: (local_assigns[:registry] ? registry.name : "" ), disabled: !local_assigns[:registry].nil?
+    .form-group.col-sm-7
+      = field.text_area :docker_registry_certificate, placeholder: "Certificate", type: "text", value: (local_assigns[:registry] ? Certificate::DockerRegistry.find(registry.certificate_docker_registry_id).certificate : ""), cols: 80, rows: 20
```
vitoravelino: Can we put `Certificate::DockerRegistry.find(registry.certificate_docker_registry_id).certificate` in a variable/helper?

app/views/setup/_registry.html.slim
```diff
@@ -0,0 +1,15 @@
+= f.fields_for "insecure_registries[]", [] do |field|
+  .row.registry-entry
+    .form-group.col-sm-3
+      = field.text_field :docker_registry_name, placeholder: "Registry name", type: "text", value: (local_assigns[:registry] ? registry.name : "" ), disabled: !local_assigns[:registry].nil?
```
vitoravelino: Could we pass `registry: DockerRegistry.new` instead of nothing to this partial when there's no registries to avoid this ternary conditions everywhere? Just wondering...

app/views/setup/welcome.html.slim
```diff
@@ -25,6 +25,23 @@ h1 Initial CaaS Platform Configuration
...
+    #registry-settings-panel.panel-collapse.collapse
+      .panel-body
+        p Local insecure registries are highly discouraged and should be avoided. However, in some setups it might be necessary. You can add an insecure registry here.
+        .form-group.registry-form
+          = f.label :insecure_registries, "List of external insecure Docker registries (optional)"
+          - if @insecure_registries.any?
+            - @insecure_registries.each do |registry|
+              = render "setup/registry", f: f, registry: registry
+          - else
+            = render "setup/registry", f: f
```
vitoravelino: Could we pass registry: DockerRegistry.new instead of nothing to this partial when there's no registries to avoid the ternary conditions inside the partial? Just wondering...

app/assets/javascripts/setup.js
MaximilianMeister: this didn't work, i also tried `$registryForm.find()` but it doesnt work anymore then

app/assets/javascripts/setup.js
MaximilianMeister: removing the pseudoclass works, but closest didnt, i wrote it like this:

```js
$(this).closest().remove();
```

vitoravelino: Try `.closest('.registry-entry')`

app/assets/javascripts/setup.js
vitoravelino: You can leave the `children`. Just saw that it's more performatic than `find` since it only goes one level down.

app/assets/javascripts/setup.js
```diff
@@ -17,4 +17,22 @@ $(function() {
       $(this).text('Show');
     }
   });
-});
\ No newline at end of file
+
+  $(document).on('click', '.btn-add-registry', function() {
+    var $registryForm = $('.registry-form');
+    var $currentEntry = $registryForm.children('.registry-entry:first');
+    var $newEntry = $($currentEntry.clone()).appendTo($registryForm);
```
vitoravelino: Does the `clone` requires to rewrap it with `$`?

app/models/certificate/docker_registry.rb
```diff
@@ -0,0 +1,2 @@
+class Certificate::DockerRegistry < ActiveRecord::Base
+end
```
ereslibre: IMO it would be better if we only had `Certificate` model with polymorphic associations (http://guides.rubyonrails.org/association_basics.html#polymorphic-associations), so we could reuse certificates among different future services (only registries for now).

If we add other models that will require certificates, the customer will have to duplicate certificates for each one if we don't use the polymorphic association.

MaximilianMeister: done

vitoravelino:
Updated UI

screenshot 2017-11-28 14 44 55

app/models/docker_registry.rb
```diff
@@ -0,0 +1,42 @@
+class DockerRegistry < ActiveRecord::Base
+  belongs_to :certificate, polymorphic: true
+
+  validates :name, presence: true
+
+  class << self
+    def apply(settings_params)
+      settings_params["insecure_registries"].each do |registry|
+        next if registry["docker_registry_name"].blank?
```
vitoravelino: `registry["docker_registry_name"]` in a variable

app/models/docker_registry.rb
```diff
@@ -0,0 +1,42 @@
...
+        certificate = unless registry["docker_registry_certificate"].blank?
```
vitoravelino: Same for `registry["docker_registry_certificate"]`

app/models/docker_registry.rb
```diff
@@ -0,0 +1,42 @@
...
+          ).tap do |c|
+            c.certificate = registry["docker_registry_certificate"]
+            c.save
+          end
+        end
+        self.find_or_initialize_by(
+          name: registry["docker_registry_name"]
+        ).tap do |r|
+          r.name = registry["docker_registry_name"]
+          r.certifiable_id = certificate.blank? ? nil : certificate.id
```
vitoravelino: I would use `nil?` instead of `blank?` since we are expecting `nil` or an object from the `unless` statement.

app/views/setup/_registry.html.slim
```diff
@@ -0,0 +1,22 @@
...
+      button.btn.btn-danger.btn-remove-registry type="button"
+        span.fa.fa-minus aria-hidden="true"
+        |  Remove registry
+      button.btn.btn-success.btn-add-registry type="button"
+        span.fa.fa-plus aria-hidden="true"
+        |  Add another registry
+
+    .form-group.col-sm-8
+      = f.label :docker_registry_certificate_name, "Certificate"
+      = f.text_area :docker_registry_certificate, class: "form-control", placeholder: "TODO example of certificate in helper", value: certificate_docker_registry, cols: 80, rows: 18, readonly: !registry.certifiable_id.blank?
```
MaximilianMeister: @inercia regarding the TODO, should we add a real example, or just some information like "paste a valid certificate here"

vitoravelino: I would say an example of a certificate to let the user know what he really needs to paste there.

Since it's gonna be a big string, put in a helper function and call it there.

MaximilianMeister: @inercia and i think this might be a bit confusing in the end to have an example in there, would you be fine as well with just an explanation?

vitoravelino: I personally prefer an example but I guess an explanation is fine too.

MaximilianMeister:
[Bug 1039877 - docker cannot pull images from registry using untrusted CA](https://trello.com/c/YkUgukMu/272-bug-1039877-docker-cannot-pull-images-from-registry-using-untrusted-ca)

db/schema.rb
```diff
@@ -11,7 +11,25 @@
...
+
+  create_table "certificates", force: :cascade do |t|
+    t.text     "certificate", limit: 65535
+    t.date     "expiration"
+    t.datetime "created_at",                null: false
+    t.datetime "updated_at",                null: false
+  end
+
+  create_table "docker_registries", force: :cascade do |t|
+    t.string   "name",             limit: 255
```
inercia: Sorry but `name` has been renamed to `url`

MaximilianMeister:
dropped the user facing UI part to build the actual UI later in a dedicated settings page. #407 will use the model(s) from here

app/assets/javascripts/setup.js
```diff
@@ -17,4 +16,20 @@ $(function() {
       $(this).text('Show');
     }
   });
-});
\ No newline at end of file
+
+  $(document).on('click', '.btn-add-registry', function() {
```
vitoravelino: Comment this just to be safe.

app/assets/javascripts/setup.js
```diff
@@ -6,8 +6,7 @@ $(function() {
     $btnGroup.find('.btn').toggleClass('btn-default');
   });
 
-
-  $(document).on('click', '.js-toggle-overlay-settings-btn', function() {
+  $(document).on('click', '.js-toggle-overlay-settings-btn, .js-toggle-registry-settings-btn', function() {
```
vitoravelino: You can revert this line.

app/models/pillar.rb
```diff
@@ -61,7 +62,13 @@ def apply(pillars, required_pillars: [], unprotected_pillars: [])
 
     private
 
+    # all pillars that are not in the Pillar database
+    def independent_pillars
+      [:insecure_registries]
```
vitoravelino: Are we gonna use `insecure_registries`? Because we'll handle insecure, secure and mirrors.

app/models/pillar.rb
```diff
@@ -41,7 +41,8 @@ def all_pillars
         ldap_admin_group_name:   "ldap:admin_group_name",
         ldap_tls_method:         "ldap:tls_method",
         ldap_mail_attribute:     "ldap:mail_attribute",
-        cloud_framework:         "cloud:framework"
+        cloud_framework:         "cloud:framework",
+        insecure_registries:     "docker:registries"
```
vitoravelino: Are we gonna use `insecure_registries`? Because we'll handle insecure, secure and mirrors.

db/migrate/20171124131432_create_docker_registries.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDockerRegistries < ActiveRecord::Migration
+  def change
+    create_table :docker_registries do |t|
+      t.string :url
+      t.boolean :mirror
```
vitoravelino: I think `mirror` needs to be a string for the remote registry. If `nil` then not a mirror. Any other idea?

app/assets/stylesheets/velum_general.scss
```diff
@@ -85,4 +85,4 @@ dl.side-by-side {
     }
 }
 
-@import 'pages/nodes_list';
+@import 'pages/**/*';
```
vitoravelino: You can revert this.

app/assets/stylesheets/pages/setup.scss
```diff
@@ -0,0 +1,21 @@
+.registry-entry {
```
vitoravelino: This file won't be needed anymore.

app/models/pillar.rb
MaximilianMeister: yeah, this was the docker term originally for the command line flag, but it's outdated. i think we can change it now

db/migrate/20171124131432_create_docker_registries.rb
MaximilianMeister: i guess any mirror is also a docker registry, so this field just specifies if it's intended as a mirror or not? that would mean we could keep it that way

vitoravelino: How are you gonna relate the mirror registry with the remote registry without knowing? Remember the daemon.json that we need to produce in the end.

app/models/pillar.rb
ereslibre: 👍

db/migrate/20171124131334_create_certificates.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCertificates < ActiveRecord::Migration
+  def change
+    create_table :certificates do |t|
+      t.text :certificate
+      t.date :expiration
+
+      t.timestamps null: false
```
ereslibre: Minor: `null: false` is not required here as activerecord will automatically fill `created_at` and `updated_at`.

db/migrate/20171124131432_create_docker_registries.rb
ereslibre: I agree with @vitoravelino, but instead I'd say that if it's not `null` then it's a mirror on `url` mirroring `mirror`. If `mirror` is null, it's a registry.

flavio:
Can we move all the UI-related code into a dedicated PR?

MaximilianMeister:
> Can we move all the UI-related code into a dedicated PR?

if you think it's worth i can do that

spec/models/certificate_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+describe Certificate, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
ereslibre: Is this intended to be committed?

app/models/docker_registry.rb
```diff
@@ -0,0 +1,63 @@
+# model that represents a docker registry
+class DockerRegistry < ActiveRecord::Base
+  belongs_to :certificate, polymorphic: true
+
+  validates :url, presence: true
```
flavio: I would enforce uniqueness over the url

app/models/pillar.rb
```diff
@@ -41,7 +41,8 @@ def all_pillars
         ldap_admin_group_name:   "ldap:admin_group_name",
         ldap_tls_method:         "ldap:tls_method",
         ldap_mail_attribute:     "ldap:mail_attribute",
-        cloud_framework:         "cloud:framework"
+        cloud_framework:         "cloud:framework",
+        registries:              "docker:registries"
```
ereslibre: If we are using our own models I don't see why we want this here, since the `Pillar` model won't save this information but `Certificate` and `DockerRegistry` will. I think we can remove this and serve this content using the `InternalApi::V1::PillarsController` to include `DockerRegistry`'s.

db/migrate/20171124131334_create_certificates.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCertificates < ActiveRecord::Migration
+  def change
+    create_table :certificates do |t|
+      t.text :certificate
+      t.date :expiration
```
flavio: How are we going to extract this information? AFAIK we are not doing that yet.

db/migrate/20171124131432_create_docker_registries.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDockerRegistries < ActiveRecord::Migration
+  def change
+    create_table :docker_registries do |t|
+      t.string :url
```
flavio: I would enforce the uniqueness of `url`

db/migrate/20171124131334_create_certificates.rb
MaximilianMeister: good point, it's not required for now, but it can be set later through the UI i would say. we could also make this mandatory already now

spec/models/certificate_spec.rb
MaximilianMeister: there is not much to test now in this model but i could add a basic test, so we can extend it later if we need to. i'd rather not drop it

db/migrate/20171124131334_create_certificates.rb
flavio: I would drop it because we are not using it.

I would envision a feature where:
  * the expiration date is extracted from the certificate itself
  * the UI shows a warning when we approach to the expiration date and eventually sends a notification mail
  * the UI shows a big fat warning when the certificate is expired and eventually sends a notification mail

However we don't have any of these requirements right now, let's keep the codebase as small as possible.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -18,4 +18,23 @@ def pillar_contents
...
+  end
+
+  def registry_contents
+    {
+      registries: [].tap do |res|
+        DockerRegistry.all.each do |reg|
+          res << {
+            url:         reg.url,
+            mirror:      reg.mirror,
+            certificate: Certificate.find(reg.certifiable_id).certificate
```
ereslibre: `certificate: reg.certificate.certificate` should work in order to serve the certificate for this registry.

app/models/docker_registry.rb
ereslibre: It's on data level (on the migration) but I agree we can add it here too.

app/models/docker_registry.rb
```diff
@@ -0,0 +1,64 @@
...
+  validates :url, uniqueness: true
```
vitoravelino: Already here... Totally missed that

app/models/docker_registry.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    # remove old registries from the db if they were deleted in the UI
+    def cleanup(settings_params)
+      passed_registries = settings_params["registries"].collect do |r|
+        r["docker_registry_url"]
+      end
+      saved_registries = all.map(&:url)
+
+      (saved_registries - passed_registries).each do |registry|
+        r = find_by(url: registry)
```
vitoravelino: Use transaction here?

ereslibre: Or just a one liner: `DockerRegistry.where(url: saved_registries - passed_registries).destroy_all`

We should add a `before_destroy` or `after_destroy` to `DockerRegistry`, if it's the latest one referencing to a `Certificate` it will automatically `destroy` the certificate too.

app/models/certificate.rb
```diff
@@ -0,0 +1,7 @@
+# generic polymorphic model for all kinds of certificates
+# rubocop:disable Rails/HasManyOrHasOneDependent
+class Certificate < ActiveRecord::Base
+  has_many :docker_registries, as: :certifiable
```
vitoravelino: Is this right? Isn't it the other way around? A certificate belongs to a certifiable entity and this certifiable entity has one certificate? A certificate should not know about docker registry since it's a polymorphic model. Does it make sense ?

I took a look at http://guides.rubyonrails.org/v4.2/association_basics.html#polymorphic-associations and compared with their scenario and it felt wrong for me the way we are doing it.

app/models/docker_registry.rb
vitoravelino: If what I wrote in https://github.com/kubic-project/velum/pull/193/files#r160642129 is right, we won't need that because we could use `dependent: :destroy` with `has_one` and just call `r.delete`.

ereslibre: No, a `Certificate` can be linked to `DockerRegistry` or other entities in the future. Imagine a case in which:

`DockerRegistry` with `id` 1 is linked to `Certificate` with `id` 1
`DockerRegistry` with `id` 2 is linked to `Certificate` with `id` 1

If you add `dependent: :destroy`, if `DockerRegistry` with `id` `1` is removed, it will destroy the `Certificate` with `id` `1` leaving `DockerRegistry` orphan.

ereslibre: You are right @vitoravelino, the polymorphic association is the other way around.

ereslibre: Here's a suggestion because right now we are not handling properly the associations:

`CertificateService` (new model): `belongs_to :certificate` and `belongs_to :service, polymorphic: true` (so it has `certificate_id`, `service_id`, `service_type`)
`Certificate`: `has_many :certificate_service`

This would allow us to link the same certificate to any `service` (for now only `DockerRegistry`), update certificates only once and update them on several services at the same time, and also destroy `CertificateService` when a `service` is destroyed.

app/models/docker_registry.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    # create or update Certificate model
+    def configure_certificate(cert)
+      errors = []
+      return [nil, errors] if cert.blank?
+      unless Certificate.find_or_create_by certificate: cert
+        errors << "Failed to validate certificate"
+      end
+      [certificate, errors]
```
MaximilianMeister: `certificate` doesnt exist in this context here

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -13,6 +15,7 @@
   describe "GET /pillar" do
     before do
       Pillar.create pillar: "dashboard", value: "dashboard.example.com"
+      DockerRegistry.create url: "https://example.com", certificate: certificate
     end
```
vitoravelino: It would be nice to check some content here instead of just http status, no?

ereslibre: Hmm, I think I'm not following :/

ereslibre: Oh, that we expand that test to also check the contents of the response! Yeah...

vitoravelino: Exactly!

ereslibre: Done :)

app/models/docker_registry.rb
```diff
@@ -0,0 +1,53 @@
+# model that represents a docker registry
+class DockerRegistry < ActiveRecord::Base
+  has_one :certificate_service, as: :service, dependent: :destroy
+  has_one :certificate, through: :certificate_service
+
+  validates :url, presence: true, uniqueness: true, url: { schemes: ["https", "http"] }
```
MaximilianMeister: @vitoravelino i was not really sure when i added this, should the user also be able to enter `some.url.local` without the protocol prefix? then we'd need to adapt the restriction

vitoravelino: I dunno... I don't think we have an use case that needs the url without the prefix.

flavio: In the beginning I would force the user to always enter the url prefix

flavio:
@MaximilianMeister can you fix the codeclimate issue please (or silence it?)

PR #194, opened 2017-06-14
AutoYaST: disable creation of swap [bsc#1043813]

flavio:
Disables the creation of the swap partition on the nodes provisioned by AutoYaST.

This fixes bsc#1043813

Tested manually.


flavio:
[Bug 1043813 - AutoYaST profile: do not create swap](https://trello.com/c/eXRNxZpL/130-bug-1043813-autoyast-profile-do-not-create-swap)

flavio:
We are not going to do that for the 1.0 release. Let's revisit that later on.

PR #195, opened 2017-06-14
Don't be so strict on some pillar fields

mssola:


app/validators/alphanumeric_validator.rb
```diff
@@ -0,0 +1,33 @@
...
+  PROTECTED_VALUES      = (ALPHANUMERICAL_VALUES + ALPHABETICAL_VALUES).freeze
+
+  def validate_each(record, attribute, value)
+    # Base case: irrelevant pillar value, or nil value.
+    return unless relevant?(record.pillar, value)
+
+    # Pillar values that should contain only alphabetical digits.
+    return if contains?(ALPHABETICAL_VALUES, record.pillar) && value.match(/^[a-zA-Z\s]+$/)
+
+    # Pillar values that can contain any ASCII character.
```
MaximilianMeister: hmm thinking about it again, any ASCII seems not to be correct, is it?

vrothberg:
Tested locally and it behaves as I would expect it to. Thanks! :+1:

mssola:
Errors are due to flaky tests... Merging...

PR #196, opened 2017-06-14
Fixed regression introduced when changed dashboard layout

vitoravelino:
For 1.0 we changed the layout of overview page. During that process,
I accidentally changed this piece of code without reminding that
the discovery page also depended on this.

Now it separates different behavior depending on the page.

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -40,13 +40,22 @@ MinionPoller = {
         var minions = data.assigned_minions || [];
         var unassignedMinions = data.unassigned_minions || [];
 
-        MinionPoller.selectedMasters = data.assigned_minions.reduce(function(memo, minion) {
-          if (minion.role === 'master') {
-            memo.push(minion.id);
-          }
+        // for the dashboard, if we rely on radio, the first time this comes
+        // it won't detect that there's a master, so we need to rely on the data
+        // itself for counting masters
+        if (MinionPoller.renderMode === 'dashboard') {
+          MinionPoller.selectedMasters = data.assigned_minions.reduce(function(memo, minion) {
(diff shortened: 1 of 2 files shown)


vitoravelino:
Only necessary because we update master's count based on the data we fetch. 

For now I guess we will have only one master, but eventually we will have more, right? I guess when we get there this piece of code won't exist anymore. :)

vitoravelino:
If we won't have more than 1 master, I can achieve what is necessary through the templates and just revert the js code.

ereslibre:
> If we won't have more than 1 master, I can achieve what is necessary through the templates and just revert the js code.

We eventually will, after 1.0 is released.

vitoravelino:
> We eventually will, after 1.0 is released.

Ok, so let's wait for travis and concurse then.

app/assets/javascripts/dashboard.js
```diff
@@ -38,13 +38,22 @@ MinionPoller = {
...
+          MinionPoller.selectedMasters = data.assigned_minions.reduce(function(memo, minion) {
+            if (minion.role === 'master') {
+              memo.push(minion.id);
+            }
+
+            return memo;
+          }, []);
+        } else {
+          MinionPoller.selectedMasters = $("input[name='roles[master][]']:checked").map(function() {
+            return parseInt($( this ).val());
```
vitoravelino: No. It was like that before. I can change if you want. 

After the release, I'll add eslint to prevent this kind of inconsistencies. Embrace yourselves for the incoming changes! KKK. :)

ereslibre: No, just wondering to keep a consistent style :)

vitoravelino:
@mssola @ereslibre merge this?

PR #197, opened 2017-06-15
WIP: remove nodes from the cluster

mssola:

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,7 @@ Rails.application.routes.draw do
 
   resource :dashboard, only: [:index]
   resource :updates, only: [:create]
+  resources :nodes, only: [:destroy]
 
   authenticated :user do
     root "dashboard#index", as: :authenticated_root
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -139,6 +139,7 @@ MinionPoller = {
     var statusHtml;
     var checked;
     var masterHtml;
+    var destroyHtml;
 
     switch(minion.highstate) {
       case "not_applied":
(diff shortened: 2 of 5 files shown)


mssola:
https://trello.com/c/uO999RGs/232-5-node-removal-salt-tasks

config/routes.rb
```diff
@@ -5,6 +5,7 @@
 
   resource :dashboard, only: [:index]
   resource :updates, only: [:create]
+  resources :nodes, only: [:destroy]
```
MaximilianMeister: with having a `NodesController` do you think we can also add the bootstrap api to its responsibility? so `destroy` is a member method and `bootstrap` is a collection method

mssola: I don't know. This is a matter for the future I'd say. The dashboard and the setup controllers do not really follow any best practices, and we simply dumped methods into those. After the first release we'll have to re-organize things a bit.

For this PR, having this functionality here makes sense (and the "adding nodes" feature could re-use it but then again, it's a matter for re-organization in the future 😉).

vitoravelino: I totally agree on moving stuff to `NodesController` and being more resources oriented but only after the release. We'll have a sprint to cleanup stuff after the release and I think we should also prioritize this.

In #173 I added the actions to the dashboard controller but it should definitely be in `NodesController`. =/

PI-Victor:
hmm, what's the current status of this?

ereslibre:
@PI-Victor It hasn't got priority since this isn't required for 1.0, so I guess it's a post-1.0 consideration :)

mssola:
Closing since I didn't work on this for a long time...

Issue #192, opened 2017-06-13
Update test suite


vrothberg:
Random tests break on random builds for random versions of Ruby. That's pretty bad as a reliable CI is our backbone.

Let's use this issue to track progress on this topic.

vitoravelino:
I've been seeing two problems so far that it's becoming annoying for feature specs:

- not being able to reach an address, complains to dns or something like that
- false negatives (always when js: true)

For the first one, I guess we can't do anything since it's inside Travis and not on us. This one doesn't worry me.

The other one, we can try to use Chrome headless that was just shipped on version 59 (last hope for stable headless testing - haha) or `capybara-webkit` instead of `poltergeist` and see if it's more stable. To contextualize, PhantomJS development was kind of discontinued in favor of Chrome headless (the author started working full-time on Chrome if I'm not wrong). I found a few  resources/examples and I'd like to try Chrome headless. :)

vrothberg:
Awesome, @vitoravelino. Thanks for your input! I am currently looking at potential sources related to dns etc. I'd like to experiment with some Travis settings, e.g. the number of concurrent jobs.

vitoravelino:
Just to document the first error I've mentioned.

```
Capybara::Poltergeist::StatusFailError:
       Request to 'http://127.0.0.1:46477/' failed to reach server, check DNS and/or server status
```

ereslibre:
https://github.com/teampoltergeist/poltergeist/issues/896 might be related.

flavio:
I think that moving to chrome headless would be right choice, I just don't know how much effort it would take right now.

Do you have any idea?

vitoravelino:
We gotta try... from what I searched about it, only a few lines in capybara.rb driver config. Not sure about the side effects. I guess it will need chromedriver but not sure if it downloads automatically and works out of the box.

twalpole:
No, chromedriver is not downloaded automatically so you will have to manually download the latest version of that.  Also, if your tests require multiple windows they can't currently be run with headless chrome due to unsupported functionality/bugs

vitoravelino:
Yeah, you are right. I found this: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/11680/diffs

Need both chrome and chromedriver and selenium-webdriver gem.

vitoravelino:
> Also, if your tests require multiple windows they can't currently be run with headless chrome due to unsupported functionality/bugs

[...]

PR #198, opened 2017-06-15
Fixed flaky tests

mssola:

### DIFF
--- a/spec/features/admin_node_update_feature_spec.rb
+++ b/spec/features/admin_node_update_feature_spec.rb
@@ -41,20 +41,30 @@ feature "Manage nodes updates feature" do
     end
     # rubocop:enable RSpec/MultipleExpectations
 
+    # rubocop:disable RSpec/ExampleLength
     scenario "User clicks on 'Reboot to update'", js: true do
-      allow(::Velum::Salt).to receive(:call)
+      allow(::Velum::Salt).to receive(:call).and_return(true)
 
       # clicks on "Update admin node"
       find(".update-admin-btn").click
 
+      # wait modal to appear
+      wait_until { page.has_text?("Reboot to update") }
+
       # clicks on "Reboot to update"
       find(".reboot-update-btn").click
 
-      # rubocop:disable RSpec/MessageSpies
-      expect(::Velum::Salt).to receive(:call)
-      # rubocop:enable RSpec/MessageSpies
+      wait_until { page.has_text?("Rebooting...") }
       expect(page).to have_content("Rebooting...")
+
+      # NOTE: The check below is flaky: sometimes it passes and sometimes it
+      # doesn't. We believe that this is Capybara/Poltergeist to blame. Since we
+      # already have this check in `spec/controllers/updates_controller_spec.rb`
+      # and getting the "Rebooting..." flashy message already guarantees that
+      # the controller was reached, we have simply commented out the line below.
+      # expect(::Velum::Salt).to have_received(:call).once
     end
+    # rubocop:enable RSpec/ExampleLength
   end
 
   scenario "Admin node has an update available (failed to update)", js: true do
--- a/spec/features/dashboard_feature_spec.rb
+++ b/spec/features/dashboard_feature_spec.rb
@@ -27,6 +27,7 @@ feature "Dashboard" do
       visit authenticated_root_path
 
       el = find("#download-kubeconfig")
+      wait_until { !find("#download-kubeconfig")[:disabled] }
       expect(el[:disabled]).to be_falsey
     end
     # rubocop:enable RSpec/ExampleLength
--- a/Gemfile
+++ b/Gemfile
@@ -61,8 +61,8 @@ unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
     gem "vcr"
     gem "webmock", require: false
     gem "simplecov", require: false
-    gem "capybara"
-    gem "poltergeist", require: false
+    gem "capybara", "~> 2.14.3"
+    gem "poltergeist", "~> 1.15.0", require: false
     gem "json-schema"
     gem "timecop"
     gem "codeclimate-test-reporter", "~> 1.0.0", require: nil
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,7 +36,7 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
-    addressable (2.5.0)
+    addressable (2.5.1)
       public_suffix (~> 2.0, >= 2.0.2)
     arel (6.0.4)
     ast (2.3.0)
@@ -48,7 +48,7 @@ GEM
       sass (>= 3.3.4)
     brakeman (3.5.0)
     builder (3.2.3)
-    capybara (2.12.0)
+    capybara (2.14.3)
       addressable
       mime-types (>= 1.16)
       nokogiri (>= 1.3.3)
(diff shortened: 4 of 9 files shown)


mssola:
I'm not done, but I'm importing something from Portus that might help

mssola:
The timeout issue from poltergeist seems to be fixed. There are other flaky tests though ...

vitoravelino:
With my changes it got better but we still get the expectation problem with the Salt class sometimes. =/

vitoravelino:
WE DID IT, BOYS! :metal: 

I guess we are good to merge this one. Good job! :clap:

PI-Victor:
i'm not 100% but i think the tests for this PR need to be re-triggered, for what i understand the CI went down today :/

MaximilianMeister:
> i'm not 100% but i think the tests for this PR need to be re-triggered, for what i understand the CI went down today :/

retriggered via concourse ui, but i think this fix is more about travis, or?

MaximilianMeister:
also with the Gemfile changes it might be that the rpm will not build due to missing dependencies, but i can try to provide the correct versions, once this is in a mergable state.

also commits could be squashed a bit :)

mssola:
I've been re-triggering builds just to make sure that it works. Once that passes, I'll review the code and squash it all.

mssola:
@ereslibre @vitoravelino this is ready to be reviewed. I've also squashed all commits.

spec/features/admin_node_update_feature_spec.rb
```diff
@@ -42,18 +42,25 @@
...
+      # wait modal to appear
+      wait_until { page.has_text?("Reboot to update") }
+
       # clicks on "Reboot to update"
       find(".reboot-update-btn").click
 
-      # rubocop:disable RSpec/MessageSpies
-      expect(::Velum::Salt).to receive(:call)
-      # rubocop:enable RSpec/MessageSpies
       expect(page).to have_content("Rebooting...")
```
vitoravelino: This is also causing some failures... Maybe use wait_until before the expect?

mssola: Done!

vitoravelino:
Will re-trigger a few times just to make sure that is stable. Don't worry.

vitoravelino:
Travis passed 4 times in a row without any failure! :tada: 

LGTM

MaximilianMeister:
before merging this, we need to check the packaging side and the new dependencies

cc @jordimassaguerpla

vrothberg:
Adding issue #192 for cross referencing and the picture of a cute cat!

MaximilianMeister:
ran it locally, green as well. :+1: @vitoravelino you could re-trigger it again if you like by force push. if not, we should be fine as well with the 2 confirmations

ereslibre:
Now the problem with the current e2e tests is that the velum docker image
with the new Gemfile.lock will be reused, and further tests will fail until
we clean the two images again.

vitoravelino:
> Now the problem with the current e2e tests is that the velum docker image
with the new Gemfile.lock will be reused, and further tests will fail until
we clean the two images again.

Rebase then?

ereslibre:
No, no, I mean that for other tests to pass on our CI infra we need to
clean up the docker images again after this one has passed. Other PR will
fail because they don't share the changes in the Gemfile.lock.

It's not about this PR, is about the e2e will reuse the velum docker image
generated for this run in future runs, so we need to clean them again.

MaximilianMeister:
could someone run the e2e-tests locally?

> It's not about this PR, is about the e2e will reuse the velum docker image
> generated for this run in future runs, so we need to clean them again.

i can do that right after this one is green, or we merge it right away, and the other PR runs should be fine, did i understand that correctly?

ereslibre:
> i can do that right after this one is green, or we merge it right away, and the other PR runs should be fine, did i understand that correctly?

So the problem is that the docker image has some dependencies in this PR than in other PR's, but the velum docker development image is reused across runs. We can fix that by when starting on every e2e test run we clean `sles12/velum:0.0` and `sles12/velum:development`, so they get regenerated on every run, and we wouldn't need to explicitly clean them manually.

MaximilianMeister:
> We can fix that by when starting on every e2e test run we clean sles12/velum:0.0 and sles12/velum:development, so they get regenerated on every run, and we wouldn't need to explicitly clean them manually.

that is likely much better, even though it slows down testing a bit :+1:

mssola:
@MaximilianMeister I'm still failing with the e2e tests, but that's more probably something from my system. Can we merge this ?

MaximilianMeister:
> I'm still failing with the e2e tests, but that's more probably something from my system. Can we merge this ?

:+1:

spec/features/admin_node_update_feature_spec.rb
```diff
@@ -41,20 +41,30 @@
...
+      # wait modal to appear
+      wait_until { page.has_text?("Reboot to update") }
+
       # clicks on "Reboot to update"
       find(".reboot-update-btn").click
 
-      # rubocop:disable RSpec/MessageSpies
-      expect(::Velum::Salt).to receive(:call)
-      # rubocop:enable RSpec/MessageSpies
+      wait_until { page.has_text?("Rebooting...") }
```
twalpole: This is no different than the following line - If it's passing because of this line it means the max wait for the next line isn't long enough, so just increase it there -  `expect(page).to have_content("Rebooting...", wait: 10)`

spec/features/admin_node_update_feature_spec.rb
```diff
@@ -41,20 +41,30 @@
...
+    # rubocop:disable RSpec/ExampleLength
     scenario "User clicks on 'Reboot to update'", js: true do
-      allow(::Velum::Salt).to receive(:call)
+      allow(::Velum::Salt).to receive(:call).and_return(true)
 
       # clicks on "Update admin node"
       find(".update-admin-btn").click
 
+      # wait modal to appear
+      wait_until { page.has_text?("Reboot to update") }
```
twalpole: This is no different than `expect(page).to have_text("Reboot to update")`

spec/features/dashboard_feature_spec.rb
```diff
@@ -27,6 +27,7 @@
       visit authenticated_root_path
 
       el = find("#download-kubeconfig")
+      wait_until { !find("#download-kubeconfig")[:disabled] }
```
twalpole: This line and the following can just be written as `expect(el).not_to match_css('[disabled]')` or `expect(el).to match_css('not([disabled])')`

vitoravelino: Does the max wait time also applies to elements? I didn't know that.

twalpole: @vitoravelino max wait time applies to pretty much every Capybara finder and matcher, except for `first` and `all` (unless you specify one of the count options which forces them into waiting mode as well)

spec/features/admin_node_update_feature_spec.rb
mssola: A question from my side (I've looked it up in a quick search and it's not clear to me), is the `wait` time blocking ? That is, will it wait for 10 seconds and then perform the check ? or will it quit as soon as the expectation matches ?

mssola: Moreover, I see that the default is `default_max_wait_time`, which in our case is 5. So raising that to 10 would have the same effect. Except that what if it needs more than 10 seconds ? Should we increase this value to 60 or 120 just to be sure ? Does it have any side-effects doing so ?

twalpole: It retries the expectation/find until they pass for up to the wait time. If the wait time expires the expectation/find fails and raises an error. You could set longer but any reasonable UI shouldn’t be taking more than 10 seconds max in a test environment, and usually less

mssola: @twalpole thanks a lot for the clarification 👍

PR #199, opened 2017-06-16
Remove phantomjs gem

ereslibre:
Check if this makes our Travis CI more stable.

Add phantomjs to travis configuration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,20 @@ matrix:
   allow_failures:
     - rvm: ruby-head
 
+cache:
+  directories:
+    - travis_phantomjs
+
 before_install:
   - gem update --system
+  - phantomjs --version
+  - export PATH=$PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
+  - hash -d phantomjs || true
+  - phantomjs --version
+  - if [ $(phantomjs --version) != '2.1.1' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi
+  - if [ $(phantomjs --version) != '2.1.1' ]; then wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2; fi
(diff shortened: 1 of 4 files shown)


vitoravelino:
You forgot to remove `phantomjs:         Phantomjs.path,` in `support/capybara.rb`.

ereslibre:
Leaving it open in case you think this can help anyhow or is somehow cleaner (given that we also have `phantomjs` installed in our devenv).

Feel free to close if you think this is not the right approach.

vitoravelino:
I'm good to go with this if there is an improvement on timing performance. 

Be aware that with this change we need to have phantomjs installed on our dev environment since it won't work out of the box anymore.

ereslibre:
@vitoravelino That was the idea, the `caasp-devenv` environment [installs it automatically](https://github.com/kubic-project/caasp-devenv/blob/5d926c72c5a0f6d4408f7e9f39defcda81550c6f/start#L99-L102).

mssola:
I don't know about this one:

- As we discovered, Phantom JS was not the root of all the instability really.
- It adds quite some complexity on the Travis configuration.
- Do we have a way to track which specific version of Phantom JS will be used in the `caasp-devenv` ?

ereslibre:
Yes, but we still have it installed, I don't see a reason for rspec having to download it again when we already have it...

> It adds quite some complexity on the Travis configuration.

IMHO is the right place to have this add-on, since having gems downloading dependencies is something hacky anyway, and can lead to other problems (if a failure happens when downloading it because of networking issues, DNS...) is the step previous to the rspec running what will fail (the step of installing requirements), not the rspec suite itself the one that will fail. When I have installed all the gems I need I expect my project to have everything locally, and not having to still download anything external.

> Do we have a way to track which specific version of Phantom JS will be used in the caasp-devenv ?

In the development environment we have full control of the version we are using: https://github.com/kubic-project/caasp-devenv/blob/5d926c72c5a0f6d4408f7e9f39defcda81550c6f/start#L99-L102

We could add something generic like a file in the root `.phantomjs.version`, that both the `caasp-devenv` project and Travis would read and download that version.

mssola:
I see. Fine by me then 👍 

> In the development environment we have full control of the version we are using: https://github.com/kubic-project/caasp-devenv/blob/5d926c72c5a0f6d4408f7e9f39defcda81550c6f/start#L99-L102
>
> We could add something generic like a file in the root .phantomjs.version, that both the caasp-devenv project and Travis would read and download that version.

Good idea.

PR #200, opened 2017-06-19
prevent caching of pages (bsc#1044237)

MaximilianMeister:
after logout you can still access the application pages via
the browser's back button

disabling caching completely prevents this, as all pages contain
sensitive information

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,11 +5,19 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_user!
   before_action :redirect_to_setup
   protect_from_forgery with: :exception
+  # https://bugzilla.suse.com/show_bug.cgi?id=1044237
+  before_action :prevent_page_cache
 
   private
 
   def redirect_to_setup
     return true unless signed_in?
     redirect_to setup_path if Minion.assigned_role.count.zero?
+  end
+
+  def prevent_page_cache
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = 1.year.ago.to_s
   end
 end


MaximilianMeister:
[Bug 1044237 - Dashboard-session data accessible after logout](https://trello.com/c/rd3PipYB/137-bug-1044237-dashboard-session-data-accessible-after-logout)

flavio:
Isn't this going to affect page caching on the whole velum site? What are the side effects of that?

MaximilianMeister:
> Isn't this going to affect page caching on the whole velum site? What are the side effects of that?

there should be no negative side-effects as we're constantly rendering the more dynamic content via javascript, and the rest of the pages render pretty quickly on the server side. i tried other cache storages as well (cache_store) but that doesn't fix the issue described in the bugreport.

we could cache some sites and others not, but i believe right now, all sites contain somewhat sensitive information

app/controllers/application_controller.rb
```diff
@@ -5,11 +5,19 @@ class ApplicationController < ActionController::Base
...
   private
 
   def redirect_to_setup
     return true unless signed_in?
     redirect_to setup_path if Minion.assigned_role.count.zero?
   end
+
+  def prevent_page_cache
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
```
vitoravelino: This is only necessary to browsers that doesn't support HTTP 1.1. Are we supporting it? All the latest browsers are already implementing HTTP 2, I guess we can get rid of 1.0 support. Just guessing...

app/controllers/application_controller.rb
```diff
@@ -5,11 +5,19 @@ class ApplicationController < ActionController::Base
...
 
   private
 
   def redirect_to_setup
     return true unless signed_in?
     redirect_to setup_path if Minion.assigned_role.count.zero?
   end
+
+  def prevent_page_cache
+    response.headers["Cache-Control"] = "no-cache, no-store"
```
vitoravelino: To be more complete: 

```ruby
response.headers["Cache-Control"] = "no-cache, max-age=0, must-revalidate, no-store"
```

app/controllers/application_controller.rb
```diff
@@ -5,11 +5,19 @@ class ApplicationController < ActionController::Base
...
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = 1.year.ago.to_s
```
vitoravelino: Is this really necessary? I don't think so... I testes just with the first line and it worked.

app/controllers/application_controller.rb
MaximilianMeister: yes, i read this as well, and i decided to keep support for older browsers to be more inclusive. not sure if it makes sense to explicitly not support them, @flavio any thoughts?

app/controllers/application_controller.rb
MaximilianMeister: yes, you probably have an up-to-date browser. it's again browser specific. this is to include browsers that still work on `HTTP/1.0` - if we explicitly not want to support that case, i can drop it

mssola:
Agreed with @ereslibre. This is really not the way to go. Plus, these headers might be further modified if the customer sets NGinx and similar in front of it all 😁

app/controllers/application_controller.rb
kiall: These should absolutely be kept. Browsers are only 1 thing which caches content. Many enterprises and ISPs will have caching proxy servers for all employee/customer traffic, and often, these will require the traditional headers to correctly function.

MaximilianMeister:
> while the browser what is doing is basically serving the same client-side code it had (since we invalidated the session on our side nothing can be really done -- interaction would just redirect you to the login page).

of course, but you can still see the input fields of the form and all other internal details of the cluster, do you know another way how we could prevent that?

ereslibre:
> of course, but you can still see the input fields of the form and all other internal details of the cluster, do you know another way how we could prevent that?

You can't, and my point being is: we shouldn't even care about it IMO. We are invalidating the session on our side, and that's all we can do. Whatever the browser does is up to it. It can even completely ignore `Cache-Control` or `Pragma` (it shouldn't, but nothing can prevent it from ignoring those headers).

For the input fields to be visible in this case, I'd have to login, fill them, logout, don't close the browser, don't close the tab, and in that same tab, click back. And of course, I would only be allowed to read the fields (no interaction allowed, cause server side will just reject any request).

MaximilianMeister:
@ereslibre thanks for the input, let's see what the outcome in the bug is, I am just a bit worried as it got flagged as P1

ukbelch:
Page caching is normal. I don't see this as a security bug, as there is no viable attack vector here.

PI-Victor:
can we close this?

ereslibre:
Closing since the bug report was resolved as invalid.

PR #201, opened 2017-06-20
Add unassigned nodes after cluster bootstrap

vitoravelino:
After the initial selection of the cluster (discovery page), new nodes can eventually come up and user might want to add them. This PR makes that available with a dedicated page similar to the discovery one.

- [X] UI
- [X] Routes
- [x] Feature's specs
- [x] Controller's specs

 [UI: create the add unassigned nodes to the cluster](https://trello.com/c/yzU3t5EC/229-5-ui-create-the-add-unassigned-nodes-to-the-cluster)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,10 @@ class ApplicationController < ActionController::Base
 
   def redirect_to_setup
     return true unless signed_in?
-    redirect_to setup_path if Minion.assigned_role.count.zero?
+    redirect_to setup_path if no_setup?
+  end
+
+  def no_setup?
+    Minion.assigned_role.count.zero?
   end
 end
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -7,6 +7,8 @@ require "velum/suse_connect"
 class DashboardController < ApplicationController
   include Discovery
 
+  rescue_from Minion::NonExistingNode, with: :node_not_found
+
   # TODO: move autoyast to its own controller (following a different logic flow). It would never get
   # authenticated (as login/password -- since it's machines requesting this endpoint). It would
   # never get redirected to setup the cluster, and it should actually read some security setting for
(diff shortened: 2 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -10,6 +10,10 @@ class ApplicationController < ActionController::Base
 
   def redirect_to_setup
     return true unless signed_in?
-    redirect_to setup_path if Minion.assigned_role.count.zero?
+    redirect_to setup_path if no_setup?
+  end
+
+  def no_setup?
```
mssola: 1. Add documentation
2. I like methods to be affirmative (aka without "no_*" and stuff). For example, a better name would be `setup_done?`.

app/controllers/dashboard_controller.rb
```diff
@@ -72,4 +76,49 @@ def kubectl_config
...
+  end
+
+  private
+
+  def update_nodes_params
+    params.require(:roles).permit(worker: [])
+  end
+
+  def redirect_to_dashboard
+    # no_setup? -> no minions to be added
```
mssola: leftover ?

config/routes.rb
```diff
@@ -6,6 +6,9 @@
   resource :dashboard, only: [:index]
   resource :updates, only: [:create]
 
+  get "/assign_nodes", to: "dashboard#unassigned_nodes"
+  post "/assign_nodes", to: "dashboard#assign_nodes"
```
mssola: I don't like this routing scheme, since we are putting too much in the dashboard controller and it's not following REST. Since we are in a bit of a hurry, let's go with it, and after the release we can re-evaluate this. Until then, could you add a `TODO` mark ?

vitoravelino: Me neither as I mentioned here https://github.com/kubic-project/velum/pull/197#discussion_r122936079. Don't worry, I have a list of stuff here to be refactored/reviewed after 1.0. :wink:

app/controllers/dashboard_controller.rb
vitoravelino: Yes.

PR #205, opened 2017-06-21
Add 'grains_refresh_every' config on minions

grahamhayes:


app/views/dashboard/autoyast.xml.erb
```diff
@@ -10,6 +10,7 @@
         <source><![CDATA[
         #!/bin/sh
         echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
+        echo "grains_refresh_every: 10" >> /etc/salt/minion.d/master.conf
```
ereslibre: Why not adding it to another file like `/etc/salt/minion.d/grains_refresh.conf`?

I'm okay with adding this temporarily here, but shouldn't this configuration be managed by salt itself? We can trigger a restart of the salt-minion when this configuration changes, in theory we only need to set the master here.

grahamhayes: The problem with doing it in salt, it that when it gets updated, we need to restart the minion. Doing that in a run will cut off the rest of the highstate / orchestration step currently in progress.

ereslibre: Not sure if it supports a `SIGHUP`-like signal that only reloads the config...

Approving though.

grahamhayes:
@ereslibre I updated the file name - good catch

app/views/dashboard/autoyast.xml.erb
```diff
@@ -9,7 +9,8 @@
         <interpreter>shell</interpreter>
         <source><![CDATA[
         #!/bin/sh
-        echo "master: <%= @controller_node %>" > /etc/salt/minion.d/master.conf
+        echo "master: <%= @controller_node %>" > /etc/salt/minion.d/minion.conf
```
ereslibre: About this filename... it was fine as `master.conf` IMHO like: this is where the master is... (contents are: `master: location`)

kiall: minion.conf is used on the admin node... So this at least makes them consistent. Lets change them later, as this is blocking update testing etc

PR #206, opened 2017-06-22
bsc#1043592 Add support for accepting nodes

grahamhayes:
Turning off `auto_accept` for salt means we need
to have a mechanism to accept nodes on a node
by node basis into the cluster.

This adds a pending nodes table that will show
all nodes that have not had their key accepted in salt
yet.

Current Issues: 
- [ ] When clicking "Accept Node" it does not disappear from the list immediately

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -32,6 +32,7 @@ MinionPoller = {
       cache: false,
       success: function(data) {
         var rendered = "";
+        var pendingRendered = "";
         var allApplied = true;
         var updateAvailable = false;
         var updateAvailableNodeCount = 0;
@@ -41,6 +42,7 @@ MinionPoller = {
         // want to account for them.
         var minions = data.assigned_minions || [];
         var unassignedMinions = data.unassigned_minions || [];
+        var pendingMinions = data.pending_minions || [];
 
         // for the dashboard, if we rely on radio, the first time this comes
         // it won't detect that there's a master, so we need to rely on the data
@@ -78,6 +80,23 @@ MinionPoller = {
           }
         }
         $(".nodes-container tbody").html(rendered);
+
+        // Build Pending Nodes display table
+
+        for (i = 0; i < pendingMinions.length; i++) {
+          pendingRendered += MinionPoller.renderPendingNodes(pendingMinions[i])
+        }
+        $(".pending-nodes-container tbody").html(pendingRendered);
+
+        // Show  / Hide the table depending the presence of pending nodes
+
+        if (pendingMinions.length == 0) {
+          $(".pending-nodes-container .panel-body").hide()
+          $(".pending-nodes-container #accept-all").attr('disabled', true)
+        } else {
+          $(".pending-nodes-container .panel-body").show()
+          $(".pending-nodes-container #accept-all").attr('disabled', false)
+        }
 
         MinionPoller.handleAdminUpdate(data.admin || {});
 
@@ -118,6 +137,15 @@ MinionPoller = {
       // make another request only after the last one finished
       setTimeout(MinionPoller.request, 5000);
     });
+  },
+
+  renderPendingNodes: function(pendingMinion) {
+    return "\
+      <tr> \
+        <td>" + pendingMinion + "</td>\
+        <td><a class='accept-minion' href='#' data-minion-id='" + pendingMinion +  "'>Accept Node</a></td>\
+      </tr> \
+    ";
   },
 
   handleAdminUpdate: function(admin) {
@@ -253,6 +281,27 @@ MinionPoller = {
     return MinionPoller.renderDiscovery(minion, true);
   }
 };
+
+
+$('body').on('click', '.accept-minion', function(e) {
+  var $link = $(this);
+  var selector = ''
+  e.preventDefault();
+
+  if ($link[0].id == "accept-all") {
+    selector = '*'
+  } else {
+    selector = $link.data('minionId')
+  }
+  $link.prop('disabled', true);
+
+  $.ajax({
+    url: '/accept-minion',
+    method: 'POST',
+    data: {minion_id: selector}
+  })
+});
+
 
 // reboot to update admin node handler
 $('body').on('click', '.reboot-update-btn', function(e) {
--- a/app/assets/stylesheets/pages/nodes_list.scss
+++ b/app/assets/stylesheets/pages/nodes_list.scss
@@ -31,6 +31,23 @@
   }
 }
 
+.pending-nodes-container {
+  .panel-title {
+    display: inline-block;
+  }
+
+  form {
+    height: 450px;
+    margin-bottom: $line-height-computed;
+    overflow-x: hidden;
+    overflow-y: scroll;
+
+    table {
+      margin-bottom: 0px;
+    }
+  }
+}
+
 .nodes-summary {
   .fa-info-circle {
     margin-left: 5px;
@@ -100,4 +117,8 @@
   .modal-dialog {
     width: 450px;
   }
+}
+
+.pending-nodes-container {
+  margin-top: 15px
 }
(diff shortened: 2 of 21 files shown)


grahamhayes:
kubic-project/caasp-container-manifests#62 allows the CA and admin node to be pre accepted

grahamhayes:
screenshot from 2017-06-22 16-54-55
screenshot from 2017-06-22 16-49-13
screenshot from 2017-06-22 15-17-51

inercia:
Is there any way to accept all the nodes?

grahamhayes:
@inercia not right now, but it would just be a matter of adding a "Add All" button that has "*" as the minion ID

lib/velum/salt.rb
```diff
@@ -41,6 +41,26 @@ def self.minions
...
     end
 
+    # Returns the minions that have not been accepted into the cluster
+    def self.pending_minions
+      res = perform_request(endpoint: "/", method: "post",
+                            data: { client: "wheel",
+                                    fun:    "key.list" ,
+                                    match:  "all",
+                                    tgt:    "ca" })
+      JSON.parse(res.body)["return"].first["data"]["return"]["minions_pre"]
```
mssola: You should handle errors as well. That is, if there was an error when executing this, I'm sure `JSON.parse(res.body)["return"].first` won't contain what you are trying to fetch, and so it will raise an exception.

lib/velum/salt.rb
```diff
@@ -41,6 +41,26 @@ def self.minions
...
+    end
+
+    # Accepts a minion into the cluster
+    def self.accept_minion(minion_id: "")
+      res = perform_request(endpoint: "/", method: "post",
+                            data: { client: "wheel",
+                                    fun:    "key.accept" ,
+                                    match:  minion_id,
+                                    tgt:    "ca" })
+      JSON.parse(res.body)["return"].first["data"]["return"]["minions"]
```
mssola: Same here.

grahamhayes:
@mssola 

> There needs to be some space between the table below and the "Bootstrap cluster" button. 
Yeap adding now.

> I had all minions accepted, but then using salt-key I manually deleted a key and it's still listed as accepted on the table above (not listed when using salt-key -L). I have to admit that after a while it got re-accepted again so the output was fine (I haven't removed the auto_accept config).

Unfortunately, this is due to how we manage minions - we keep a second state store about the cluster, and it is not updated when nodes are deleted from salt.

lib/velum/salt.rb
grahamhayes: Yeap - this was a quick push to get a quick round of feedback on the concept itself. Will add it now

lib/velum/salt.rb
grahamhayes: Same as above

mssola:
Besides the known issue you mentioned, when you click "Accept node", forcing a page refresh takes a *long* time ... I don't know if this is a problem on `caasp-devenv` 😕

grahamhayes:
@mssola yes - it is unfortunately.

We do not have a "admin" node in `kubic-project/caasp-devenv` so salt is waiting for the node to time out for before returning.

I have been running `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-key -d admin` to speed things up

mssola:
On the "Pending nodes" table I have two further comments:

- It's weird when it's empty. I'd either set an "empty table" message on the lines of "All nodes have been accepted for the bootstrapping process", or something like that.
- Stating "Pending nodes" can be confusing for the end user that knows nothing about salt. We should add more info around that.

grahamhayes:
> It's weird when it's empty. I'd either set an "empty table" message on the lines of "All nodes have been accepted for the bootstrapping process", or something like that.

Yeah - thats a better idea than just collapsing the panel - I will add it.

> Stating "Pending nodes" can be confusing for the end user that knows nothing about salt. We should add more info around that.

Yeah - I can add some help text - and we can iterate on the exact content

vitoravelino:
Just opened https://github.com/kubic-project/velum/pull/208 that impacts a bit the discovery page. Nothing to worry about.

lib/velum/salt.rb
```diff
@@ -41,6 +41,26 @@ def self.minions
...
+      JSON.parse(res.body)["return"].first["data"]["return"]["minions_pre"]
+    end
+
+    # Accepts a minion into the cluster
+    def self.accept_minion(minion_id: "")
+      res = perform_request(endpoint: "/", method: "post",
+                            data: { client: "wheel",
+                                    fun:    "key.accept" ,
+                                    match:  minion_id,
+                                    tgt:    "ca" })
```
ereslibre: Same here.

lib/velum/salt.rb
```diff
@@ -41,6 +41,26 @@ def self.minions
       JSON.parse(res.body)["return"].first
     end
 
+    # Returns the minions that have not been accepted into the cluster
+    def self.pending_minions
+      res = perform_request(endpoint: "/", method: "post",
+                            data: { client: "wheel",
+                                    fun:    "key.list" ,
+                                    match:  "all",
+                                    tgt:    "ca" })
```
ereslibre: I think you can remove tgt from this call, since the wheel client is executed on the master, and this argument will probably be ignored.

grahamhayes:
e2e will pass with this change. what we will need to update is the e2e tests for kubic-project/salt#131 

This can merge before, it will just be blank until kubic-project/salt#131 is merged

ereslibre:
> e2e will pass with this change. what we will need to update is the e2e tests for kubic-project/salt#131

The devenv takes the configuration from the `salt` repo, so once kubic-project/salt#131 gets merged, they will be applied on the e2e tests too. It should be the latest one to be merged then.

> This can merge before, it will just be blank until kubic-project/salt#131 is merged

You're absolutely right. We can get this one merged before. We need to fix the specs though so they are green and 100% coverage. Please, tell me if I can help.

grahamhayes:
The last push should have added the 100% coverage + passing

grahamhayes:
screenshot from 2017-06-23 11-56-53
screenshot from 2017-06-23 11-55-14

New spacing

grahamhayes:
@inercia I added an "Accept All Nodes" button as well

grahamhayes:
So - should we let #208 clean up this, and add the required text - having the functionality in place is vital to get started on e2e tests / qa tests and merging kubic-project/salt#131

lib/velum/salt.rb
grahamhayes: Unfortunately not - if you leave it as '*', all nodes will return the data.

app/views/dashboard/_pending_nodes.html.slim
```diff
@@ -0,0 +1,23 @@
+.pending-nodes-container data-url=authenticated_root_path
+  .panel.panel-default
+    .panel-heading
+      h3.panel-title Pending Nodes
+
+      = link_to '#', id: "accept-all", class: "btn btn-xs btn-default pull-right accept-minion" do
```
vitoravelino: I guess this could be a button instead of an anchor.

app/assets/javascripts/dashboard.js
```diff
@@ -79,6 +81,23 @@ MinionPoller = {
...
+        for (i = 0; i < pendingMinions.length; i++) {
+          pendingRendered += MinionPoller.renderPendingNodes(pendingMinions[i])
+        }
+        $(".pending-nodes-container tbody").html(pendingRendered);
+
+        // Show  / Hide the table depending the presence of pending nodes
+
+        if (pendingMinions.length == 0) {
+          $(".pending-nodes-container .panel-body").hide()
+          $(".pending-nodes-container #accept-all").attr('disabled', true)
```
vitoravelino: If we change to button, we can change this from `attr` to `prop`.

app/assets/stylesheets/pages/nodes_list.scss
```diff
@@ -31,6 +31,23 @@
   }
 }
 
+.pending-nodes-container {
```
vitoravelino: You can remove this and add `.pending-nodes-container` to https://github.com/kubic-project/velum/blob/a8ed51120b3278611730fa645dfd3cee1b5ab39b/app/assets/stylesheets/pages/nodes_list.scss#L17

```sass
.nodes-container, 
.pending-nodes-container
```

PR #208, opened 2017-06-23
ui: discovery page layout enhancement

vitoravelino:
This is a simple cosmetic trying to improve the UX of our end users. I know that there are some duplicated code but decided to do that way to make things easier to read/understand and simpler code-wise (less js changes).

## Before
screenshot from 2017-06-23 08-20-35
screenshot from 2017-06-23 08-20-12

## After
screenshot from 2017-06-23 08-13-43
screenshot from 2017-06-23 08-14-23

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,4 +7,8 @@ module ApplicationHelper
     email = owner.nil? ? nil : owner.email
     gravatar_image_tag(email)
   end
+
+  def any_minion?
+    Minion.any?
+  end
 end
--- a/app/views/setup/discovery.html.slim
+++ b/app/views/setup/discovery.html.slim
@@ -1,45 +1,53 @@
-.alert.alert-info.alert-dismissible role="alert"
-  button.close[type="button" data-dismiss="alert" aria-label="Close"]
-    span[aria-hidden="true"]
-      | &times;
-  i.fa.fa-4x.pull-left aria-hidden="true"
-  span
-    | After choosing the master and clicking "Bootstrap cluster" all the other selected nodes will be set to the worker role
+h1 Bootstrap Cluster
 
-h1 Nodes
+.panel.panel-default.discovery-empty-panel class=('hide' if any_minion?)
+  .panel-heading
+    h3.panel-title No nodes found
+  .panel-body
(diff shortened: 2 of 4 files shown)


vitoravelino:
@MaximilianMeister fixed! 

Nice catch! I didn't realize that. Hehe. :smile:

app/views/setup/discovery.html.slim
```diff
@@ -1,44 +1,52 @@
...
-              = radio_button_tag "roles[master][]", m.id
-
-    .clearfix.text-right.steps-container
+h1 Bootstrap Cluster
+
+.panel.panel-default.discovery-empty-panel class=('hide' if any_minion?)
+  .panel-heading
+    h3.panel-title No nodes found
+  .panel-body
+    p You must spawn your cluster before bootstrapping it. Your nodes will automatically appear when they are running.
```
MaximilianMeister: MINOR: I'd rather say sth like "once they are up and running"

vitoravelino: Will change. Content review are always more than welcome. Thanks!

vitoravelino:
> the additional logic with any_minion? is a really good solution :+1: didn't think about that

We should avoid any logic/query in the view and try using helpers to abstract that. Will work on this after 1.0.

app/helpers/application_helper.rb
```diff
@@ -7,4 +7,8 @@ def user_image_tag(owner)
     email = owner.nil? ? nil : owner.email
     gravatar_image_tag(email)
   end
+
+  def any_minion?
+    !Minion.all.count.zero?
```
mssola: This could be simplified to: `Minion.any?` 😉 In which case, is it really needed this helper then ?

vitoravelino: Didn't know about `any?` on model. I always prefer to expose a helper/variable instead of classes on templates but that's my preference. 

It would be nice to have explicit code standards for stuff that is not automated linted. I know we have rubocop but it doesn't analyze template files. I found https://github.com/flyerhzm/rails_best_practices, might help with some tweaks? Need to investigate.

mssola: That's something worth-investigating indeed. Then yeah, simply expose `Minion.any?` through this helper.

PR #209, opened 2017-06-23
'Reboot to update' after click behavior

vitoravelino:
After clicking "Reboot to update", fixed wrong button text and locked modal until admin node is rebooted.

bsc#1045380, bsc#1045730

peek 2017-06-26 18-41

- [X] Implementation
- [X] Update specs
- [x] GIF

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -282,7 +282,6 @@ MinionPoller = {
   }
 };
 
-
 $('body').on('click', '.accept-minion', function(e) {
   var $link = $(this);
   var selector = ''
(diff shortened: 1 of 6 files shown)


vitoravelino:
I decided to simplify a bit the loading part suggested by @flavio adding a loading icon to the rebooting button.

Also, I wasn't able to automate testing (:pensive:) because I'm not able to mock browser requests and we don't have js testing yet. I found https://github.com/oesmith/puffing-billy#setup-for-capybara that would do the trick but it seems to exist an issue with phantomjs. If someone has an idea, drop it here and I'll check it out.

app/controllers/updates_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
       arg:     "systemctl reboot"
     )
 
-    redirect_to root_path, flash: { info: "Rebooting..." }
+    render json: { status: "rebooting" }
```
mssola: Nitpicking: Maybe hide this status in an `enum` ?

app/assets/javascripts/dashboard.js
```diff
@@ -302,32 +301,75 @@ $('body').on('click', '.accept-minion', function(e) {
...
 
+$('.update-admin-modal').on('hide.bs.modal', function(e) {
+  var isRebooting = $('.update-admin-modal').data('rebooting');
+
+  if (isRebooting) {
+    e.preventDefault();
+  }
+});
+
+// unlock moda, now closable
```
mssola: Typo: modal*

app/assets/javascripts/dashboard.js
```diff
@@ -302,32 +301,75 @@ $('body').on('click', '.accept-minion', function(e) {
...
+  var $modal = $('.update-admin-modal');
+  var $btn = $modal.find('.btn');
+
+  $modal.data('rebooting', false);
+  $modal.find('.close').show();
+  $btn.prop('disabled', false);
+}
+
+// lock modal, won't close
+function lockUpdateAdminModal() {
```
mssola: I think we can merge this function with the one above.

kiall:
@vitoravelino given concourse won't show me this jobs logs, but it failed after 34 seconds, I'm assuming this needs to just be forced pushed.. Can you do that? I'd like to merge this ASAP.

PR #210, opened 2017-06-26
Add support for running Puma with SSL

ereslibre:
Puma will be executed using SSL. If the port provided is 80 (we will
be serving autoyast with no safe transport in 1.0), then we only
create the autoyast route, since we should not let the customers handle
Velum through an insecure connection.

Connect to the salt-api using TLS.

Remove certificate details from the UI and related code.

Fixes: bsc#1043570

Depends on:

- https://github.com/kubic-project/salt/pull/136
- https://github.com/kubic-project/caasp-container-manifests/pull/67


config/routes.rb
```diff
@@ -1,33 +1,40 @@
 # frozen_string_literal: true
+
+# rubocop:disable Metrics/BlockLength
 Rails.application.routes.draw do
-  devise_for :users, controllers: { registrations: "auth/registrations",
-                                    sessions:      "auth/sessions" }
```
MaximilianMeister: as we need that same route in both cases we could also write:
```ruby
Rails.application.routes.draw do
  get "/autoyast", to: "dashboard#autoyast"
  next if Rails.env.production? && ENV["VELUM_PORT"].to_i != 443
  ...
end
```

config/puma.rb
```diff
@@ -42,3 +42,11 @@
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
+
+pidfile "/tmp/puma.pid"
```
MaximilianMeister: would it make sense to keep that pid file in Rails's tmp dir?
we also have puma's statefile, i'd save it in the same place ideally. it can be set via `state_path`

ereslibre: If we save it in the rails tmp dir an unclean exit of the container on the development environment will fail to start again, and regarding production it doesn't affect, I'll leave it at `/tmp`...

spec/controllers/setup_controller_spec.rb
```diff
@@ -243,25 +238,14 @@
       end
 
       it "warns and redirects to the setup_path" do
-        put :configure, settings: settings_params.each { |k, _v| settings_params[k] = "" }
+        put :configure, settings: Hash[settings_params.map { |k, _| [k, ""] }]
```
MaximilianMeister: there's another line down below for json formatted requests that needs to be adapted as well

spec/models/pillar_spec.rb
```diff
@@ -6,71 +6,14 @@
...
       {
-        apiserver:    "something",
-        city:         "Nuremberg",
-        company_name: "SUSE Linux GmbH",
-        company_unit: "Research and development",
-        country:      "DE",
-        dashboard:    "something",
-        email:        "containers@suse.de",
-        state:        "Bavaria"
+        dashboard: "something",
```
MaximilianMeister: even though it's not technically important, and it has been like that before, i'd rather keep some real world values here that make sense (IP address or fqdn)

config/puma.rb
MaximilianMeister: makes sense, but to ensure that you need to configure the state_path as well. i don't know where it will be stored by default, but i think it's not in `/tmp`

spec/controllers/setup_controller_spec.rb
ereslibre: Thanks!

spec/models/pillar_spec.rb
ereslibre: Done!

config/puma.rb
ereslibre: Done!

spec/vcr_cassettes/kubeconfig/cluster_not_ready.yml
```diff
@@ -57,7 +57,7 @@ http_interactions:
     uri: http://127.0.0.1:8000/
```
ukbelch: Shouldn't this be TLS?

spec/vcr_cassettes/kubeconfig/cluster_ready.yml
```diff
@@ -57,7 +57,7 @@ http_interactions:
     uri: http://127.0.0.1:8000/
```
ukbelch: Shouldn't this be TLS?

bin/init
```diff
@@ -2,7 +2,6 @@
 
 # This script will setup the database and then start the rails application
 
-set -x
 set -e
 
 : ${VELUM_PORT:=80}
```
ukbelch: Shouldn't this be 443?

bin/init
```diff
@@ -62,4 +61,4 @@ setup_database() {
 
 setup_database
 bundle exec rake db:migrate
-bundle exec "rails s -b 0.0.0.0 -p $VELUM_PORT --pid /tmp/puma.pid Puma"
```
ukbelch: Binding on all interfaces is generally not recommended (0.0.0.0), is there a way to explicitly bind to the public interface?

config/puma.rb
```diff
@@ -42,3 +42,12 @@
...
+state_path "/tmp/puma.state"
+
+if ENV["VELUM_PORT"].to_i == 443
+  ssl_bind "0.0.0.0", ENV["VELUM_PORT"], key: "/etc/pki/velum.key", cert: "/etc/pki/velum.crt"
```
kiall: Re 0.0.0.0 - I think, in this particular instance, we should be OK with this (caveat below..). This specific bind is intended to be world accessible, so 0.0.0.0 doesn't bind us to any networks we consider less secure than we intended to bind to.

Caveat: We should be OK with this, assuming it's nontrivial to reliably find the correct bind address here, as we gain TLS.

In our case, this is non-trivial to find. Given a machine with 2 NICs, we don't know which one to expose the public interfaces on. We could assume the first NIC, but this will undoubtedly be wrong in many scenarios. What we would need to do, is ask at install time what NICs should be used for which purposes, but that itself relies on a better network architecture.

app/validators/alphanumeric_validator.rb
```diff
@@ -1,34 +0,0 @@
...
-  def validate_each(record, attribute, value)
-    # Base case: irrelevant pillar value, or nil value.
-    return unless relevant?(record.pillar, value)
-
-    # Pillar values that should contain only alphabetical digits.
-    return if contains?(ALPHABETICAL_VALUES, record.pillar) && value.match(/^[a-zA-Z\s]+$/)
-
-    # Pillar values that can contain characters from "space" to "~" in the
-    # ASCII table.
-    return if contains?(ALPHANUMERICAL_VALUES, record.pillar) && value.match(/^[\x20-\x7e]+$/)
```
kiall: Heh, luckily, all this code is being deleted in this pull request, so that issue is gone once it merges.

config/puma.rb
mssola: @ereslibre for what it's worth, take a look at what we are doing in Portus: https://github.com/mssola/Portus/blob/2061d5d1e78bc20f362863c4699b6d18e854c00d/config/puma.rb. That being said, I'm not sure how the `pidfile` is useful inside of a container ...

app/views/setup/worker_bootstrap.html.slim
```diff
@@ -14,19 +14,19 @@ p
...
   |  Bootstrap all the nodes you want to make part of this platform by adding the following boot parameter
-  |  <strong>autoyast=#{autoyast_url host: @controller_node, port: ENV["VELUM_PORT"]}</strong>
+  |  <strong>autoyast=#{autoyast_url host: @controller_node, protocol: :http}</strong>
 
 h2 Tips
 p
   | You don't need to boot each node by hand. More information on how to embed an AutoYaST profile in your PXE environment is available&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#bootmedium.pxe" here
-  | . Where <strong>#{autoyast_url host: @controller_node, port: ENV["VELUM_PORT"]}</strong> is the real, generated path to the AutoYaST profile served by the dashboard.
+  | . Where <strong>#{autoyast_url host: @controller_node, protocol: :http}</strong> is the real, generated path to the AutoYaST profile served by the dashboard.
```
ereslibre: @flavio @eotchi: could we get some text to be added on how to use the autoyast file with an external storage?

config/puma.rb
ereslibre: I think in this case we can just stick to `0.0.0.0`, I agree with @kiall that this can be improved, but we can do this at a later time. It's not ideal, but when we change this to a specific interface we should also consider generate the velum certificate with the chosen IP address only.

spec/vcr_cassettes/kubeconfig/cluster_not_ready.yml
ereslibre: These are VCR cassettes that reproduce a previous recorded HTTP session. Unfortunately I couldn't find a way for them to support a TLS connection. Since this is for testing purposes I'd suggest revisit it during july or any other time.

spec/vcr_cassettes/kubeconfig/cluster_ready.yml
ereslibre: Ditto.

bin/init
ereslibre: Yes! Thanks! (We are never falling back to the default, but with this changes the default should be `443`).

bin/init
ereslibre: This is the removal of the previous command that we were using, we can keep the discussion on the previous `0.0.0.0` bind thread :)

app/validators/alphanumeric_validator.rb
ereslibre: Yup, this code is going away with this PR.

PI-Victor:
[Bug 1043570 - Velum – Content served over an insecure connection (HTTP)](https://trello.com/c/tBHbY9VU/98-bug-1043570-velum-content-served-over-an-insecure-connection-http)

PI-Victor:
[Bug 1043589 - Saltstack – API runs with SSL disabled](https://trello.com/c/4hGDBEMF/104-bug-1043589-saltstack-api-runs-with-ssl-disabled)

config/country_codes.yml
```diff
@@ -1,254 +0,0 @@
...
-      "VU": Vanuatu
-      "VE": Venezuela, Bolivarian Republic of
-      "VN": Viet Nam
-      "VG": Virgin Islands, British
-      "VI": Virgin Islands, U.S.
-      "WF": Wallis and Futuna
-      "EH": Western Sahara
-      "YE": Yemen
-      "ZM": Zambia
-      "ZW": Zimbabwe
```
jordimassaguerpla: Why are you removing all these country codes? You sure we dont' use them anywhere else?

ereslibre: They were used for the country selector which we won't have. It's on the git history, so it's not lost, but I'd try to keep the working copy as clean as possible, so we don't end up with zillions of unused things that nobody knows for what they were and what's their purpose :)

mssola: We are going to handle certificate generation in an entirely different way, so we no longer need all the certificate setup stuff on velum.

kiall:
Force merged 3 TLS PRs that go together 

https://github.com/kubic-project/velum/pull/210
https://github.com/kubic-project/salt/pull/136
https://github.com/kubic-project/caasp-container-manifests/pull/67

PR #212, opened 2017-06-27
fix empty settings check

MaximilianMeister:


app/controllers/setup_controller.rb
```diff
@@ -108,4 +107,9 @@ def check_empty_bootstrap
       format.json { render json: msg, status: :unprocessable_entity }
     end
   end
+
+  def valid_settings?
+    required_pillars = (Pillar.all_pillars.keys - Pillar::OPTIONAL_PILLARS).map(&:to_s)
+    (required_pillars - settings_params.keys).empty?
```
ereslibre: I don't think this is checking what we expect, we want to check that no Pillar value with this set of keys is blank...

MaximilianMeister:
fix was included in https://github.com/kubic-project/velum/pull/210

PR #213, opened 2017-06-27
ui: polishing and minor enhancements

vitoravelino:
Most of changes were done based on [wireframes](https://projects.invisionapp.com/share/PJBM8UUB2#/screens/236598224):

- General
  - Added missing Roboto Medium (used on panel titles for example)
  - Slightly larger line-height on `.panel-heading` to fit `.btn-sm`
  - Buttons on `.panel-heading` moved from `.btn-xs` to `.btn-sm`
- Configuration
  - Added missing `.panel-title` (except for proxy settings to avoid conflict)
- Overview page
  - Centered status icon to column name
  - Status text in the same column as icon
  - Aligned table data text in the middle
  - Wrong use of th instead of td on nodes table
  - Master checkbox aligned before role text
  - Changed update all nodes from button to link
- Overview and Discovery
  - Added text when Pending Nodes panel is empty

## Overview (before)

screenshot from 2017-06-28 13-04-03

## Overview (after)

screenshot from 2017-06-28 13-00-01

## Discovery page (before)

screenshot from 2017-06-28 13-05-19

## Discovery page (after)

screenshot from 2017-06-28 12-31-26

PR #215, opened 2017-06-28
Clarification on the discovery page

mssola:


mssola:
It's not enough that nodes are up and running, but they also need to be accepted on the "Pending nodes"  table that is located below.

vitoravelino:
@mssola I'm going to add some text when the pending nodes panel is empty too as you suggested [here](https://github.com/kubic-project/velum/pull/206#issuecomment-310637836) on https://github.com/kubic-project/velum/pull/213.

flavio:
LGTM

mssola:
My forced push was simply a rebase, so the old approvals also applied.

PR #218, opened 2017-06-29
Increase the poll timeout to 30s

grahamhayes:
Currently velum makes a lot of calls to the salt-api which get
logged as events

This is causing MySQL and the Velum event processor to get
swampped, which is causing delays in showing node status.

30 seconds is a good number as it allows for information to be
updated, and if a user is impatient, they can refresh the page.

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -132,7 +132,7 @@ MinionPoller = {
       }
     }).always(function() {
       // make another request only after the last one finished
-      setTimeout(MinionPoller.request, 5000);
+      setTimeout(MinionPoller.request, 30000);
     });
   },
 


vitoravelino:
If the change is reasonable then the change is good to go.

mssola:
Looks like we need some more "wait"'s here and there. Otherwise, you could also raise the `config.default_max_wait_time` from 5 to something bigger like 10 or even 20 in the `spec/support/capybara.rb` file.

vitoravelino:
> Looks like we need some more "wait"'s here and there. Otherwise, you could also raise the config.default_max_wait_time from 5 to something bigger like 10 or even 20 in the spec/support/capybara.rb file.

You are right, totally forgot that. This will break the specs. It would be nice to be configurable and we could adjust this on specs with some variable.

grahamhayes:
OK, even updating the `config.default_max_wait_time` to 30s still causes the tests to fail - I am not sure what needs to be done here.

PR #219, opened 2017-06-29
[WIP] Stop calling all nodes for grains

grahamhayes:
We were targeting all nodes when we did a `grains.get` api call
which meant we caused the master to call all minions to tell them
to call the master to find out what it thought the grains were,
and then return the same list of data n times, where n == number of
minions.


grahamhayes:
Unfortunately, it looks like this is the only way to pull grains out of the salt-api :/

PR #220, opened 2017-06-29
bootstrap: added titles to checkboxes and radio (bsc#1046035)

vitoravelino:
screenshot 2017-06-30 10 32 56
screenshot 2017-06-30 10 33 04
screenshot 2017-06-30 10 33 16

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -250,7 +250,7 @@ MinionPoller = {
     } else {
       masterHtml = '<td class="text-center">\
         <input name="roles[master][]" id="roles_master_' + minion.id +
-        '" value="' + minion.id + '" type="radio" ' + masterChecked + '></td>';
+        '" value="' + minion.id + '" type="radio" title="Select node as master" ' + masterChecked + '></td>';
     }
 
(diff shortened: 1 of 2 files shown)


vitoravelino:
Why is this failing miserably? It's a silly change and it doesn't contain the ssl code. It shouldn't fail like that.:(

Anyne?

ereslibre:
@vitoravelino Try adding https://github.com/kubic-project/velum/pull/221 :)

vitoravelino:
But this branch was checked out before the ssl thing was merged. The `next if` is not present in [routes.rb](https://github.com/vitoravelino/velum/blob/add-checkbox-title/config/routes.rb). Am I missing something?

ereslibre:
This is really weird. I can say the errors shown on Travis matches the
error with the `next` in the route (without my PR), but you are right that
your branch doesn't contain this `next` in the route. Try to rebase it to
`master` and add the mentioned PR... Does it locally fail like this to you
with the development environment?

vitoravelino:
I didn't rebase devenv locally yet and it passes. Travis doesn't use devenv, right? So it shouldn't affect it either. We'll discover tomorrow. :)

mssola:
Restarting the build. I think this is related with #221.

vitoravelino:
Just rebased. Thanks for the comment on bugzilla.

PR #222, opened 2017-06-29
fix up feature spec use of Capybara

twalpole:
This fixes up some of the feature specs use of Capybara, and also fixes a number of incidences where the specs weren't actually doing anything since

    expect do
       # stuff in here gets ignored
    end

won't actually evaluate the block being passed to expect without calling .to/.not_to ...  at the end

The specs currently fail, but they're failing in master at the moment too.

### DIFF
--- a/spec/features/bootstrap_cluster_feature_spec.rb
+++ b/spec/features/bootstrap_cluster_feature_spec.rb
@@ -23,7 +23,7 @@ feature "Bootstrap cluster feature" do
 
     before do
       # mock salt methods
-      [:minion, :master].each do |role|
+      [:worker, :master].each do |role|
         allow_any_instance_of(Velum::SaltMinion).to receive(:assign_role).with(role)
           .and_return(role)
       end
@@ -61,14 +61,12 @@ feature "Bootstrap cluster feature" do
       click_button "Proceed anyway"
 
       using_wait_time 10 do
-        expect do
-          # means it went to the overview page
-          page.to have_content("Summary")
-          page.to have_content(minions[0].fqdn)
-          page.to have_content(minions[1].fqdn)
-          page.not_to have_content(minions[2].fqdn)
-          page.not_to have_content(minions[3].fqdn)
-        end
+        # means it went to the overview page
+        expect(page).to have_content("Summary")
+        expect(page).to have_content(minions[0].fqdn)
+        expect(page).to have_content(minions[1].fqdn)
+        expect(page).not_to have_content(minions[2].fqdn)
+        expect(page).not_to have_content(minions[3].fqdn)
       end
     end
 
@@ -85,14 +83,12 @@ feature "Bootstrap cluster feature" do
       click_on_when_enabled "#bootstrap"
 
       using_wait_time 10 do
-        expect do
-          # means it went to the overview page
-          page.to have_content("Summary")
-          page.to have_content(minions[0].fqdn)
-          page.to have_content(minions[1].fqdn)
-          page.not_to have_content(minions[2].fqdn)
-          page.not_to have_content(minions[3].fqdn)
-        end
+        # means it went to the overview page
+        expect(page).to have_content("Summary")
+        expect(page).to have_content(minions[0].fqdn)
+        expect(page).to have_content(minions[1].fqdn)
+        expect(page).to have_content(minions[2].fqdn)
+        expect(page).not_to have_content(minions[3].fqdn)
       end
     end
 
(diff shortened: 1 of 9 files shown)


vitoravelino:
Huge thanks, @twalpole! We currently have master broken until #221 gets merged apparently. I guess all of the `expect ... do` were my code, so I completely misunderstood that.

I'll be opening a PR to your branch soon with the code fixes so you can include them in this PR (you can squash it later, I don't mind), ok?

vitoravelino:
@twalpole Just opened https://github.com/twalpole/velum/pull/1. Please merge it and squash with your commit and we are good to go.

Huge thanks again for this!

vitoravelino:
Rubocop is complaining:

```
spec/features/bootstrap_cluster_feature_spec.rb:48:5: C: Example has too many expectations [6/2]
    scenario "An user bootstraps anyway a cluster with only 2 minions", js: true do
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/features/bootstrap_cluster_feature_spec.rb:73:5: C: Example has too many expectations [5/2]
    scenario "An user selects a subset of nodes to be bootstraped", js: true do
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/features/bootstrap_cluster_feature_spec.rb:95:5: C: Example has too many expectations [5/2]
    scenario "An user check all nodes at once to be bootstraped", js: true do
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/features/bootstrap_cluster_feature_spec.rb:115:3: C: Example has too many expectations [4/2]
  scenario "It shows the minions as soon as they register", js: true do
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Forgot that. Could you fix that? :sweat_smile:

twalpole:
@vitoravelino That ruboocop  requirement is pretty ridiculous for feature specs

vitoravelino:
I agree with you. It's one of the topics I'm planning to discuss with the rest of the team. It's pretty annoying having to add comments to disable some rules. 

You don't need to reduce the number of expects just wrap the scenarios like

``` ruby
# rubocop:disable RSpec/ExampleLength
... scenario
# rubocop:enable RSpec/ExampleLength
```

Sorry for this. :(

twalpole:
I just disabled that rule for all feature specs - it makes no sense in features

spec/features/bootstrap_cluster_feature_spec.rb
```diff
@@ -85,14 +83,12 @@
...
-          page.to have_content(minions[0].fqdn)
-          page.to have_content(minions[1].fqdn)
-          page.not_to have_content(minions[2].fqdn)
-          page.not_to have_content(minions[3].fqdn)
-        end
+        # means it went to the overview page
+        expect(page).to have_content("Summary")
+        expect(page).to have_content(minions[0].fqdn)
+        expect(page).to have_content(minions[1].fqdn)
+        expect(page).to have_content(minions[2].fqdn)
```
twalpole: @vitoravelino This swapped from .not_to to .to, not sure if that was my change or your changes I merged in.  Do you know which it's supposed to be?  (as mentioned previously it wasn't actually being evaluated before)

vitoravelino: I did that change. It was a typo from the previous scenario. It's right. I ran the tests and it passes. Don't worry!

vitoravelino:
> I just disabled that rule for all feature specs - it makes no sense in features

I see no problem with that. Hopefully nobody won't either.

Since you disabled for feature specs, would you mind to remove the `# rubocop:disable RSpec/ExampleLength` and `# rubocop:enable RSpec/ExampleLength` present on the rest of files?

twalpole:
@vitoravelino  I didn't disable the ExampleLength check for features - just the MultipleExpectations check -- Do you want the ExampleLength check disabled for features too?

vitoravelino:
> @vitoravelino I didn't disable the ExampleLength check for features - just the MultipleExpectations check -- Do you want the ExampleLength check disabled for features too?

Oops! My bad, you are right. It's MultipleExpectations.

twalpole:
@vitoravelino I have already moved the MultipleExpectations pragmas from the feature files.  As far as ExampleLength removal, It's up to you -- it's not raising any warnings currently, although feature specs can validly be pretty long depending on the feature being tested -- It probably does make sense to exclude features from that requirement, but it doesn't need to be this PR -- willing to remove it here if you want though.

vitoravelino:
> @vitoravelino I have already removed the MultipleExpectations comments from the feature files. As far as ExampleLength removal, It's up to you -- it's not raising any warnings currently, although feature specs can validly be pretty long depending on the feature being tested -- It probably does make sense to exclude features from that requirement, but it doesn't need to be this PR -- willing to remove it here if you want though.

I agree that sometimes you can end up with longer scenarios but it's not something very common (at least not yet). So, I guess we can leave it for now.

It's getting late here so I gotta go... Thanks again for this! Appreciated it!

vitoravelino:
@twalpole Could you rebase against master? Travis should be green after that.

twalpole:
@vitoravelino Done, it had one failure though.  I'm not familiar enough with this app/code to know whether it's just that it could take more than 10 seconds for the result to show up? If not, it's most likely the "Proceed anyway" is being clicked before the modal has fully animated (and the click actually misses the button, can happen with Poltergeist due to the way the event queue works).  If it's the latter  the solution is to either wait for the modal to have a class that indicates animation is done (if there is one), or (better yet) disable all animations in the test environment (which has the happy side effect of speeding up tests)

vitoravelino:
@twalpole You are right and that's fine. Occasionally happens because travis vms are too slow, so don't worry. Everything looks great!

vitoravelino:
@flavio, @mssola 

Can we force merge this since it doesn't touch e2e and e2e tests are not yet stable apparently?

PR #223, opened 2017-06-30
Fix caching for salt `update_status` call

ereslibre:
Do not use the cache at all if `cache: false`, cache for 30 seconds
by default, and use the cache if `cache: true`.

Related: bsc#1046847

### DIFF
--- a/lib/velum/cache.rb
+++ b/lib/velum/cache.rb
@@ -1,0 +1,14 @@
+module Velum
+  # This class has cache utilities that can be used for the whole project
+  class Cache
+    def self.fetch(key:, use_cache: true, expires_in:, &blk)
+      if use_cache
+        Rails.cache.fetch(key, expires_in: expires_in) do
+          blk.call
+        end
+      else
+        blk.call
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


mssola:
I don't really like this approach. I'd rather add a helper:

```
def rails_cache_fetch(key, cached, &blk)
    if cached
        Rails.cache.fetch(key, expires_in: 30.seconds) do
          yield blk
        end
    else
      yield blk
    end
end
```

And then use this method with the Salt.call inside of the block (similarly to what we are doing now). I think this is cleaner.

ereslibre:
@mssola agree, thanks! Please check again the syntax. I will update the tests.

mssola:
Syntax looks alright. I'm going to test it locally, just in case there's something funny with the caching side that went unnoticed...

flavio:
I haven't tested it yet, but the code looks good to me

grahamhayes:
This is not caching for me - the salt-api is still getting the requests every 5 seconds

flavio:
I tested with the devenv and the ISO @ereslibre built last evening. The deployment worked fine, but `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.event pretty=True` shows that the salt-api is still being hit.

Caching doesn't seem to be working here as well

ereslibre:
Caching by default in rails is only enabled in production, that might explain why is not caching for you using the dev. env.

ereslibre:
Superseded by https://github.com/kubic-project/velum/pull/229

PR #225, opened 2017-06-30
Remove explicit set for `pidfile` and `statefile`

ereslibre:


mssola:
Also, I don't think these files are relevant when running inside of a container, no?

ereslibre:
> ...and this makes me wonder if we need a pidfile at all for the devenv. we might just drop pidfile and state_path altogether.

The idea was to avoid it creating it on the mounted project hierarchy (making new runs fail because the pidfile is there yet if you did an unclean exist -- frequently on development).

However, you are right, I cannot see the `pidfile` any longer, so I'm updating the PR to directly remove those directives.

ereslibre:
Requesting for reviews again.

ereslibre:
Approved by @MaximilianMeister previously, minor change, tests passing, and security related. Merging.

PR #226, opened 2017-07-03
Remove `restart.txt` puma functionality

ereslibre:
Puma allows you to restart the server based on the `mtime` of a file
`tmp/restart.txt`. We won't be using this functionality as the server
runs in a container and we have other means for restarting it.

This makes e2e-tests stable again.


ereslibre:
Closing, this has no effect at all.

PR #227, opened 2017-07-03
Fixed more flaky tests

mssola:

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -20,8 +20,7 @@ module Helpers
 
   # click on `selector element when enabled
   def click_on_when_enabled(selector)
-    # will wait until it becomes enabled
-    find("#{selector}:not([disabled])").click
+    find("#{selector}:not([disabled])", match: :first).click
   end
 end
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ rvm:
   - 2.1.9
 
   # Other ruby versions
-  - 2.4.1
-  - 2.3.4
   - 2.2.7
+  - 2.3.4
+  - 2.4.1
 
   # Latest ruby
   - ruby-head
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,8 +1,15 @@
 MinionPoller = {
   selectedNodes: [],
+  pollingTimeoutId: null,
 
   poll: function() {
     this.request();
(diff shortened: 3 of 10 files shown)


vitoravelino:
On average I guess the tests are way faster but I'm still not sure about the stability of rebooting expectation and click_when_enabled helper. 

I agree on reverting the click when enabled change of but not the rebooting text.

mssola:
@vitoravelino I think this PR is fine as it is now. It's way faster, and unit tests passed. I'll trigger more rebuilds, but I think it should good to go. We can improve tests in another PR if you want, but the top priority was to fix the flaky tests...

vitoravelino:
I re-triggered a few times yesterday, the first two had 1 failure and the other 3 passed. I've just re-triggered. It's not so flaky, eventually with 1 or 2 re-triggers pass. If someone feels annoyed we can get back to this.

spec/support/helpers.rb
```diff
@@ -20,8 +20,14 @@ def disabled?(selector)
...
   def click_on_when_enabled(selector)
-    # will wait until it becomes enabled
-    find("#{selector}:not([disabled])", match: :first).click
+    wait_until { have_css("#{selector}:not([disabled])") }
+    find(selector, match: :first).click
+  end
+
+  def wait_until_having_content(text)
+    has_content = false
+    wait_until { has_content = have_content(text) }
```
mssola: Are you sure this doesn't return always true ?

vitoravelino: You mean the `wait_until`?

vitoravelino:
After I added the wait_until, duration increased a lot. I ran a few times and not sure if it's just a coincidence. :(

spec/support/helpers.rb
mssola: The block inside of it.

vitoravelino: Yeah, you are right. Need to use the `page.has_text?` as before. Totally forgot.

vitoravelino:
@mssola any other suggestion? If not, feel free to squash everything.

mssola:
This approach is taking us nowhere. We'll start re-visiting how we mock the backend, since maybe it's stuck on some heavy-lifting from that side...

mssola:
Rebased on top of master to gain all the changes on the CI

mssola:
Pushed the branch into the upstream repository so it can be picked up by another PR.

PR #231, opened 2017-07-04
Redirect from HTTP to HTTPS

mssola:
This will not be applied to some paths like /_health or /autoyast.
Moreover, I haven't added tests for this because doing so is a can of
worms for Capybara (the existing tests), and it's not worth it.

Also note that the proper fix should come from a reverse-proxy side. We
need this for other features as well, but it would bring too much
complexity for the 1.0 release.

Fixes bsc#1047065


app/controllers/application_controller.rb
```diff
@@ -19,4 +20,11 @@ def redirect_to_setup
   def setup_done?
     !Minion.assigned_role.count.zero?
   end
+
+  # TODO: (mssola) remove once we have a reverse-proxy setup in place.
+  def redirect_to_secure
+    return if params[:action] == "autoyast" || params[:controller] == "health"
```
ereslibre: I think you can modify the `before_action` so you can do: `before_action :redirect_to_secure, unless: [:autoyast, :health]`

ereslibre: It's weird to see in the application controller refer to the `action`, or even the `controller`... why not have a unconditional `before_action :redirect_to_secure` and on the dashboard controller: `skip_before_action :redirect_to_secure, only: :autoyast` and on the health controller `skip_before_action :redirect_to_secure` ?

mssola: That makes much more sense. Thanks!

mssola: Done!

flavio:
Why did you remove `app/views/shared/_aside.html.slim`, how is that related with this PR?

mssola:
> Why did you remove app/views/shared/_aside.html.slim, how is that related with this PR?

It's not related, and I should probably move that into another PR. I simply noticed this unused file while working on this 😅

flavio:
Tested locally with devenv.

The code works fine but there's a limited time frame during the first boot in which strange things happen:

  * I got an error from rails because the migrations were not done
  * Once the migrations were done I saw the login page of velum, I was even able to press the "sign up" button
  * Reloading the page lead me got me automatically redirected to the https site, which at that time was not yet ready

This can be reproduced by opening http://localhost, firing up the devenv and pressing F5.

I honestly doubt we can fix some of the issues reported here without introducing yet another entrypoint for the container. As we all know the easiest fix would be to put apache in front of a single container...

ereslibre:
@mssola Is there a reason for `e2e-tests` failing here? Could you please `push -f` so it can be retriggered?

PR #232, opened 2017-07-04
Serve "db is not ready" static page over https [bsc#1047203]

flavio:
Ensure the "db is not yet ready" is served over https too. This fixes bsc#1047203.

The fix has been tested with devenv. Everything works fine until velum comes up, then velum will be served over port 3000 but with tls disabled (this is fixed by https://github.com/kubic-project/caasp-devenv/pull/23).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -48,3 +48,7 @@ AllCops:
     - db/schema.rb
     - db/migrate/*
     - vendor/**/*
+    # This is a patched version of ruby's core un.rb file. Upstream has
+    # a particular style that conflicts with our rules -> let's ignore
+    # this file
+    - bin/httpsd.rb
--- a/bin/httpsd.rb
+++ b/bin/httpsd.rb
@@ -1,0 +1,44 @@
+#!/usr/bin/env ruby
+
+# This is a patched version of the `httpd` function defined inside of ruby's
+# stdlib (see the `un.rb` file shipped with your ruby distribution).
+# The upstream version cannot serve contents over https.
+# A patch is going to be sent upstream, in the meantime this code does the job.
+
+def httpd
+  require "un"
+  setup("", "BindAddress=ADDR", "Port=PORT", "MaxClients=NUM", "TempDir=DIR",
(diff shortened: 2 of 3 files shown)


flavio:
Upstream "pull request": https://bugs.ruby-lang.org/issues/13714

kiall:
Should we serve both 80 and 443 during this phase? It looks like we can just start both up, stash both pids, and kill both at the end.

Unrelated - and defiantly not for this patch! but, should the `bundle exec rake db:migrate` call move to beside the `bundle exec rake db:create` call - so there is a smaller window where nothing is being served?

flavio:
> Should we serve both 80 and 443 during this phase? It looks like we can just start both up, stash both pids, and kill both at the end.

The situation is a little more complicated than expected. Right now we have two containers serving velum: one on port 443 (`velum-dashboard`) and the other on port 80 (`velum-autoyast`).

The `bin/init` script touched by this PR is the entrypoint of the `velum-dashboard` container. The `velum-autoyast` one starts straight with `puma` serving the web application.

The `velum-autoyast` container is serving only the `GET /autoyast` endpoint. No other routes are specified in that case. @mssola is currently working on [bsc#1047065](https://bugzilla.suse.com/show_bug.cgi?id=1047065) which is about having this container redirect http requests to https (except of course of `GET /autoyast`). Once this is in place the "db is not ready" page will be visible both from http and https.

flavio:
> Unrelated - and defiantly not for this patch! but, should the bundle exec rake db:migrate call move to beside the bundle exec rake db:create call - so there is a smaller window where nothing is being served?

I think the `db:migrate` task is left deliberately out of the `setup_database()` to ensure the db of an existing velum instance is migrated after an update of the velum image.

flavio:
Question: this is more aesthetic probably, but would you like me to change the code to enable the usage of ssl certificate and key only when `VELUM_PORT` is `443`?

flavio:
> That being said, while pressing "refresh", there was a time in which the server was not available anymore, but I think this is a known issue... (it's the time when the DB is available, so we shut down the temporary server and start the real one)

Yep, that's it. Merging in the meantime to get it into RC2

PR #233, opened 2017-07-05
Also restrict unsafe routes on development

ereslibre:
We will only allow autoyast being served in an insecure manner, so bring
development close to production here too.

Only when running specs under a test environment is when we avoid restricting
the routes, however this would also need to be reviewed.


ereslibre:
Closing, https://github.com/kubic-project/velum/pull/231 will properly fix this!

PR #239, opened 2017-07-05
views: update node bootstrapping instructions with https value.

PI-Victor:
i believe, with the introduction of https in autoyast, ~this is no longer served over http~.

### DIFF
--- a/app/views/setup/worker_bootstrap.html.slim
+++ b/app/views/setup/worker_bootstrap.html.slim
@@ -14,19 +14,19 @@ p
   |  This process leverages AutoYaST and is (almost) fully automated.
   |  In case you are not familiar with it, you can find more information about AutoYaST in the&nbsp;
   a href="https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html" official documentation.
-  |  The automatic installation gets invoked by adding <strong>autoyast=#{autoyast_url host: @controller_node, protocol: :http}</strong> to the kernel parameter list.
+  |  The automatic installation gets invoked by adding <strong>autoyast=#{autoyast_url host: @controller_node, protocol: :https}</strong> to the kernel parameter list.
   |  If you aren't under a PXE environment you can also use <strong>netsetup=dhcp</strong> kernel parameter for the network to be automatically configured using a reachable DHCP server.
   |  As installation media, you can use the very same image you bootstrapped the admin node with.
 
   |  A ready to use AutoYaST profile has already been generated for you during the bootstrap of the admin node.
   |  Bootstrap all the nodes you want to make part of this platform by adding the following boot parameter
-  |  <strong>autoyast=#{autoyast_url host: @controller_node, protocol: :http}</strong>
+  |  <strong>autoyast=#{autoyast_url host: @controller_node, protocol: :https}</strong>
 
 h2 Tips
 p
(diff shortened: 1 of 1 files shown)


ereslibre:
Unfortunately we cannot do this, since the CA is autogenerated and the workers when using autoyast won't be able to verify the identity of `autoyast=https://server/autoyast` because they don't know the CA and they don't trust it, making the installation fail, we still need `http` for the autoyast profile serving.

PI-Victor:
i'm... unsure what i'm missing, i totally understand the scenario, but how come i've used autoyast=https://whatever/autoyast to manually bootstrap clusters for the last few days, and that works just great? with the latest ISO, i mean?

ereslibre:
@PI-Victor :astonished: I just tried it and **it works**. To be honest, I never even tried it because I've always assumed it was going to fail... if this works and is supposed to work we can completely remove puma listening on port 80 on `caasp-container-manifests` too...

Also, if this is the case, completely removing `protocol: :https` is the best option, as the protocol will be just inferred from the current environment (you'll be using `https` while on the dashboard as a user).

@flavio can you please provide some information about autoyast working with `https` with an untrusted chain? Is this supposed to work? Could we remove autoyast serving on port 80?

ereslibre:
@PI-Victor Now I'm wondering... does it work because I passed `insecure=1` so this check is also ignored?

PI-Victor:
i don't pass anything besides netsetup and autoyast parameters.
i also believe the work on autoyast for https was done by @flavio or @kiall.
#1043570 has some details in @kiall's comment no.9

ereslibre:
Yes, but still the CA cannot be trusted on the workers when using autoyast (they cannot know if the CA used to create the `velum` cert can be trusted or not).

However I just tried it with the latest product image [build 393] (only passing `netsetup` and `autoyast`) and I can confirm it works... If this is supposed to work (and not something expected to be fixed on YaST side), I think we can remove autoyast from port 80 completely...

:astonished:

flavio:
A couple of clarifications: the autoyast file is still served over http, we have a container that serves it up. The solution is a temporary fix until we put a reverse proxy in front of a single velum container.

The fact autoyast is not complaining about the file grabbed from the https endpoint is really strange. I'll ask around to the YaST folks.

PI-Victor:
i think it's related to: https://github.com/yast/yast-packager/pull/164/files

flavio:
I'm not so sure it's related because you didn't provide this `ssl_verify=no` param at boot time...

PI-Victor:
indeed. i thought we might do that automatically or something, but i guess it should turn up in the kernel boot params section.

PI-Victor:
so...
a) do i close this PR or is it valid?
b) do i open a bugzilla issue about autoyast not complaining about https over self-generated certs without parameters being passed?

flavio:
Sorry, yesterday I got distracted by other bugs I didn't come back to this one. I would not merge the PR, I wanted to replicate the issue and file a bug against the YaST developers on bugzilla.

So please go ahead with filing the bug on bugzilla and, once you have the number, add it here as a comment and close the PR.

I'll find some time to reproduce that locally, but I don't want to block you further.

flavio:
@PI-Victor I can reproduce the issue locally 😨 I even disable the http container to ensure there was no fallback from https -> http. I wonder why this happens, please report the issue to the YaST devs.

PI-Victor:
great. created bsc#1047697

PR #243, opened 2017-07-06
Take orchestration result into consideration when bootstrapping

ereslibre:
When bootstrapping, if the orchestration fails, depending in what stage
it failed, we might never retrieve individual highstate returns from minions,
so they will be in `pending` status forever.

With this change, we also listen for the overall bootstrap orchestration
result (`orch.kubernetes` for now), so we will mark all pending nodes as
failed.

Fixes: bsc#1046706


ereslibre:
Tests are passing, I am going to validate the behavior forcing a real world example.

ereslibre:
This is how I tested it works: I ran a successful orchestration from the very beginning. Then, I ran:

`docker exec -it $(docker ps | grep velum-dashboard | awk '{print $1}') entrypoint.sh bundle exec "rails runner 'Minion.update_all highstate: Minion.highstates[:pending]'"`

so the minion status on the UI were spinning again (pending highstate). I then applied: https://gist.github.com/ereslibre/682f4daca06a03bfdab9bd5a73e154e3

And ran:

`docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.kubernetes`

I can confirm both minions were green, as both reached highstate, and the error was a late orchestration error.

Then, I again forced them to be in pending status:

`docker exec -it $(docker ps | grep velum-dashboard | awk '{print $1}') entrypoint.sh bundle exec "rails runner 'Minion.update_all highstate: Minion.highstates[:pending]'"`

And applied this patch instead: https://gist.github.com/ereslibre/768606ccced1338ea4119a92152ac6b3, which forces an early failure, before highstates have been reached on the minions.

I can confirm both minions were in failed status (red) after running the orchestration with:

`docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.kubernetes`

app/models/salt_handler/minion_highstate.rb
```diff
@@ -17,8 +17,6 @@ def initialize(salt_event)
     @salt_event = salt_event
   end
 
-  # This method is responsible for all actions needed when a minion starts.
-  # For now we simply store the Minion in our database.
```
mssola: Don't simply remove the documentation, update it 😉

kiall:
Merging, passes tests and had 2x +A before some code comments were added

PR #237, opened 2017-07-05
Improve the style of "db is initializing" page [bsc#1047305]

flavio:
Try to use the same style as the main velum web application. This is something I hacked together in a couple of minutes. I think @vitoravelino will have lots of comments about that :)

## Before

before

## After

after

PR #245, opened 2017-07-07
adapt style of static pages - follow up #237

MaximilianMeister:
we need to keep all the static pages in sync with css styles:

static


MaximilianMeister:
merged, as this needs to go on top of the latest commit (#237) in the product

MaximilianMeister:
we actually had a bug for that https://bugzilla.suse.com/show_bug.cgi?id=1045571

PR #248, opened 2017-07-10
Add missing indexes for `salt_events` table

ereslibre:
Fix some slow queries that we were having due to missing indexes.

Increase batch job size.

Fixes: bsc#1046862

### DIFF
--- a/app/models/salt_event.rb
+++ b/app/models/salt_event.rb
@@ -57,7 +57,7 @@ class SaltEvent < ApplicationRecord
     return false if taken_event.zero?
 
     SaltEvent.where(worker_id: worker_id, processed_at: nil)
-             .where.not(taken_at: nil).limit(1).first.process!
+             .where.not(taken_at: nil).limit(1).each(&:process!)
   end
 
   # Memoizes the @handler instance which is responsible to process this event.
--- a/app/models/salt_handler/minion_start.rb
+++ b/app/models/salt_handler/minion_start.rb
@@ -26,6 +26,7 @@ class SaltHandler::MinionStart
     return false if id == "ca" || id == "admin"
 
     minion_info = Velum::Salt.minions[id]
+    return false if minion_info.blank?
 
     # false if a minion with this minion_id or fqdn already exists (uniqueness validation)
     Minion.new(minion_id: id, fqdn: minion_info["fqdn"]).save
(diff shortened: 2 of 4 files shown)


ereslibre:
Please, note that this isn't a formal benchmark, just some numbers to have a rough idea. The patch contains two different changes: one is adding the indexes, the other is increasing the batch size for fetching jobs. No improvement on `processed_at` has been done, but that could also be batched.

Also note, that since some events (like new minion event) require calls to the salt-api, the `salt_events` table gets increased when this calls happens (I didn't comment this calls from our event processors). It's a very small amount of events, but still, increasing on each run.

## With indexes only

Imported a big data set from production with a big cluster provided by @kiall. Updated the `salt_events` table:

```
MariaDB [velum_development]> update salt_events set taken_at=null,processed_at=null,worker_id=null;
Query OK, 115089 rows affected (9.37 sec)
Rows matched: 115089  Changed: 115089  Warnings: 0
```

The salt event processor started:

`Salt event processor worker1 started at 2017-07-10 09:38:52 UTC`

After the last event updated, taking new jobs (no more jobs to take):

```
SQL (0.2ms)  UPDATE `salt_events` SET `salt_events`.`worker_id` = 'worker1', `salt_events`.`taken_at` = '2017-07-10 09:51:50' WHERE `salt_events`.`processed_at` IS NULL AND (worker_id IS NULL OR (worker_id = 'worker1') OR taken_at < '2017-07-10 09:46:50.566216') LIMIT 1
```

So it finished at `09:46:50` having started at `09:38:52`. Roughly 8 minutes.

## With indexes and read batch increased

```
MariaDB [velum_development]> update salt_events set taken_at=null,processed_at=null,worker_id=null;
Query OK, 115928 rows affected (10.24 sec)
Rows matched: 115929  Changed: 115928  Warnings: 0
```

The salt event processor started:

`Salt event processor worker1 started at 2017-07-10 09:57:09 UTC`

After the last event updated, taking new jobs (no more jobs to take):

```
SQL (0.2ms)  UPDATE `salt_events` SET `salt_events`.`worker_id` = 'worker1', `salt_events`.`taken_at` = '2017-07-10 10:03:09' WHERE `salt_events`.`processed_at` IS NULL AND (worker_id IS NULL OR (worker_id = 'worker1') OR taken_at < '2017-07-10 09:58:09.508617') LIMIT 1000
```

So it finished at `10:03:09` having started at `09:57:09`. Roughly 6 minutes.

ereslibre:
Update: batch removed from the patch for now.

ereslibre:
Interesting read: https://dev.mysql.com/doc/refman/5.7/en/multiple-column-indexes.html

PR #249, opened 2017-07-10
Handle failures of the update orchestration

kiall:
Handle the failure of an update orchestration in the same way we handle
a failure for a bootstrap orchestration.

bsc#1047895

### DIFF
--- a/app/models/salt_handler/minion_orchestration.rb
+++ b/app/models/salt_handler/minion_orchestration.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 # This class is responsible to handle the salt events that are orchestration results.
-# Only the bootstrap orchestration will be handled by this class.
-class SaltHandler::BootstrapOrchestration
+# Only the bootstrap and update orchestrations will be handled by this class.
+class SaltHandler::MinionOrchestration
   attr_reader :salt_event
 
   TAG_MATCHER = %r{salt/run/\d+/ret}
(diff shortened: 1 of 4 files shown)


kiall:
I'm not sure if this is the right change, there is a "retryable update" state which may be more appropriate?

ereslibre:
> I'm not sure if this is the right change, there is a "retryable update" state which may be more appropriate?

It looks fine to me, what do you mean with the "retryable update" state?

kiall:
Theres a "Update Failed - Retryable" state, I just don't see how to "trigger" it ;)

https://github.com/kubic-project/velum/blob/master/app/assets/javascripts/dashboard.js#L223

ereslibre:
> Theres a "Update Failed - Retryable" state, I just don't see how to "trigger" it ;)

Oh, that should be automatic. When it has the pending update grain available, and the status is failed, that's what the UI should show.

kiall:
So, this shouldn't merge yet. I've spotted an issue with it, + the original fix for the main orch

ereslibre:
Rebased on top of master, I'm still validating it. Some things got in between when I was validating it.

app/models/salt_handler/minion_orchestration.rb
```diff
@@ -12,16 +12,18 @@ def self.can_handle_event?(event)
     parsed_event_data = JSON.parse event.data
     parsed_event_data["fun"] == "runner.state.orchestrate" &&
       (parsed_event_data["fun_args"].first == "orch.kubernetes" ||
-       parsed_event_data["fun_args"].first["mods"] == "orch.kubernetes")
+       parsed_event_data["fun_args"].first == "orch.update" ||
+       parsed_event_data["fun_args"].first["mods"] == "orch.kubernetes" ||
+       parsed_event_data["fun_args"].first["mods"] == "orch.update")
```
mssola: To avoid repetition, maybe you can create a helper method that checks "fun_args". So you can call it:

```ruby
salt_args_match(parsed_event_data, "orch.kubernetes", "orch.update")

...

def salt_args_match(data, *values)
  d = data.first
  return if d.nil?

  values.each { |v| return true if d == v || d["mods"] == v }
  false
end
```

mssola: You could even check whether `data["fun_args"]` is nil beforehand, just to extra-safe 😉

mssola: You could also go a bit beyond and allow this method to also check that ` parsed_event_data["fun"] == "runner.state.orchestrate"`. So the method call would cover "is this an orchestration result we care about?"

app/models/salt_handler/minion_orchestration.rb
```diff
@@ -11,17 +11,26 @@ def self.can_handle_event?(event)
...
+  end
+
+  def self.salt_fun_args_match(parsed_event_data, *orchestrations)
+    fun_args = parsed_event_data["fun_args"]
+
+    orchestrations.each do |o|
+      return true if fun_args.first == o || fun_args.first["mods"] == o
+    end
+
+    false
```
ereslibre: This whole loop and `return true/false` can be replaced with a one liner: `orchestrations.any? { |o| fun_args.first == o || fun_args.first["mods"] == o }`

ereslibre:
Minor comment, testing for approval.

kiall:
Simple rebase to retrigger concourse, had approvals from @ereslibre and @mssola - will merge once it passes.

kiall:
Backport to 1.0 @ https://github.com/kubic-project/velum/pull/259

Issue #228, opened 2017-07-04
spinner on 'Accept Node'?


PI-Victor:
I know it's feature freeze, but i was wondering if we can add like a spinner for `Accept Node`. 
![giphy](https://user-images.githubusercontent.com/6294120/27822278-66d5d090-60a5-11e7-9f18-a6c3ab2b4dca.gif)

Mostly because it takes some time for the node to move to `nodes found` and it doesn't let me know if it's doing something. I'm just looking at it and i have the impulse to keep clicking `Accept node`.
I think it would be better to have some feedback, even if it's a short operation.

vitoravelino:
There's already a [bug](https://bugzilla.suse.com/show_bug.cgi?id=1046663) regarding this. I didn't work on it yet because there are still some unknowns on how to solve the slowness issues we are currently having with salt and stuff.

PR #250, opened 2017-07-11
bsc#1046663: added indicator when user accepts node

vitoravelino:
After user accepts one or more nodes, 'Accept Node' link
is replaced by 'Acceptance in progress' text. The specific
minion is marked in `sessionStorage` and even if the user
refreshes the page, the 'Accept Node' won't appear until
it gets accepted.

### None accepted

screenshot 2017-07-10 20 51 13

### Partially accepted

screenshot 2017-07-10 20 33 36

### All accepted

screenshot 2017-07-10 20 33 43

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -98,8 +98,9 @@ MinionPoller = {
         }
 
         // Show / Hide the table depending the presence of pending nodes
-        $(".pending-nodes-container #accept-all").prop('disabled', pendingMinions.length === 0)
-        $(".pending-nodes-container .table").toggleClass('hidden', pendingMinions.length === 0);
+        var acceptLinks = $('.pending-nodes-container td > a');
+        $(".pending-nodes-container #accept-all").prop('disabled', acceptLinks.length === 0)
+        $(".pending-nodes-container .has-content").toggleClass('hidden', pendingMinions.length === 0);
         $(".pending-nodes-container .empty-text").toggleClass('hidden', pendingMinions.length > 0);
 
         // show/hide panels on discovery page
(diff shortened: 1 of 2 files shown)


vitoravelino:
> It works as explained. Something I noticed (and then in your screenshots as well), is that when you accept a node, the column somehow changes: it gets shifted to the left 😕 (I guess that's why it's marked as WIP 😉)

It wasn't because of that but I'll fix that since you came up with. I wanted to review the code by myself before letting others review.

vitoravelino:
One thing to consider is... I'm using `sessionStorage` so the `Acceptance in progress` label will remain over refreshes/restores of that specific tab. If user opens a new tab, `Accept Node` and `Accept all nodes` will be available. Is that okay?

To be honest, this is more like a workaround since there's no `in progress`state coming from the api. Since we are still unsure about how to fetch these kind of data, I guess we are fine for now. Also, I didn't consider `localStorage` because there might be an error while accepting a node and that would block the user from trying to accept a node again.

vitoravelino:
Also, the changes seems to be bigger compared to yesterday's code because I've decided to break the `.accept-minion` callback in smaller pieces to become easier to be reviewed by other people and further refactoring.

mssola:
> One thing to consider is... I'm using sessionStorage so the Acceptance in progress label will remain over refreshes/restores of that specific tab. If user opens a new tab, Accept Node and Accept all nodes will be available. Is that okay?

I think so. This is something that we can re-visit once we change how we handle this.

PI-Victor:
will test and merge in 10 min.

PI-Victor:
tested and LGTM: fixes: https://github.com/kubic-project/velum/issues/228

PR #254, opened 2017-07-12
`force_ssl` is documented to be executed from the class context

ereslibre:
Not all redirections to `https` were properly handled. By making this
change we manually redirect to `https` as a `before_action` as before,
but handling a `redirect_to` in the chain (effectively stopping the
chain and in the expected order).

The reason behind this is that `force_ssl` called from our own filter:
https://github.com/rails/rails/blob/03925dc26a747075ff37660d1f0a060a8178bf66/actionpack/lib/action_controller/metal/force_ssl.rb#L62-L68
installs a `before_action`, but it's already too late for that very
request (we are already in the `before_actions` for that request).

Also, with the current implementation the filter gets installed after
our other `before_action` filters, what implies that the redirection
to `https` is not the very first filter in the chain, as it should
be.

By calling to the method itself:
https://github.com/rails/rails/blob/03925dc26a747075ff37660d1f0a060a8178bf66/actionpack/lib/action_controller/metal/force_ssl.rb#L76-L96
from our filter, we ensure all requests are redirected to TLS.

Fixes: bsc#1048134

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,7 +2,7 @@
 
 # ApplicationController is the superclass of all controllers.
 class ApplicationController < ActionController::Base
-  before_action :redirect_to_secure
+  before_action :force_ssl_redirect, unless: -> { Rails.env.test? }
   before_action :authenticate_user!
   before_action :redirect_to_setup
   protect_from_forgery with: :exception
@@ -19,11 +19,5 @@ class ApplicationController < ActionController::Base
   # return false otherwise
   def setup_done?
     !Minion.assigned_role.count.zero?
-  end
-
-  # TODO: (mssola) remove once we have a reverse-proxy setup in place.
-  def redirect_to_secure
-    self.class.force_ssl unless Rails.env.test?
-    true
   end
 end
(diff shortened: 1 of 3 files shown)


ereslibre:
This code will be removed from  here in the short term, as this will be properly handled by a reverse proxy and rails won't need to care about this at all. This is a short term solution.

ereslibre:
Way to reproduce the issue. Start a fresh devenv, when it's up, run:

```
~ curl http://localhost/setup/discovery
<html><body>You are being <a href="http://localhost/users/sign_in">redirected</a>.</body></html>
```

This is not what's expected, as the first redirection should always be to https, we'd expect that redirect to be: https://localhost/setup/discovery

In any case, if we follow the proposed redirection:

```
~ curl http://localhost/users/sign_in
<!DOCTYPE html><html><head><title>SUSE CaaSP: Velum</title>...
```

I can properly see the sign in page, which is not supposed to happen at all.

With this PR:

```
~ curl http://localhost/setup/discovery
<html><body>You are being <a href="https://localhost/setup/discovery">redirected</a>.</body></html>
```

Following the proposed redirection:

```
~ curl -k https://localhost/setup/discovery
<html><body>You are being <a href="https://localhost/users/sign_in">redirected</a>.</body></html>
```

PI-Victor:
testing this right nao as well.

PI-Victor:
tested and i can confirm this works as expected, i now get redirected, properly, from http to https on all endpoints i can access in the dashboard. no more ruby errors.

PR #255, opened 2017-07-13
Detect orchestration results when `mods` has been provided as argument

ereslibre:
We usually run the orchestrations manually as:
`salt-run state.orchestrate orch.kubernetes`

And this case was detected properly. This change adds support for also
detecting the orchestration result when the orchestration is called using
the `salt-api` on Velum (the general case):
`salt-run state.orchestrate mods=orch.kubernetes`

With this change we support both return results.

### DIFF
--- a/app/models/salt_handler/bootstrap_orchestration.rb
+++ b/app/models/salt_handler/bootstrap_orchestration.rb
@@ -11,7 +11,8 @@ class SaltHandler::BootstrapOrchestration
     return false unless event.tag =~ TAG_MATCHER
     parsed_event_data = JSON.parse event.data
     parsed_event_data["fun"] == "runner.state.orchestrate" &&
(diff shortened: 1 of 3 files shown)


ereslibre:
How I tested this: follow the instructions at https://github.com/kubic-project/velum/pull/243#issuecomment-313443754, but instead of running:

```
docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.kubernetes
```

Run:

```
docker exec -it $(docker ps | grep velum-dashboard | awk '{print $1}') entrypoint.sh bundle exec "rails runner \"require 'velum/salt'; Velum::Salt.orchestrate\""
```

This will mimic a call to the orchestration using the `salt-api` backend implementation in Velum.

ereslibre:
I can confirm this works as expected with the instructions above: removing `WIP` flag.

PR #258, opened 2017-07-17
Simplify tests using contexts and simplify orchestration detection

ereslibre:


app/models/salt_handler/minion_orchestration.rb
```diff
@@ -16,12 +16,7 @@ def self.can_handle_event?(event)
 
   def self.salt_fun_args_match(parsed_event_data, *orchestrations)
```
mssola: From the previous PR I miss a comment explaining what this method does.

mssola:
I haven't tested it, but it should be equivalent.

PR #261, opened 2017-07-19
bsc#1045734: fix show/hide update link based on update/highstate

vitoravelino:
`Update all nodes` link should be shown if all nodes are in applied
state and there's an update available/retryable. If for some reason a node
goes to a different state than "applied" or there's no update,
link becomes unavailable.

PR #262, opened 2017-07-20
bsc#1049165 : avoid orchestration action from ui while updating/adding node

vitoravelino:
- [x] When an admin node update is available, do not allow users to press "update all nodes". The admin node should be updated first.
- [x] Never allow both a workers update and an admin update to be started at the same time.
- [x] Velum should not allow add node while a bootstrap/upgrade/etc is in progress
- [X] Hide "(new)" link to adding unassigned nodes if upgrade is in progress 

Depends on: #261

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -42,6 +42,7 @@ MinionPoller = {
         var pendingRendered = "";
         var allApplied = true;
         var updateAvailable = false;
+        var hasPendingStateNode = false;
         var updateAvailableNodeCount = 0;
 
         // In discovery, the minions to be rendered are unassigned, while on the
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/dashboard.js
```diff
@@ -127,7 +138,7 @@ MinionPoller = {
           // overview page otherwise
           if ($("#node-count").length > 0) {
             $("#node-count").text(unassignedMinions.length + " nodes found");
-          } else {
+          } else if (!hasPendingStateNode) {
```
grahamhayes: This causes the number of nodes to be hidden, not just the link.
screenshot from 2017-07-20 18-26-25

mssola:
Works like a charm, good job 👏

PR #267, opened 2017-07-21
Fix loopback IP for proxy exception during initial configuration

cwickert:
There is a typo in the pre-configured proxy exceptions during initial config: 127.0.01 instead of 127.0.0.1.  Note the missing dot.


vitoravelino:
Nice catch! Thanks!

PR #269, opened 2017-07-25
allow kubeconfig download when master is ready (bsc#1048483)

MaximilianMeister:
~~still need to test it~~

tested and works as expected

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -40,7 +40,7 @@ MinionPoller = {
       success: function(data) {
         var rendered = "";
         var pendingRendered = "";
-        var allApplied = true;
+        var masterApplied = false;
         var updateAvailable = false;
         var hasPendingStateNode = false;
         var updateAvailableNodeCount = 0;
@@ -79,8 +79,8 @@ MinionPoller = {
         for (i = 0; i < minions.length; i++) {
           rendered += MinionPoller[renderMethod].call(MinionPoller, minions[i]);
 
-          if (minions[i].highstate != "applied") {
-            allApplied = false;
+          if (minions[i].role == "master" && minions[i].highstate == "applied") {
+            masterApplied = true;
           }
 
           if (minions[i].highstate === "pending") {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/dashboard.js
```diff
@@ -125,7 +130,7 @@ MinionPoller = {
         var hasAdminNodeUpdate = data.admin.update_status === 1 || data.admin.update_status === 2;
         $("#update-all-nodes").toggleClass('hidden', !updateAvailable || hasAdminNodeUpdate);
 
-        MinionPoller.enable_kubeconfig(minions.length > 0 && allApplied);
+        MinionPoller.enable_kubeconfig(minions.length > 0 && masterApplied);
```
ereslibre: If `allApplied` and `masterApplied` started with the safest default value (`false`), and then we only set them to `true` inside their special cases if they are met, then this line would just be `MinionPoller.enable_kubeconfig(masterApplied);`

MaximilianMeister:
@vitoravelino @ereslibre fixed. I also moved the `if master` condition into the existing `if state applied` block

app/assets/javascripts/dashboard.js
```diff
@@ -79,8 +80,11 @@ MinionPoller = {
         for (i = 0; i < minions.length; i++) {
           rendered += MinionPoller[renderMethod].call(MinionPoller, minions[i]);
 
-          if (minions[i].highstate != "applied") {
-            allApplied = false;
+          if (minions[i].highstate == "applied") {
+            allApplied = true;
```
ereslibre: Ouch, this will set `allApplied` to `true` with only one applied, is this intended? Maybe we can use `Array.prototype.every()` to set `allApplied`

MaximilianMeister: thanks for the hint. can you check again? not sure if that's fully correct or optimal

MaximilianMeister: i just realized we don't even use allApplied anymore... i think we can just drop it

PR #270, opened 2017-07-25
Serve the autoyast content even when a SCC/SMT connection error happens

ereslibre:
The service unavailable http error code (503) will only be returned now
until the first page of the setup is completed (so we have the dashboard
address to serve the autoyast profile).

When one of the following exceptions happens:
  - Missing registration code
  - Missing credentials for SCC/SMT
  - Any HTTP error while connecting to SCC/SMT

We will still serve the autoyast content with a comment explaining why
the system cannot be registered. The returned http error code will be
success (200), but the XML will contain a comment stating why the system
cannot be registered.

Fixes: bsc#1049948

### DIFF
--- a/lib/velum/suse_connect.rb
+++ b/lib/velum/suse_connect.rb
@@ -6,11 +6,23 @@ module Velum
   # This class handles interaction with the SUSE Connect service (whether is the SCC or SMT).
   class SUSEConnect
     # Raised when there is a connection exception with the SMT/SCC service
-    class SCCConnectionException < StandardError; end
+    class SCCConnectionException < StandardError
+      def initialize
+        super "Error connecting to SCC/SMT"
+      end
+    end
     # Raised when an active registration code for CaaSP is missing
-    class MissingRegCodeException < StandardError; end
+    class MissingRegCodeException < StandardError
+      def initialize
+        super "Missing registration code"
+      end
+    end
     # Raised when no credentials were found for SCC service
-    class MissingCredentialsException < StandardError; end
+    class MissingCredentialsException < StandardError
+      def initialize
+        super "Missing credentials for SCC/SMT"
+      end
+    end
 
     DEFAULT_SMT_URL = "https://scc.suse.com".freeze
 
@@ -138,8 +150,8 @@ module Velum
       else
         return response, nil
       end
-    rescue *HTTPExceptions::EXCEPTIONS => e
-      raise SCCConnectionException, e
+    rescue *HTTPExceptions::EXCEPTIONS
+      raise SCCConnectionException
     end
   end
 end
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -43,7 +43,9 @@ class DashboardController < ApplicationController
         @suse_regcode = suse_connect_config.regcode
         @do_registration = true
       rescue Velum::SUSEConnect::MissingRegCodeException,
-             Velum::SUSEConnect::MissingCredentialsException
+             Velum::SUSEConnect::MissingCredentialsException,
+             Velum::SUSEConnect::SCCConnectionException => e
+        @registration_error = e.to_s
         @do_registration = false
       end
       ssh_key_file = "/var/lib/misc/ssh-public-key/id_rsa.pub"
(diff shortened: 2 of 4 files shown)


ereslibre:
Backport for `1.0/update` here: https://github.com/kubic-project/velum/pull/271

ereslibre:
Instead of a comment on the XML I'm trying something fancier with `ask-list` on the autoyast profile. Putting it back to `WIP`.

ereslibre:
Updated PR. This message is shown when using autoyast when the system is not being registered. The message depends on the error:

screenshot from 2017-07-25 17-20-49

ereslibre:
I thought this way is a nicer way to provide feedback than a comment in the XML file, however this doesn't allow an unattended installation (if anything goes south with the registration, which might be good actually).

Please provide your feedback to know if this is a feasible solution or we should be silent during the whole installation process when using autoyast without mattering the error we are facing.

robdaemon:
I think this looks pretty great. I can't really test it (I haven't built an ISO yet myself!) but I think it looks good.

Maybe AJ should sign off on it too, since it's user-facing?

ereslibre:
@robdaemon Just as a side note, I launched the `devenv` and used an existing ISO (in this case I didn't build the ISO myself), so if you raise a VM with qemu you can still point it with `netsetup=dhcp autoyast=http://your-ip-address/autoyast` as we are only testing that it fetches the error from the XML served from your `devenv` in this case :)

ereslibre:
Adding @flavio because it's user facing, who we should ask too as @robdaemon says? @ajaeger, @eotchi?

ereslibre:
This will also very probably need adapting the QA suite tests.

lib/velum/suse_connect.rb
```diff
@@ -78,6 +79,17 @@ def regcode
                         smt_insecure: smt_insecure,
                         credentials:  credentials).regcode
       end
+
+      def human_error_for_exception(exception)
```
MaximilianMeister: instead of abstracting the messages outside of the error class itself into a helper method, why not add the messages to the class itself, that we define above?

```ruby
class MissingRegCodeException < StandardError
  def initialize(msg="Missing registration code")
    super
  end
end
```

ereslibre: Updated but in a bit different style, as we don't really expect `msg` to be settable in this case.

thkukuk:
Yes, and this warning is completly wrong and needs to be removed immediately again. Not that this breaks autoinstallation (even a timeout of 30 seconds is not Ok), for 99% of the installations you don't need a registration code, because customers normally use SMT or SUSE Manager.

ereslibre:
> Yes, and this warning is completly wrong and needs to be removed immediately again. Not that this breaks autoinstallation (even a timeout of 30 seconds is not Ok), for 99% of the installations you don't need a registration code, because customers normally use SMT or SUSE Manager.

This message will only pop up if:

- There was a problem connecting to SCC or SMT when the autoyast profile is going to be served (so the worker won't be registered).
- No regcode was provided during the admin installation (so there's no way to register new workers against SCC), and no SMT is being used.
- Credentials for SCC/SMT are missing (API), this should not happen under normal circumstances.

If this message is being shown with SMT it's a bug, it shouldn't happen.

thkukuk:
How does velum know what my installation source is? That's impossible to know for velum, since this is an installer option and not visible to velum.

ereslibre:
> How does velum know what my installation source is? That's impossible to know for velum, since this is an installer option and not visible to velum.

It certainly can, done here: https://github.com/kubic-project/velum/blob/7625b36d2a34f9b100d6800c4fad9cbc57e0e8ea/lib/velum/suse_connect.rb#L58-L60

All our docker containers have the secrets patch, which includes:

```
linux-jbau:~ # docker exec -it 9a6 bash
bash-4.3# cat /var/run/secrets/SUSEConnect 
---
insecure: false
url: https://scc.suse.com
```

If `url` is `https://scc.suse.com` then it's SCC, otherwise it's SMT.

thkukuk:
What you are seeing there is how the admin node was installed, not what the installation source for the cluster nodes will be.
QA, Rob and me run into this, so this are already three ways of installation where velum is wrong with the guess.
I use the DVD for the admin node and PXE boot with autoyast via SMT. I'm not interested in running around with a DVD for installation, but I'm also not interested in writing an autoyast profile only for the admin node. Everything is configured in the PXE/tftpboot config files on my install server. Velum has no access to this and is not able to find out what the installation source for the cluster nodes is and what is needed.

flavio:
I spoke with @ereslibre, we are going to revert this PR.

PR #271, opened 2017-07-25
Serve the autoyast content even when a SCC/SMT connection error happens

ereslibre:
The service unavailable http error code (503) will only be returned now
until the first page of the setup is completed (so we have the dashboard
address to serve the autoyast profile).

When one of the following exceptions happens:
  - Missing registration code
  - Missing credentials for SCC/SMT
  - Any HTTP error while connecting to SCC/SMT

We will still serve the autoyast content with a comment explaining why
the system cannot be registered. The returned http error code will be
success (200), but the XML will contain a comment stating why the system
cannot be registered.

Fixes: bsc#1049948

Backport of https://github.com/kubic-project/velum/pull/270


ereslibre:
Marking as `WIP` since this is a `1.0/update` candidate. If we believe is not, we can just close this one.

PR #272, opened 2017-07-25
Set VERSION to 1.0.0+dev

kiall:


flavio:
I think it should be `1.0.0+dev` - it would be cool to have the commit SHA into it

MaximilianMeister:
> it would be cool to have the commit SHA into it

that requires some adaptions in our overall packaging automation (meaning other repos as well) but it's a good idea. i'll create a card for that

kiall:
Marking as WIP due to flavios comment, will update to 1.0.0+dev

PR #274, opened 2017-07-26
set secure flag in cookie (bsc#1050484)

MaximilianMeister:

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: "_velum_session"
+Rails.application.config.session_store :cookie_store,
+  key:    "_velum_session",
+  secure: !Rails.env.test?


config/environments/production.rb
```diff
@@ -43,7 +43,7 @@
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
ereslibre: This needs not to be set on `test` environment, otherwise tests will fail.

ereslibre:
We cannot set `config.force_ssl` globally as the autoyast profile needs to be served over `http` for now too (as well as the health controller), so this setting will not take that into account. We are already redirecting to `https` since https://github.com/kubic-project/velum/commit/79cccb5d36ed3b196c230686bc4fa8c0dee20562.

Isn't it enough to just set `secure` on the cookie? After all we are not going to have any sessions over `http`.

We can set `STS` headers on our own after every request using the application controller I think by setting `response.headers["Strict-Transport-Security"] = 'max-age=0'`.

andreasstieger:
If the session is exclusive to HTTPS, and there are remaining HTTP non-session endpoints, flagging the session cookie as secure would be the correct fix.

ereslibre:
> If the session is exclusive to HTTPS, and there are remaining HTTP non-session endpoints, flagging the session cookie as secure would be the correct fix.

Please, correct me if I'm wrong, but there's still an attack vector despite /autoyast and the health controller are session-less (and served over http). If I login through `https` and then manually visit from the same browser the `autoyast` endpoint over http (could also be visited over `https`, but let's do it with `http`), the cookie is exposed.

So I think HSTS is also necessary, not sure if it will affect `autoyast` itself and the profile will be correctly loaded.

andreasstieger:
Compliant user agents will sent cookies received with `secure=1` only on HTTPS requests. So to protect the session cookie, this is sufficient. HSTS goes further as to indicate to user agents that they should expect all resources, including e.g. the login page, to be on HTTPS. But that is an extended scenario.

ereslibre:
> Compliant user agents will sent cookies received with secure=1 only on HTTPS requests. So to protect the session cookie, this is sufficient. HSTS goes further as to indicate to user agents that they should expect all resources, including e.g. the login page, to be on HTTPS. But that is an extended scenario.

Of course, sorry. I was trying an attack vector and forgot to set the secure cookie, so of course I had it :man_facepalming: 

Thank you!

config/initializers/session_store.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: "_velum_session"
+Rails.application.config.session_store :cookie_store,
+  key:    "_velum_session",
+  secure: Rails.env.production?
```
grahamhayes: Why are we only setting this in production?

ereslibre: We can also set it on development, but not on the test environment as tests are going to fail, maybe we can change it to `secure: !Rails.env.test?` ?

MaximilianMeister: mainly because that's the only place where it matters

grahamhayes: We should stop trying to make dev/test/prod environments different.

Why would setting ^ cause test environments to fail? Do they still access non https endpoints?

MaximilianMeister: i didn't know for sure that the failure is related to adding this header, at least the test error doesn't make sense in that regard, as all other tests are passing

ereslibre: > We should stop trying to make dev/test/prod environments different.

They are different by nature. Rails will reload classes on `development`, while it won't on `production`, logging is also different as well as lots of other small (and big) details. We can try to converge in this special case, but we just cannot make dev/test/prod environments to be the same, mainly because they are not meant to be the same.

> Why would setting ^ cause test environments to fail? Do they still access non https endpoints?

Tests are completely run with `http`, no `https` at all: https://github.com/kubic-project/velum/blob/927bc4972aa969ea505f3631b2a3cd2b5806ae46/app/controllers/application_controller.rb#L5. Also, calls to salt-api: https://github.com/kubic-project/velum/blob/927bc4972aa969ea505f3631b2a3cd2b5806ae46/lib/velum/salt_api.rb#L82-L86

I haven't looked in detail why tests fail, but if we enable `https` they would for sure be orders of magnitude slower, do we want them to be slower for controller testing over `https`? Don't we have other tests for that? I'm wondering here, open to discussion. I'd expect this tests to be as fast as possible, so you can run them several times while you are hacking on something, I don't want tests to be slow because they are using `https` in an environment not meant for safety but speed.

I really believe it's not a matter of having all environments in the same way, but to have the right tool with the right settings for the job.

My 2 cents.

ereslibre: Also, not sure at all VCR cassettes can use SSL/TLS.

MaximilianMeister: i also think we don't need to have enhanced security on dev/test environments, especially if it slows down the testing.

and i think this error is not related to the cookie setting, it's flakiness because locally, and in 2/3 ruby versions on travis everything is green

grahamhayes: OK - then for this case we should ignore the test environment.

My concern about leaving this off for dev is that this is how we do work on velum, and we may not see an error while doing other development that will only show in a production environment.

The environments are different - but we should limit the differences to only what is needed for each environment type.

ereslibre: > The environments are different - but we should limit the differences to only what is needed for each environment type.

I totally agree with this statement.

MaximilianMeister: i'll update the PR, this will also retrigger travis

PR #278, opened 2017-08-01
Allow multiple master selection

kiall:
Allow the user to choose multiple masters during the initial
bootstrap.

We don't currently allow for adding a master post bootstrap.

Depends-On: https://github.com/kubic-project/automation/pull/40
Depends-On: https://github.com/kubic-project/automation/pull/60

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -295,24 +295,24 @@ MinionPoller = {
     minionHtml = '<input name="roles[worker][]" id="roles_worker_' + minion.id +
       '" value="' + minion.id + '" type="checkbox" title="Select node for bootstrapping" ' + minionChecked + '>';
 
-    var roleHtml = `
-      <td class="role-column">
-        ${masterHtml}
-        ${minionHtml}
-        <div class="btn-group role-btn-group">
-          <button type="button" class="btn btn-default ${isMaster ? 'btn-primary' : ''}" data-minion-id="${minion.id}" data-minion-role="master">Master</button>
-          <button type="button" class="btn btn-default worker-btn ${isWorker ? 'btn-primary' : ''}" data-minion-id="${minion.id}" data-minion-role="worker">Worker</button>
-          <button type="button" class="btn btn-default unused-btn ${isUnused ? 'btn-primary' : ''}" data-minion-id="${minion.id}">Unused</button>
-        </div>
-      </td>
-    `;
+    var roleHtml = '\
+      <td class="role-column">' +
+        masterHtml +
+        minionHtml +
+        '<div class="btn-group role-btn-group">\
+          <button type="button" class="btn btn-default master-btn '+ (isMaster ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="master">Master</button>\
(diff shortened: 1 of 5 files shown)


kiall:
The UX here really isn't great, I'd like to see someone more familiar with HTML/JS etc update the UI to change over to use a dropdown / select on the right hand side to choose between "Unused" / "Master" / "Worker", and the column of checkboxes on the left be removed entirely.

app/assets/javascripts/dashboard.js
```diff
@@ -538,26 +557,29 @@ $('body').on('change', 'input[name="roles[worker][]"]', function() {
...
-    var $checkbox = $(this).closest('tr').find('input[name="roles[worker][]"]');
+    MinionPoller.selectedMasters.push(value);
 
     $checkbox.prop('checked', true);
     $checkbox.prop('disabled', true);
+  } else {
+    var index = MinionPoller.selectedMasters.indexOf(value);
+    MinionPoller.selectedMasters.splice(index, 1);
+
+    $checkbox.prop('disabled', false);
```
vitoravelino: We need to maintain the previous behavior of unchecking the worker/node checkbox if it wasn't previously checked before selecting the node as master. Renaming the variable to `$workerCheckbox` would also help the code understanding.

Scenario 1:

1. No nodes selected yet
1. Select a node as master
1. Unselect as master the node from previous step 
1. Worker checkbox should not be selected

Scenario 2:

1. A node is selected to bootstrap
1. Select a node as master
1. Unselect as master the node from previous step 
1. Worker checkbox should be kept as checked

We can write a feature test for this if it doesn't exist yet for further refactoring.

vitoravelino:
> The UX here really isn't great, I'd like to see someone more familiar with HTML/JS etc update the UI to change over to use a dropdown / select on the right hand side to choose between "Unused" / "Master" / "Worker", and the column of checkboxes on the left be removed entirely.

Sounds like a good idea. I can help you with this. @mssola, what do you think?

If we go with this approach, you can ignore the comment I made previously.

vitoravelino:
> The UX here really isn't great, I'd like to see someone more familiar with HTML/JS etc update the UI to change over to use a dropdown / select on the right hand side to choose between "Unused" / "Master" / "Worker", and the column of checkboxes on the left be removed entirely.

Just realized that I saw a [tip](https://uxdesign.cc/design-better-forms-96fadca0f49c#1a5f) a while ago and we can go further and get rid of the dropdown and show the 3 options as button groups. What do you think?

kiall:
> Just realized that I saw a tip a while ago and we can go further and get rid of the dropdown and show the 3 options as button groups. What do you think?

I think that makes sense, assuming your talking about the "Show all selection options if under 6" tip.

mssola:
@vitoravelino @kiall this change on the design makes sense and it shouldn't break too much what we had before 👍

kiall:
Good State: 3+3 - worked
image

Good State: 3+1 - worked
image

Bad State 3+0 worked:
image

Bad State 2+3 worked:

image

Good State 1+3 worked:

kiall:
Bad State 1+1 worked:

image

Good State 1+2 worked:

image

kiall:
Validated locally, works nicely!

PR #281, opened 2017-08-04
WIP: disable signup feature

flavio:
This PR allows a velum administrator to completely disable the signup feature. This is a requirement of the public cloud team, they want to perform the user creation only through a local command (this is coming after with another PR).

The PR also introduces the configuration mechanism used by Portus:
  * we provide sane settings inside of a `config/config.yml` file
  * users can override/extend them by doing any combination of:
     * provide a custom `config/config-local.yml` file
     * use environment variables

For example, to disable entirely the signup page the admin can start the velum container setting `VELUM_SIGNUP_ENABLED` to `"false"`.

This PR is a WIP, tests have not been extended yet.

### DIFF
--- a/app/controllers/auth/registrations_controller.rb
+++ b/app/controllers/auth/registrations_controller.rb
@@ -15,7 +15,7 @@ class Auth::RegistrationsController < Devise::RegistrationsController
 
   # Restrict the amount of admin users to 1, see bsc#1040831
   def restrict_admin_users
-    return if User.count.zero?
+    return if User.count.zero? && APP_CONFIG.enabled?("signup")
     flash[:alert] = "Admin user already exists."
     redirect_to root_path
   end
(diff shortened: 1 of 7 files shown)


robdaemon:
This will slot in nicely with the RBAC work too 👍

flavio:
I've seen you moved the config code to a dedicated gem: https://github.com/mssola/cconfig

I think this PR should be changed to use that gem :) I'll try to do that before leaving for vacations

mssola:
This is now available on rubygems: https://rubygems.org/gems/cconfig

flavio:
@bear454: does it still make sense to revamp this PR or should I just close it?

bear454:
For Public Cloud, we'll be creating the admin user account via a `rake` task, and therefore have no need for the UI to create the admin user.

ereslibre:
I think we can drop this one for now, right @flavio?

PR #283, opened 2017-08-09
Set database name via environment variable

flavio:
Allow to specify a custom database name using the `VELUM_DB_NAME` environment variable.

The code falls back to the old value when the environment variable is not set.

This is needed to solve some issues with the new CI system.


flavio:
I think we should backport this to the 1.0 branch. Any objections?

PR #285, opened 2017-08-10
use caasp_fqdn instead of unreliable fqdn (bsc#1045841)

MaximilianMeister:
trying to fix https://bugzilla.suse.com/show_bug.cgi?id=1045841 but i haven't tested it yet

it looks like it could be a missing bit and a follow up of https://github.com/kubic-project/salt/pull/93

### DIFF
--- a/app/models/salt_handler/minion_start.rb
+++ b/app/models/salt_handler/minion_start.rb
@@ -29,6 +29,6 @@ class SaltHandler::MinionStart
     return false if minion_info.blank?
 
     # false if a minion with this minion_id or fqdn already exists (uniqueness validation)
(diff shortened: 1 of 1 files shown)


ereslibre:
Since this property is for making nodes easily recognizable I wouldn't go this way, because using `fqdn` looks to me more natural than our internal `caasp_fqdn`.

To avoid issues when a hostname/fqdn of a machine changes we should keep this attributes up to date with a cronjob or something similar that updates all nodes `fqdn` attribute on a timely basis.

MaximilianMeister:
> Since this property is for making nodes easily recognizable I wouldn't go this way, because using fqdn looks to me more natural than our internal caasp_fqdn.

thanks for the input, closing

PR #287, opened 2017-08-14
Fixed 'Accept All Nodes' button ajax request

vitoravelino:
The 'Accept All Nodes' button had the class '.accept-minion' that
is used to accept individually a node. When clicking on it the wrong
event listener callback was being called instead of the correct one.

Now it calls the expected event listener callback and makes the request
passing the expected arguments for all the nodes to be accepted.

For some reason I'm not aware, even not passing `minion_id` Salt was still able to accept all minions.

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -320,7 +320,7 @@ function setPendingAcceptance(minionId) {
 
 function requestMinionApproval(selector) {
   $.ajax({
-    url: '/accept-minion',
+    url: '/accept-minion.json',
     method: 'POST',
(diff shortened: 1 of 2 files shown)


ereslibre:
> For some reason I'm not aware, even not passing minion_id Salt was still able to accept all minions.

When that button is clicked, https://github.com/kubic-project/velum/blob/master/app/assets/javascripts/dashboard.js#L400 is called, what ends up in https://github.com/kubic-project/velum/blob/master/lib/velum/salt.rb#L54-L60 sending `key.accept` with a `*` match, effectively accepting all minions.

PR #288, opened 2017-08-14
[1.0] Fixed 'Accept All Nodes' button ajax request

vitoravelino:
The 'Accept All Nodes' button had the class '.accept-minion' that
is used to accept individually a node. When clicking on it the wrong
event listener callback was being called instead of the correct one.

Now it calls the expected event listener callback and makes the request
passing the expected arguments for all the nodes to be accepted.

(cherry picked from commit b031f0f0216c08d5b0ff7410302bce1f0ad32056)


vitoravelino:
Opened a backport for this. If it's not necessary, feel free to close it.

PR #289, opened 2017-08-24
Add rake task velum:create_user

ereslibre:
This rake task takes two parameters: an email and a password, and will
create a user entry in the Velum database so it's possible to login and
work with it.

### DIFF
--- a/lib/tasks/velum.rake
+++ b/lib/tasks/velum.rake
@@ -1,0 +1,23 @@
+namespace :velum do
+  desc "Create a user"
+  task :create_user, [:email, :password] => :environment do |_, args|
+    if args.count != 2
+      puts "There are 2 required arguments: email and password"
+      exit(-1)
+    end
+
+    args.each do |k, v|
+      if v.empty?
+        puts "You have to provide a value for `#{k}'"
+        exit(-1)
+      end
+    end
+
+    begin
+      User.create! email: args["email"], password: args["password"]
+      puts "User #{args["email"]} created successfully"
+    rescue ActiveRecord::RecordInvalid
+      puts "User #{args["email"]} could not be created. Does it already exist?"
+    end
+  end
+end


ereslibre:
Validations on the model will be executed and the user won't be created if the model validations aren't met. Should we print the validation errors in case this happens?

MaximilianMeister:
> Validations on the model will be executed and the user won't be created if the model validations aren't met. Should we print the validation errors in case this happens?

not necessarily IMO, i leave it up to you

flavio:
> Validations on the model will be executed and the user won't be created if the model validations aren't met. Should we print the validation errors in case this happens?

I agree with @MaximilianMeister. Let's keep things simple, we can always add them later on.

ereslibre:
@flavio Should we backport this patch to 1.0?

bear454:
Sorry, but I don't see how this is any more straightforward than using rails/runner, as most of the work is just getting _anything_ to run on the container:

```
CID="$(docker ps | grep velum-dashboard | cut -f1 -d' ')"
CMD="bundle exec rails runner \"User.create(email: \\\"${ADMIN_EMAIL}\\\", password: \\\"${ADMIN_PASSWORD}\\\")\""
docker exec -it $CID bash -c "VELUM_DB_PASSWORD=\"$(cat /var/lib/misc/infra-secrets/mariadb-velum-password)\" ${CMD}"
```

The thing you guys can do that makes it easier for us is to provide a script at the host-level; but we already have the above code to do that, so I think this issue/FATE can be closed based on the above snippet, unless you can provide a cleaner solution to do this _from the host_.

ereslibre:
@bear454 Well, certainly it doesn't make it _easier_ or more straightforward from the host point of view, but I think it's a nice to have because is a generator that we maintain, so the people wanting to create a user account don't need to mess with the Rails API directly or our implementation, what makes it a better fit for external usage.

Also wanted to comment that you don't need to do `docker exec -it $CID bash -c "VELUM_DB_PASSWORD=\"$(cat /var/lib/misc/infra-secrets/mariadb-velum-password)\" ${CMD}"`, you can do `docker exec -it $CID entrypoint.sh "${CMD}"` and you should be good to go.

From that snippet I'm also wondering if double quotes are going to work in the password field.

I agree on having a wrapper on the host so it's straightforward to make this kind of calls, however that doesn't invalidate the fact that a rake task is better fit for external usages because we maintain it and it will work if we modify our internal schema, or if the rails API change for whatever reason, for example.

bear454:
👍

PR #290, opened 2017-08-24
Clean old jobs on our own, since salt won't do it automatically

ereslibre:
Given our current configuration in which we are no longer using the
mysql `job_cache`, salt won't clean old jobs on its own, so we should
do it so the database won't keep growing without control.

We are executing raw SQL commands to avoid creating `Jid` and `SaltReturn`
models just for the purpose of cleaning, as we know that we want to move
away from a mysql backed salt master.

Fixes: bsc#1054256


app/models/salt_event.rb
```diff
@@ -39,11 +39,30 @@ def self.process(worker_id:)
...
+    ActiveRecord::Base.connection.execute("
+     delete from `jids` where jid in (select distinct jid from salt_returns
+                                      where alter_time < #{date.to_i})
+    ")
+    ActiveRecord::Base.connection.execute("
+      delete from `salt_returns` where alter_time < #{date.to_i};
+    ")
+    ActiveRecord::Base.connection.execute("
+      delete from `salt_events` where alter_time < #{date.to_i};
+    ")
```
MaximilianMeister: this could be done in a loop as the block is so similar:

```ruby
[:events, :returns].each do |table|
  ActiveRecord::Base.connection.execute("
    delete from `salt_#{table}` where alter_time < #{date.to_i};
  ")
end
```

ereslibre:
Please review again.

ereslibre:
Force pushed, so CI retriggers and we can merge.

ereslibre:
Force pushed once again, now with brakeman issue ignored.

PR #293, opened 2017-09-02
Only override blank pillars on the page they are shown

ereslibre:
Before this change the last step of the setup would remove the pillar
information for the first step proxy values (since they were not
protected pillars).

Instead, explicitly say what unprotected pillars there are in each step
and allow the controller to override them with blank if they were provided
blank after having some other value.

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -32,7 +32,9 @@ class SetupController < ApplicationController
   # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
   def configure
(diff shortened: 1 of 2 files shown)


ereslibre:
Sorry, force pushed because I realized I was also missing `no_proxy` field on the `unprotected_attributes`.

ereslibre:
Fixed conflicts, and rebased. Waiting for CI.

ereslibre:
Also, going to adapt this one to also include https://github.com/kubic-project/velum/pull/302 into account.

ereslibre:
PR was updated, waiting for CI.

PR #295, opened 2017-09-04
Allow to restart orchestrations

ereslibre:
This commit includes a few changes:

* `Orchestration` model has been introduced
  * We have a history of all the orchestrations that happened,
    what type of orchestration it was, how long it took to be
    completed and the result of the orchestration.

* `Retryable?` concept has been included in `Orchestration`
  * Right now, only the bootstrap retryable concept has been
    included in the UI, but the backend supports the update orchestration
    too.

* `SaltOrchestration` proxy class has been included, so it can retrieve
  some information from salt-api directly (accessed through the
  Orchestration model).

* Highstates now don't mark minions as successful. Highstates now can
  only mark minions as failed if the highstate failed.

* Only the orchestrations will now mark success when it finishes. All
  pending minions will be marked as succeeded when this happens.

* External orchestrations are detected (we can trigger them completely
  outside Velum and it will detect there is an orchestration running).
  This is because of the introduced salt handler.

* `MinionOrchestration` handler has been renamed to `OrchestrationResult`.

* Helper methods have been introduced to `Velum::Salt`, so it's easier to
  retrieve the list of jobs as well as the detail of one job, given its
  `jid`.

* `salt-api` http calls has now always the same options as in `development`
  and `production` when executed in a `test` environment. Slowness problems
  were because of flaky tests with Capybara waiting too much on things
  we weren't expecting.

* Some VCR cassettes and tests have been added.

orch-restart

Fixes: bsc#1069154


ereslibre:
Should we get this merged for 2.0?

lib/velum/salt_orchestration.rb
```diff
@@ -0,0 +1,22 @@
...
+    attr_accessor :orchestration
+
+    # Initializes a new salt orchestration backend matching orchestration
+    def initialize(orchestration:)
+      @orchestration = orchestration
+    end
+
+    # Return information for this orchestration
+    def info
+      res = perform_request(endpoint: "/jobs/#{@orchestration.jid}", method: "get")
```
MaximilianMeister: i think if you have defined an accessor, you dont need to reference class variables anymore, meaning `s/@orchestration.jid/orchestration.jid/g`

config/routes.rb
```diff
@@ -27,6 +27,10 @@
   post "/update", to: "salt#update"
   post "/accept-minion", to: "salt#accept_minion"
 
+  resource :orchestration, only: [] do
+    match "/retry-bootstrap", action: :retry_bootstrap, via: [:put, :patch]
```
MaximilianMeister: do you think it makes sense also to expose a `retry-upgrade` endpoint?

ereslibre: I think we should do this in the future, yeah... so we can allow to restart upgrade orchestrations from the UI.

lib/velum/salt_orchestration.rb
ereslibre: Oh, yeah, I think is a matter of taste, I try to avoid one extra method call if I have direct access to the variable.

ereslibre:
Heads up: now that this PR is officially out of 2.0 I have to rename the migration and rerun the renamed one (with a newer date), so the migration will run if this gets in as an update.

ereslibre:
PR has been updated with a newer migration, so if this gets in as an update we won't skip this new migration.

Should be ready to be reviewed again.

ereslibre:
[(8) Allow to restart the provisioning orchestration](https://trello.com/c/CwwuN3sU/331-8-allow-to-restart-the-provisioning-orchestration)

ereslibre:
Rebased. I want to check that everything's fine.

db/migrate/20170913145421_create_orchestrations.rb
```diff
@@ -0,0 +1,13 @@
+class CreateOrchestrations < ActiveRecord::Migration
+  def change
+    create_table :orchestrations do |t|
```
MaximilianMeister: @ereslibre could you bump the migration timestamp? as we recently added another one, it wont run.

app/models/orchestration.rb
```diff
@@ -0,0 +1,62 @@
...
+    when :upgrade
+      Orchestration.upgrade.last.try(:status) == "failed"
+    end
+  end
+
+  # Returns the proxy for the salt orchestration
+  def salt
+    @salt ||= Velum::SaltOrchestration.new orchestration: self
+  end
+
```
MaximilianMeister: could you add another method that will check if an orchestration is currently ongoing? this would be another safety layer and would be helpful for a bug where i mentioned this PR as a potential fix: https://bugzilla.suse.com/show_bug.cgi?id=1067194#c8 however to fix this bug, velum needs some more adaption

but nevertheless a method like `self.orchestrating?` would enhance further this new model's usefulness

MaximilianMeister: i just realized this is already in the status of the model! so the information is already there, even if no orchestration has been run yet. so i leave it up to you if you like to put this in another method

ereslibre:
[Reactor callbacks](https://trello.com/c/p7L7JVQo/451-reactor-callbacks)

ereslibre:
It's working fine. Just double checked.

app/models/salt_handler/orchestration_trigger.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+# This class is responsible to handle the salt events that are orchestration triggers.
+# Only the bootstrap and update orchestrations will be handled by this class.
+class SaltHandler::OrchestrationTrigger
```
MaximilianMeister: can we add a base class (e.g. `SaltHandler::Orchestration::Base` or just `SaltHandler::Orchestration`) so that `SaltHandler::Orchestration::Trigger` and `SaltHandler::Orchestration::Result` can inherit from which contains the class methods:

* self.can_handle_event?
* self.salt_fun_args_match

and the constants? this would be easy to do i think and get rid of the duplication

ereslibre:
This is again ready for review! All green since CI is stable again.

app/controllers/orchestrations_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# OrchestrationsController allows to handle orchestrations
+class OrchestrationsController < ApplicationController
+  before_action :can_retry?, only: :retry_bootstrap
+
+  def retry_bootstrap
```
mssola: Would it make sense, to keep a REST style, to change this to the `create` method ?

app/controllers/orchestrations_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    Orchestration.run
+    redirect_to root_path
+  end
+
+  private
+
+  def can_retry?
```
mssola: The question mark on a method in ruby means that a boolean is expected to be returned. I'd chose another name like `redirect_unless_retryable` 🤓

app/models/orchestration.rb
```diff
@@ -0,0 +1,62 @@
...
+  def run
+    raise OrchestrationAlreadyRan unless jid.blank?
+    update_column :status, Orchestration.statuses[:in_progress]
+    _, job = case kind
+             when "bootstrap"
+               Velum::Salt.orchestrate
+             when "upgrade"
+               Velum::Salt.update_orchestration
+    end
+    update_column :jid, job["return"].first["jid"]
```
mssola: Why don't you use `update_columns` and set both `status` and `jid` in the same statement ?

app/models/orchestration.rb
```diff
@@ -0,0 +1,62 @@
...
+    true
+  end
+  # rubocop:enable Rails/SkipsModelValidations
+
+  def self.run(kind: :bootstrap)
+    Orchestration.create(kind: kind).tap(&:run)
```
mssola: This assumes that create will always work, which might not be true in the future if we add validations to it. Maybe you can replace it with `create!`, so it fails more clearly ?

config/routes.rb
```diff
@@ -31,6 +31,10 @@
   get "/oidc/done", to: "oidc#done"
   get "/oidc/kubeconfig", to: "oidc#kubeconfig"
 
+  resource :orchestration, only: [] do
+    match "/retry-bootstrap", action: :retry_bootstrap, via: [:put, :patch]
```
mssola: As I said in my comment, change it to a mere create action (also notice that with a `put` an ID was expected, and you are not doing this here either...)

app/models/orchestration.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def self.retryable?(kind: :bootstrap)
+    case kind
+    when :bootstrap
+      Orchestration.bootstrap.last.try(:status) == "failed"
```
mssola: Isn't status supposed to be an integer ? Shouldn't you compare it with the proper enum value ?

spec/spec_helper.rb
```diff
@@ -8,6 +8,9 @@
   coverage_dir "public/coverage"
 end
 
+ENV["VELUM_SALT_HOST"] ||= "127.0.0.1"
```
mssola: Why have you done this ? Is this relevant to this PR ? 😕

app/models/orchestration.rb
ereslibre: Because you need to set `status` to `in_progress`, then call the orchestration, and then save the `jid` that was returned by calling the orchestration.

app/controllers/orchestrations_controller.rb
ereslibre: It's not exactly a `create`... as we are requesting a retry. Calling to this endpoint when no previous orchestration exists will fail, also it doesn't make sense to POST on `/orchestrations/:id/retry` or something similar, as we aren't really retrying a specific run, just a specific orchestration type.

Honestly, I'd leave it as it is for now... I know it's not very RESTy...

config/routes.rb
ereslibre: It's not a very RESTy approach. You can allow `put` or `patch` without the need of an ID if what you are updating is a (let's call it) singleton resource, like for example a global `/config` endpoint. It makes sense to allow `put` and `patch` on the global `/config` resource without the need of an id.

spec/spec_helper.rb
ereslibre: Leftover! Thanks!

app/models/orchestration.rb
mssola: Gotcha

app/models/orchestration.rb
mssola: I see.

spec/spec_helper.rb
vitoravelino: Forgot to remove this?

PR #301, opened 2017-09-08
LDAP - remove hardcoded bind password

robdaemon:
Passes the LDAP parameters into the pillar so Dex can be bootstrapped
properly

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -29,8 +29,7 @@ class User < ApplicationRecord
     # check to see if this is because the LDAP auth succeeded, or if we're coming from registration
     # we do this by performing an LDAP search for the new user. If it fails, we need to create the
     # user in LDAP
-    cfg_file = ::Devise.ldap_config || Rails.root.join("config", "ldap.yml")
-    ldap_config = YAML.safe_load(ERB.new(File.read(cfg_file)).result)[Rails.env]
+    ldap_config = Velum::LDAP.ldap_config
 
     conn_params = {
       host: ldap_config["host"],
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -99,7 +99,7 @@ class SetupController < ApplicationController
       settings["https_proxy"] = ""
       settings["no_proxy"] = ""
     end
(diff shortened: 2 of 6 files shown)


robdaemon:
Depends-on: kubic-project/velum#301
Depends-on: kubic-project/salt#222
Depends-on: kubic-project/caasp-container-manifests#105

app/models/user.rb
```diff
@@ -29,8 +29,7 @@ def create_ldap_user
     # check to see if this is because the LDAP auth succeeded, or if we're coming from registration
     # we do this by performing an LDAP search for the new user. If it fails, we need to create the
     # user in LDAP
-    cfg_file = ::Devise.ldap_config || Rails.root.join("config", "ldap.yml")
-    ldap_config = YAML.safe_load(ERB.new(File.read(cfg_file)).result)[Rails.env]
+    ldap_config = Velum::LDAP.ldap_config
```
ereslibre: We can certainly refactor this after the merge so we focus only on the functionality being introduced, but I think this would be good to be read on an initializer (e.g. `config/initializers/ldap.rb`) so we only load this once during the initialization phase of the application.

kiall:
These work fine, however there is a issue with the ldap container entrypoint and escaping slashes, which seems to hit every few runs.

ereslibre:
I can also confirm I see authentication failures when I see a slash in the password, like this last run (output of the openldap container):

```
+ password_hash='{SSHA}gY9/R+q8nYvCzNihaHK3rNpsCHt+1NV6'
+ sed_safe_password_hash='{SSHA}gY9\/R+q8nYvCzNihaHK3rNpsCHt+1NV6'
+ rootpwd_string='rootpw {SSHA}gY9\/R+q8nYvCzNihaHK3rNpsCHt+1NV6'
```

And what happens when I'm going to register:

screenshot from 2017-09-12 17-44-23

robdaemon:
Fixed the password escaping in IBS 140879

robdaemon:
rebased, ready to go

kiall:
LDAP container fixed, working just fine now!

PR #303, opened 2017-09-11
Set etcd2 backend for old installations

ereslibre:
Depends on: https://github.com/kubic-project/salt/pull/224

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -58,6 +58,9 @@ class SetupController < ApplicationController
       redirect_to setup_bootstrap_path, alert: res
       return
     end
+    Pillar.find_or_create_by pillar: "api:etcd_version" do |pillar|
+      pillar.value = "etcd3"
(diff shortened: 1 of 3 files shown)


app/controllers/setup_controller.rb
```diff
@@ -58,6 +58,9 @@ def do_bootstrap
       redirect_to setup_bootstrap_path, alert: res
       return
     end
+    Pillar.find_or_create_by pillar: "api:etcd_version" do |pillar|
+      pillar.value = "etcd3"
+    end
```
ereslibre: This is created for consistency reasons, so all installations have the `api:etcd_version` override. Old installations will have `etcd2` because of the migration, while new installations will have `etcd3` because of this.

I still have my doubts about the need of a consistent overlay between versions, but included so it's possible to discuss this.

mjura:
it is also related to https://github.com/kubic-project/salt/pull/210

PR #304, opened 2017-09-11
Fixed proxy settings when revisiting the page

### LINKED ISSUE #300: Proxy settings are lost visually after refreshing the page
The proxy settings, while saved in the database do not match the visual page when it is refreshed, in particular:

* Enable it, go to next step and back again -- looks like it is disabled
* Same for "systemwide" proxy selection

vitoravelino:
After the first step of setting up the proxy settings,
when the user goes back to that initial page, the proxy
settings were lost. This can not only confuse the user
but also introduce undesired changes.

With this fix it doesn't happen anymore.

Fixes #300


vitoravelino:
Rebased!

PR #306, opened 2017-09-15
Consistent naming of the product

ereslibre:
- SUSE CaaSP
- SUSE Container as a Service Platform

both should be SUSE CaaS Platform.

Fixes: bsc#1058720


ajaeger:
Those changes look fine for the text itself, thanks!

PR #309, opened 2017-09-18
WIP: If a system won't be registered, set a default timeout for that warning during the installation

ereslibre:
Right now it's completely necessary for the customer to confirm the "this
system is not going to be registered" warning, or the installation won't
continue.

Make this dialog have a timeout so we show it at the very beginning, but
it's also possible to install unattended workers without further action.

Fixes: bsc#1059016

### DIFF
--- a/app/views/dashboard/autoyast.xml.erb
+++ b/app/views/dashboard/autoyast.xml.erb
@@ -81,6 +81,7 @@
       <ask>
         <default>This system will not be registered. Reason: <%= @registration_error %></default>
         <type>static_text</type>
+        <timeout config:type="integer">30</timeout>
       </ask>
     </ask-list>
     <% end %>


ereslibre:
`WIP` because needs to be tested.

thkukuk:
This is only a workaround for something, which looks like a bug in autoyast. If autoyast behavior has really changed, we need to fix that and not workaround it, because a lot of customers would be affected by this.
Between, there is already an older bug report for the YaST2 team.

ereslibre:
> This is only a workaround for something, which looks like a bug in autoyast. If autoyast behavior has really changed, we need to fix that and not workaround it, because a lot of customers would be affected by this.Between, there is already an older bug report for the YaST2 team.

I'm not aware of any change in the autoyast behaviour. We have introduced a new warning for workers when they are not going to be registered without at timeout, what forces the customer to click "OK" on every single worker when they are performing the autoyast installation.

I don't consider this a workaround, but setting a timeout for keeping the old behaviour the same: we show a warning that will be dismissed after 30 seconds, so if someone is performing a big unattended install, they don't need to go through every single worker clicking "OK".

ereslibre:
The behavioral change happened here: https://github.com/kubic-project/velum/pull/270

ereslibre:
Some discussions regarding this change happening at: https://github.com/kubic-project/velum/pull/270#issuecomment-330191616

ereslibre:
Closing, because instead I'm reverting https://github.com/kubic-project/velum/pull/270 partially.

PR #310, opened 2017-09-18
ux: improved messages for 'Next' button on nodes selection

vitoravelino:
The title message for the 'Next' button on the nodes selection
page were out of date and weren't helping the user understand
what was wrong.

For each rule we use to validate the 'Next' button we
are adding a dedicated message to show to the user. This way
the user will know exactly what's wrong with the cluster
configuration.

I also fixed the "minimum of three nodes" alert to only be displayed
if the user didn't select 3 nodes (dones't matter if master was not
selected). The previous behavior was making it display if the
configuration wasn't supported even having 4 nodes selected. This
info now goes on the 'Next' button title message.

See bsc#1058915

screenshot 2017-09-25 10 29 08
screenshot 2017-09-25 10 26 35
screenshot 2017-09-25 10 27 13

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,3 +1,8 @@
+State = {
+  nextClicked: false,
+  bootstrapErrors: [],
+},
+
 MinionPoller = {
   selectedNodes: [],
   selectedMasters: [],
@@ -149,7 +154,7 @@ MinionPoller = {
 
         MinionPoller.handleAdminUpdate(data.admin || {}, hasPendingStateNode);
 
-        handleBootstrapButtonTitle();
+        handleBootstrapErrors();
 
         // show/hide "update all nodes" link
         var hasAdminNodeUpdate = data.admin.update_status === 1 || data.admin.update_status === 2;
@@ -185,7 +190,7 @@ MinionPoller = {
         $('.nodes-loading').hide();
         $('.nodes-content').removeClass('hidden');
 
-        toggleBootstrapButton();
+        handleBootstrapErrors();
         toggleMinimumNodesAlert();
       }
     }).always(function() {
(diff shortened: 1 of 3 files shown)


ereslibre:
Sorry for being this late on this review...

What about following the same strategy as the box that is on the top of the discovery page if you click less than 3 nodes? I think it would be nice if the contents of that box could be dynamic (with bullet points). Something like:

```
This cluster is not bootstrappable. The reasons are:
  * You have not selected one master at least
  * The number of masters has to be an odd number
  * You haven't selected one worker at least
```

In this case, instead of being an alert it would be an error (red background), so that catches the attention on why the `Next` button is disabled. I think it would integrate much better on the current page.

vitoravelino:
Showing the alert since the beginning doesn't sound good. What I thought was that we can leave the button enabled and when the user clicks on it and the configuration is not valid, the alert appears. The rest stays as you suggested.

What do you think?

ereslibre:
> What do you think?

I think it's fine, as long as the error contains all the conditions that were not met (as it is now), so the user does not end in a trial & error loop. :)

vitoravelino:
Cool. I'll work on it. Thanks!

vitoravelino:
@ereslibre Updated!

vitoravelino:
I need to update the specs since the behavior changed a little bit on the "Next" button.

vitoravelino:
Done! 😃

app/assets/javascripts/dashboard.js
```diff
@@ -499,73 +504,51 @@ function selectedMastersLength() {
...
-    return false;
-  }
-
   // We need at least three nodes in total
-  if ((selectedWorkersLength() + selectedMastersLength()) < 3) {
-    return false;
-  }
-
-  return true;
+  return (selectedWorkersLength() + selectedMastersLength()) >= 3;
```
ereslibre: Can't we remove `isSupportedConfiguration`? I think we could just use `isBootstrappable` with an option to update `State.bootstrapErrors` only when we clicked on `Next`. I think it would simplify some logic, since `isSupportedConfiguration` is just a simplified version of `isBootstrappable`, no?

vitoravelino: We need `isSupportedConfiguration` to decide if we show or not the warning modal when 1 master and 1 worker and also the minimum nodes alert.

ereslibre: Fair enough.

vitoravelino:
@ereslibre should we add someone else to review this or merge it?

ereslibre:
@vitoravelino Added @MaximilianMeister and @mssola, it should get two approvals at least.

app/views/setup/discovery.html.slim
```diff
@@ -3,6 +3,13 @@
   span
     | A supported deployment of SUSE CaaS Platform requires a minimum of three nodes. Please select a minimum of three nodes.
 
+.alert.alert-danger.discovery-bootstrap-alert role="alert" hidden="true"
+  i.fa.fa-4x.pull-left aria-hidden="true"
+  span
+    p This cluster is not bootstrappable. The reasons are:
```
MaximilianMeister: i think `bootstrappable` as an internal name is fine, but in the UI i'd call it differently, maybe:

```
Unable to bootstrap cluster:

* list
* list
```

vitoravelino: Done!

ereslibre:
We should open a PR against the `release-2.0` branch too.

flavio:
> We should open a PR against the release-2.0 branch too.

+1 

Can we merge this PR into master, it looks like everything is in place.

flavio:
> Can we merge this PR into master, it looks like everything is in place.

I just realized we already have a backport of this PR: https://github.com/kubic-project/velum/pull/318

PR #312, opened 2017-09-18
[CNI] Set default flannel backend to UDP for old installations

ereslibre:
This migration will only be executed on old installations, while on new ones
(where we will use the default salt pillar value) this migration won't be
run, so there won't be a pillar override.

Depends on: https://github.com/kubic-project/salt/pull/201

### DIFF
--- a/db/migrate/20180109145421_set_udp_as_default_flannel_backend.rb
+++ b/db/migrate/20180109145421_set_udp_as_default_flannel_backend.rb
@@ -1,0 +1,10 @@
+class SetUdpAsDefaultFlannelBackend < ActiveRecord::Migration
+  def up
+    Pillar.find_or_create_by pillar: "flannel:backend" do |pillar|
+      pillar.value = "udp"
+    end
+  end
+  def down
+    Pillar.where(pillar: "flannel:backend").destroy_all
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171117145421) do
+ActiveRecord::Schema.define(version: 20180109145421) do
 
   create_table "jids", id: false, force: :cascade do |t|
     t.string "jid",  limit: 255,      null: false


app/controllers/setup_controller.rb
```diff
@@ -74,9 +74,6 @@ def do_bootstrap
       redirect_to setup_bootstrap_path, alert: res
       return
     end
-    Pillar.find_or_create_by pillar: "api:etcd_version" do |pillar|
-      pillar.value = "etcd3"
```
MaximilianMeister: this can be dropped again with https://github.com/kubic-project/velum/pull/313

ereslibre:
[[CNI] Set default flannel backend to UDP for old installations](https://trello.com/c/qHQZZflz/452-cni-set-default-flannel-backend-to-udp-for-old-installations)

ereslibre:
Rebasing again and bumping schema migration version so it'll run.

ereslibre:
Rebased. Ready to be merged.

This will set `udp` as the default backend for old installations in a CNI context.

flavio:
We could merge this code into the master branch now that 3.0 is on the horizon, however:

  * There are some conflicts to be fixed
  * Is it really wise to stay with the `udp` backend? I think we should move to the vxlan one as [recommended by upstream](https://github.com/coreos/flannel/blob/master/Documentation/backends.md#recommended-backends).

inercia:
> We could merge this code into the master branch now that 3.0 is on the horizon, however:
> * There are some conflicts to be fixed
> * Is it really wise to stay with the udp backend? I think we should move to the vxlan one as recommended by upstream.

@flavio New installations will use VXLAN as the default backend, but we are explicitly setting UDP as the default backend _for old nodes_ because, otherwise, they would be upgraded automatically to a VXLAN configuration on the upgrade. That would make them lose connectivity with the rest of the nodes that have not been upgraded yet, as the old nodes would still use UDP and the upgraded nodes would use VXLAN. By configuring old nodes with UDP, all the nodes will use UDP aftfer the upgrade.

ereslibre:
I will update this PR to bump the migration number so it really runs. Will fix.

ereslibre:
This has been rebased and the migration version has been bumped.

flavio:
@inercia, I see. But that means that new clusters are still going to use `udp` as flannel backend. Am I right?

kiall:
@flavio No, new installs will be unaffected by this PR, as migrations are only ran during upgrades and not during install. Another PR to the salt project will change the default to vxlan, This change, and that one, combined will mean new installs get set to vxlan, and old installs stay on udp.

PR #313, opened 2017-09-18
Do not add an override that will have the same value as the overriden one

ereslibre:
This will only be effective on new installations, where the default pillar
value matches this value. Let's keep things simple and only override what
is in the UI or what will/can be different between versions.

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -74,9 +74,6 @@ class SetupController < ApplicationController
       redirect_to setup_bootstrap_path, alert: res
       return
     end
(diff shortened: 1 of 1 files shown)


ereslibre:
I would like to merge this for `2.0`. Is a minor change with small impact, and will avoid having unneeded overrides.

ereslibre:
`etcd3` is already the default in the salt pillar: https://github.com/kubic-project/salt/pull/224

Let's avoid a useless override, at least until it's necessary.

MaximilianMeister:
just confirmed with alex that this will go into 2.0 (but not RC1 anymore)

PR #314, opened 2017-09-20
Perform an OIDC auth and return a full kubeconfig

robdaemon:
This will return a full kubeconfig file after forcing the user though an auth with Dex on the running cluster

bsc#1058568


robdaemon:
Depends on kubic-project/salt/pull/235

robdaemon:
The goal here:

Any user (anonymous!) hits `https://dashboard/oidc` - they are redirected into Dex in the CaaSP cluster, where they enter their cluster username and password. After Dex authorizes them, Velum fetches and decodes their JWT, and writes a full kubeconfig file, exactly like caasp-cli does today. This means you can choose not to use caasp-cli, because you have access to a browser.

This doesn't really have a UI. The "Download kubeconfig" button should basically open a popup that goes through this flow.

Note: we really need to rebrand the Dex login so it looks like a SUSE thing, I think :)

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  # Dex.
+  skip_before_action :authenticate_user!  
+
+  def start
+    redirect_to authorization_uri(SecureRandom.hex(16))
+  end
+
+  def done
+    render "start.html.erb"
+  end
```
MaximilianMeister: everything below should be in a model or a library, there is currently too much business logic in the controller

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+# OidcController is used for performing an OpenID Connect auth
+# flow against Dex, allowing Velum to generate a complete kubeconfig
+# file for a user, including their JWT token
+class OidcController < ApplicationController
+  # allow anyone to start this, the auth will happen inside
+  # Dex.
+  skip_before_action :authenticate_user!  
+
+  def start
```
MaximilianMeister: i'd prefer if `caasp-cli` would be able to connect to this endpoint, then we have a single authorization entrypoint for UI and CLI

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    client.authorization_uri(
+      response_type: :code,
+      nonce: nonce,
+      state: nonce,
+      scope: [:openid, :profile, :email, :offline_access, :groups].collect(&:to_s))    
+  end
+
+  def client
+    OpenIDConnect.http_config do |config|
+      config.ssl_config.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
ereslibre: Do we need to skip verification if we are adding the trusted CA now (with the changes below and on `bin/init`)? I understand this can be removed so it's actually verified.

bin/init
```diff
@@ -6,6 +6,14 @@ set -e
 
 : ${VELUM_PORT:=80}
 
+setup_root_ca() {
+  # Velum is going to need this CA to talk to the running CaaSP Cluster
+  [ -f "/etc/pki/trust/anchors/SUSE_CaaSP_CA.pem" ] && return
```
ereslibre: I think we can tweak this in case the CA changes for whatever reason in the future (I can think of a process in which the administrator completely requests a CA reissuing along with all certificates because of a security issue within the company for example). Maybe we can only return if `/etc/pki/ca.crt` sha2 does not match `/etc/pki/trust/anchors/SUSE_CaaSP_CA.pem` sha2. In any case I'm fine as it is right now, I'd create an issue for not forgetting that at least.

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    client.redirect_uri = url_for(controller: "oidc", action: "done")
+    client.authorization_uri(
+      response_type: :code,
+      nonce: nonce,
+      state: nonce,
+      scope: [:openid, :profile, :email, :offline_access, :groups].collect(&:to_s))    
+  end
+
+  def client
+    OpenIDConnect.http_config do |config|
```
ereslibre: Cannot this block be completely removed with the initializer?

config/puma.rb
```diff
@@ -44,7 +44,8 @@
 # end
 
 if ENV["VELUM_PORT"].to_i == 443
-  ssl_bind "0.0.0.0", ENV["VELUM_PORT"], key: "/etc/pki/velum.key", cert: "/etc/pki/velum.crt"
+  ssl_bind "0.0.0.0", ENV["VELUM_PORT"], key: "/etc/pki/velum.key", cert: "/etc/pki/velum.crt",
+    ca: "/etc/pki/ca.crt"
```
ereslibre: As far as I can tell `ssl_bind` does not support `ca`, it's just ignored: http://www.rubydoc.info/github/puma/puma/Puma%2FConfiguration%2FDSL%3Assl_bind

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  # allow anyone to start this, the auth will happen inside
+  # Dex.
+  skip_before_action :authenticate_user!  
+
+  def start
+    redirect_to authorization_uri(SecureRandom.hex(16))
+  end
+
+  def done
+    render "start.html.erb"
```
ereslibre: `render :start` should also do it

ereslibre:
Regarding `openid_connect` gem addition (along with some dependencies), I think we really need this one, please @robdaemon correct me if I'm wrong, but discovery specially can be a can of worms.

app/controllers/oidc_controller.rb
robdaemon: There's something wrong and the trusted CA doesn't work. Basically, the ruby httpclient library does not use the system trust store. It ships with its own. The line below this one that adds the system trust store isn't working. I need to dig in further.

bin/init
robdaemon: oh, good point.

config/puma.rb
robdaemon: okay. That was part of me trying to get the CA working elsewhere. I can remove this part.

robdaemon:
@ereslibre @MaximilianMeister Yeah, we really do need that gem. OpenID Connect is a big ball of mud. It would be tricky to do this properly without it. 😢  Also, there could be security implications if we did it manually and wrong.

robdaemon:
Current state: **It works!**

Now to address the RSpec tests and make it pretty....

MaximilianMeister:
> @ereslibre @MaximilianMeister Yeah, we really do need that gem. OpenID Connect is a big ball of mud. It would be tricky to do this properly without it.  Also, there could be security implications if we did it manually and wrong.

@robdaemon i provided the deps now in `devel:languages:ruby:extensions`, but i'll wait to submit the gems to casp until we finished the branching for 2.0 (on monday) if you don't mind

ereslibre:
> Current state: It works!
>
> Now to address the RSpec tests and make it pretty....

👏 I will test this today! 👍

robdaemon:
@ereslibre okay this is done! As far as making it "pretty" or something, I need someone other than me to take that. I got the workflow going and it generates a complete kubeconfig file now.

spec/controllers/oidc_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+    end
+
+    it "generates a kubeconfig if auth succeeds" do
+      # VCR can do the salt stuff, we'll just mock Dex
+      VCR.use_cassette("kubeconfig/cluster_ready2", record: :none) do
+        setup_done
+        external_fqdn_pillar
+
+        mock_json :get, "https://api.example.com:32000/.well-known/openid-configuration",
```
ereslibre: Why don't use `VCR` instead?

app/assets/javascripts/oidc.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
```
ereslibre: We should remove this file if it's blank.

app/assets/stylesheets/oidc.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the oidc controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
ereslibre: Same here.

spec/helpers/webmock_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+    response = {}
+    response[:body] = File.new(File.join(File.dirname(__FILE__), "../fixtures",
+      "#{response_file}.#{options[:format] || :json}"))
+    response[:status] = options[:status] if options[:status]
+    response
+  end
+end
+
+include WebMockHelper
+WebMock.disable_net_connect!
```
ereslibre: Please correct me if I'm wrong, but could we completely remove this and use `VCR` instead?

spec/controllers/oidc_controller_spec.rb
robdaemon: I spent three days trying to use VCR with no luck. The OIDC handshake is fairly complex - the URLs have to match, the nonce has to match, etc. It was much easier to perform a handshake using the CLI and record the HTTP traffic and use that to form these tests.

robdaemon:
If someone else can make VCR work, great, but I really couldn't figure it out. :(

ereslibre:
@robdaemon I was able to finally test this along with https://github.com/kubic-project/salt/pull/235 and for some reason if I click on the `kubectl config` button all I get is a view with "OpenID handshake starts here" and if I manually go to `/oidc` I get an error `certificate verify failed`. I'm not sure I'm testing this properly though... or maybe I'm missing something...

robdaemon:
@ereslibre okay, I clearly messed up some stuff and didn't even see it because I was so focused on RSpec. :( Looks like I got it now!

robdaemon:
@ereslibre Things to think about with this -

It's difficult to write a test for the OpenID Connect handshake, either using the mocking method I used, or with VCR, because the tokens will expire, then the test will fail.

So, I can think of sogratme possible ways to do this:

1. Run an OAuth server during the test, and test talking to that. Problem: The goal here is integration with Dex. Running the OAuth server in Rails and authing to it doesn't test the actual integration we need.
1. Stub out all the method calls. While this would work to achieve full test coverage, it isn't a good enough test - it only tests that the methods are called. It doesn't prove that they are called correctly, or in the correct order, so wouldn't be a valid way of knowing if OAuth integration was happening as desired.
1. Move this responsibility to the end-to-end tests. Integration is what matters here, not a unit test. Unit tests are great for algorithms, but mean little in testing interactions with an external system. This means we must lower the unit coverage requirements. Philosophically, 100% test coverage doesn't really mean anything - it doesn't guarantee the tests are good tests, for example. See https://martinfowler.com/bliki/TestCoverage.html . Note, if we go this route, then the real test coverage numbers have to run *during* the e2e tests also, and then aggregate the results. That would tell you the true coverage.

robdaemon:
Okay, FINALLY got it! I had to switch from using VCR or any other mocked OIDC responses to using class and instance doubles, but, it gets 100% coverage and doesn't have any issues with tokens expiring breaking the build.

ereslibre:
Testing this now.

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+      response_type: :code,
+      nonce:         nonce,
+      state:         nonce,
+      scope:         [:openid, :profile, :email, :offline_access, :groups].collect(&:to_s)
+    )
+  end
+
+  def oidc_config
+    @oidc_config ||= OpenIDConnect::Discovery::Provider::Config.discover! \
+      "https://#{Pillar.value(pillar: :apiserver)}:32000"
```
ereslibre: On my deployment I see that the dashboard container is not able to resolve the hostname I have set to the apiserver. My point here: the `apiserver` pillar is the client reachable name of the `apiserver` machine, but that doesn't mean the admin node will be able to resolve this name.

I think it would be interesting to get the `caasp_fqdn` grain for one machine having the `kube-master` role, and then perform this discovery using this name (the admin node has this information on `/etc/hosts` so it will be able to reach the apiserver with this name, and the apiserver cert contains this name). In my current instance it looks like: `aeda402b236d41ab8476fb9cdefaba3a.infra.caasp.local`.

This would suffice to get this information: `Minion.master.sample.salt.info["caasp_fqdn"]`, so something like this:

```
 @oidc_config ||= OpenIDConnect::Discovery::Provider::Config.discover! \
      "https://#{Minion.master.sample.salt.info["caasp_fqdn"]}:32000"
```

Something similar could be done for the `issuer` method (`Minion.master.sample` could take different masters though, so maybe we could go with `first` for now).

ereslibre:
Arrggg, I'm getting `Issuer mismatch` now (modified the controller to return `"https://10.17.2.0:32000"` on the `issuer` as well as changing it on the `discover!` call.

What value do you set for `apiserver` in order for this feature to work?

app/controllers/oidc_controller.rb
robdaemon: That won't work. It must be the same name, or we're going to have certificate issues. Why is your machine different than mine? I'm using caasp-kvm, and it works?

Sounds like we must update the /etc/hosts file on the admin node so that it's always able to find itself by the same name the user uses.

robdaemon:
In my case, I leave it set as the default. I run caasp-kvm, and connect to velum as https://10.17.1.0

app/controllers/oidc_controller.rb
ereslibre: > That won't work. It must be the same name, or we're going to have certificate issues. Why is your machine different than mine? I'm using caasp-kvm, and it works?

If we use `first` instead of `sample` it should work. The admin node has the `/etc/hosts` updated automatically by salt but the container has its own ideas about the `/etc/hosts` file.

I'll check this PR first thing in the morning again, not sure why it didn't work for me, even forcing the IP address (hardcoding it) -- that's when I got the `Issuer mismatch`, maybe something wrong on my side 😞

ereslibre: As in it should work I mean we should be referring to the same machine, if that would or not work with OpenID I have no idea (I don't know OpenID connect internals well enough).

I confirmed this `caasp_fqdn` name is present on the apiserver certificate, but didn't check further.

ereslibre:
> In my case, I leave it set as the default. I run caasp-kvm, and connect to velum as https://10.17.1.0

What are you setting the apiserver to? Just want to confirm everything so I can properly 👍 this tomorrow.

robdaemon:
@ereslibre yep, I left it default. In caasp-kvm, the default works as long as you're on the R&D VPN. Otherwise, I use `api.10.17.2.0.nip.io` - nip.io is really handy.

ereslibre:
/me happy dance

It failed for me if I do it very fast (go to download it immediately after deploying the cluster), but upon retry it works. Now for the real review! Thanks for all your efforts!

ereslibre:
I believe this has something to do with the "not immediately available":

```
96635e4f7753        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Up 6 minutes                                   k8s_dex_dex-3532614258-sh534_kube-system_f476549c-a8ec-11e7-8542-46c3af56d27b_1
1c4379121ea2        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Up 6 minutes                                   k8s_dex_dex-3532614258-w628p_kube-system_f4762d3a-a8ec-11e7-8542-46c3af56d27b_1
8cde318e19e8        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Up 6 minutes                                   k8s_dex_dex-3532614258-0931q_kube-system_f4760708-a8ec-11e7-8542-46c3af56d27b_1
6dbe9eecf34a        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Exited (2) 6 minutes ago                       k8s_dex_dex-3532614258-sh534_kube-system_f476549c-a8ec-11e7-8542-46c3af56d27b_0
0e0d1926e98f        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Exited (2) 6 minutes ago                       k8s_dex_dex-3532614258-w628p_kube-system_f4762d3a-a8ec-11e7-8542-46c3af56d27b_0
e6f3b331a240        ab57cd5d2cd4         "/usr/bin/dex serve /"   6 minutes ago       Exited (2) 6 minutes ago                       k8s_dex_dex-3532614258-0931q_kube-system_f4760708-a8ec-11e7-8542-46c3af56d27b_0
```

All of them raise the same issue:

```
caasp-worker-0:~ # docker logs 6dbe9eecf34a
time="2017-10-04T10:16:39Z" level=info msg="config issuer: https://caasp-master-0.devenv.caasp.suse.net:32000" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created auth-code.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created auth-request.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created o-auth2-client.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created signing-key.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created refresh-token.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created password.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created offline-sessions.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="third party resource already created connector.oidc.coreos.com" 
time="2017-10-04T10:16:39Z" level=info msg="config storage: kubernetes" 
time="2017-10-04T10:16:39Z" level=info msg="config static client: caasp-cli" 
time="2017-10-04T10:16:39Z" level=info msg="config connector: ldap" 
time="2017-10-04T10:16:39Z" level=info msg="config skipping approval screen" 
failed to initialize server: server: failed to list connector objects from storage: failed to list connectors: not found
```

This seems to be a transient issue that solves itself without further problems.

ereslibre:
I could verify that it works perfectly logged in and not logged in (visiting directly `/oidc`).

Checking how we can brand it to be more SUSE-like, but overall it looks extremely good for me.

Will continue with the technical review as comments on the patch itself.

ereslibre:
I think it would make sense to mount `/usr/share/caasp-dex/web/themes/caasp` in `dex` container. The question would be: what package should include this theme for this to be clean? A new package maybe? I'd try to avoid carrying a patch along with dex with our own theme, it would be more work to upgrade dex. What do you think?

flavio:
> I think it would make sense to mount /usr/share/caasp-dex/web/themes/caasp in dex container. The question would be: what package should include this theme for this to be clean? A new package maybe? I'd try to avoid carrying a patch along with dex with our own theme, it would be more work to upgrade dex. What do you think?

I would prefer to not introduce another mount on the host. Let's just make sure the dex image we are using has the theming in place.

I would go for one of these options:

  * Add the theme as a patch to our dex rpm
  * Create a repo with the theme, create a package shipping the theme, add this theme package to the dex image

MaximilianMeister:
> I would go for one of these options:
> 
> Add the theme as a patch to our dex rpm
> Create a repo with the theme, create a package shipping the theme, add this theme package to the dex image

We could also have a fork and a branding-branch to generate the patch from and update it when necessary, that's more easy when it comes to rebasing than having just a patch in obs i think

ereslibre:
The second option @flavio suggests LGTM, I think is the easier option to maintain.

flavio:
Note well: I have no idea about how theming is handled by dex. If that's just a matter of dropping an html file and some css into a folder then great.

But if you have to do heavier changes to it (eg: alter go files) I would definitely go with @MaximilianMeister approach.

ereslibre:
> But if you have to do heavier changes to it (eg: alter go files) I would definitely go with @MaximilianMeister approach.

No, only config settings (that are provided by the dex.yaml manifest in our salt states) and dropping this themes. Confirming this ATM though.

flavio:
> No, only config settings (that are provided by the dex.yaml manifest in our salt states by a ConfigMap) and dropping this themes. Confirming this ATM though.

Cool, if that's the case I think the easiest approach would be:

  * leave dex package as it currently is
  * create git repo with the suse branding
  * create git repo with the openSUSE branding (maybe the same repo above)
  * create dex-(suse|openSUSE)-branding packages
  * add the dex(suse|openSUSE)-branding package to the dex image spec file
  * add the configuration/set the environment variable that enables our them inside of the dex image spec file

ereslibre:
Hey @robdaemon, could you please squash the commits and refer to the bug number it's fixing in the resulting commit? Thank you!

robdaemon:
@ereslibre squashed, and added the bsc# to the commit message

PR #315, opened 2017-09-22
Implement Pillar endpoint and introduce internal-api

ereslibre:
Depends on:

* https://github.com/kubic-project/salt/pull/238
* https://github.com/kubic-project/caasp-container-manifests/pull/116

Fixes: bsc#1069145

### DIFF
--- a/app/controllers/concerns/api.rb
+++ b/app/controllers/concerns/api.rb
@@ -1,0 +1,24 @@
+# Api implements general use case methods when working with API controllers.
+module Api
+  extend ActiveSupport::Concern
+
+  def ok(status: 200, content: nil)
+    if content.blank?
+      render status: status, nothing: true
+    else
+      respond_to do |format|
+        format.json { render status: status, json: content.as_json }
+      end
+    end
+  end
+
+  def ko(status: 422, content: nil)
+    if content.blank?
+      render status: status, nothing: true
+    else
+      respond_to do |format|
+        format.json { render status: status, json: content.as_json }
+      end
+    end
+  end
+end
--- a/app/controllers/internal_api/v1/pillars_controller.rb
+++ b/app/controllers/internal_api/v1/pillars_controller.rb
@@ -1,0 +1,19 @@
+# Serve the pillar information
+class InternalApi::V1::PillarsController < InternalApiController
+  def show
+    ok content: pillar_contents
+  end
+
+  private
+
+  def pillar_contents
+    pillar_struct = {}.tap do |h|
+      Pillar.all_pillars.collect { |k, v| h[v] = Pillar.value(pillar: k.to_sym) }
+    end
+    {}.tap do |json_response|
+      pillar_struct.each do |key, value|
+        json_response.deep_merge! key.split(":").reverse.inject(value) { |a, n| { n => a } }
+      end
+    end.deep_symbolize_keys
+  end
+end
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -1,36 +1,5 @@
 # frozen_string_literal: true
 # Pillar represents a pillar value on Salt.
-#
-# This can override already existing pillar values, or create completely new ones.
-#
-# For example, given that we have on our static pillar:
-#
-#   certificate_information:
-#     subject_properties:
-#       C: DE
-#       O: SUSE
-#
-# We could override them by creating the following Pillar records:
-#
-#   Pillar.create pillar: "certificate_information:subject_properties:C", value: "FR"
-#   Pillar.create pillar: "certificate_information:subject_properties:O", value: "My Company"
-#
-# Lists can be represented as collisions. So, creating the following Pillar records:
-#
-#   Pillar.create pillar: "my_service:extra_ips", value: "127.0.0.1"
-#   Pillar.create pillar: "my_service:extra_ips", value: "127.0.0.2"
-#
-# Would match the following YAML representation:
-#
-#   my_service:
-#     extra_ips:
-#       - 127.0.0.1
-#       - 127.0.0.2
-#
-# This is because how we configure our salt master using this table as an external pillar. For
-# further information you can visit:
-#   - https://github.com/kubic-project/velum/blob/master/docs/salt.md
-#   - https://github.com/kubic-project/salt/blob/master/config/master.d/50-returner.conf
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
   validates :value, presence: true
--- a/config/initializers/internal_api.rb
+++ b/config/initializers/internal_api.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+# Be sure to restart your server after modifying this file.
+
+Rails.application.config.internal_api = {
+  username_path: "/var/lib/misc/infra-secrets/velum-internal-api-username",
+  password_path: "/var/lib/misc/infra-secrets/velum-internal-api-password"
+}
+
+if File.exist?(Rails.application.config.internal_api[:username_path]) &&
+    File.exist?(Rails.application.config.internal_api[:password_path])
+  Rails.application.secrets.internal_api = {
+    username: File.read(Rails.application.config.internal_api[:username_path]).strip,
+    password: File.read(Rails.application.config.internal_api[:password_path]).strip
+  }
+end
--- a/app/controllers/internal_api_controller.rb
+++ b/app/controllers/internal_api_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# InternalApiController is the superclass of all internal API controllers.
+class InternalApiController < ActionController::Base
+  include Api
+
+  before_action :authenticate_request
+
+  private
+
+  def authenticate_request
+    authenticate_or_request_with_http_basic("Velum Internal API") do |username, password|
+      username == Rails.application.secrets.internal_api[:username] &&
+        password == Rails.application.secrets.internal_api[:password]
+    end
+  end
+end
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -21,6 +21,25 @@
       "note": ""
     },
     {
+      "warning_type": "Cross-Site Request Forgery",
+      "warning_code": 7,
+      "fingerprint": "b5774d21ea475ca65a2589e83898ea1571a5ce810e3715f8af7fb9638057b842",
+      "check_name": "ForgerySetting",
+      "message": "'protect_from_forgery' should be called in InternalApiController",
+      "file": "app/controllers/internal_api_controller.rb",
+      "line": 4,
+      "link": "http://brakemanscanner.org/docs/warning_types/cross-site_request_forgery/",
+      "code": null,
+      "render_path": null,
+      "location": {
+        "type": "controller",
+        "controller": "InternalApiController"
+      },
+      "user_input": null,
+      "confidence": "High",
+      "note": ""
+    },
+    {
       "warning_type": "SSL Verification Bypass",
       "warning_code": 71,
       "fingerprint": "fcd6d876d77ab745b6d968b6e91bd07a4390469664220843227a10672aad89f4",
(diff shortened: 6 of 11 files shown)


app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+               bind_dn: Pillar.value(pillar: :ldap_bind_dn),
+               bind_pw: Pillar.value(pillar: :ldap_bind_pw),
+               domain: Pillar.value(pillar: :ldap_domain),
+               group_dn: Pillar.value(pillar: :ldap_group_dn),
+               people_dn: Pillar.value(pillar: :ldap_people_dn),
+               base_dn: Pillar.value(pillar: :ldap_base_dn),
+               admin_group_dn: Pillar.value(pillar: :ldap_admin_group_dn),
+               admin_group_name: Pillar.value(pillar: :ldap_admin_group_name),
+               tls_method: Pillar.value(pillar: :ldap_tls_method),
+               mail_attribute: Pillar.value(pillar: :ldap_mail_attribute)
```
MaximilianMeister: although it's probably quite efficient already, we're doing many db queries here, do you think it would enhance performance a bit if we cache all pillars in a variable and generate the response from that variable?

app/controllers/concerns/api.rb
```diff
@@ -0,0 +1,25 @@
...
+  extend ActiveSupport::Concern
+
+  def ok(status: 200, content: nil, root: nil)
+    if content.blank?
+      render status: status, nothing: true
+    else
+      respond_to do |format|
+        format.json { render status: status, json: content.as_json }
+        format.xml  { render status: status, xml: content.to_xml(root: root, skip_types: true) }
+        format.yaml { render status: status, text: content.to_yaml, content_type: "text/yaml" }
```
MaximilianMeister: do we have a specific use case for yml and xml? otherwise i'd be in favor of keeping it thin, meaning json only

ereslibre: Not really, I can remove them yeah... Updating

app/controllers/internal_api/v1/pillars_controller.rb
ereslibre: I was thinking in improving this, yes. The pillar would invalidate the cache when anything is saved. Updating.

app/models/pillar.rb
```diff
@@ -104,6 +106,7 @@ def set_pillar(key:, pillar_key:, value:, required_pillars:, errors:)
           errors << "'#{key}' could not be saved#{exp}."
         end
       end
+      Rails.cache.delete "pillars/#{key}"
```
MaximilianMeister: i'd put this in an `ensure` block just to be sure that it is always invalidated, even when this method throws an exception for whatever reason

ereslibre: Good catch! Updated.

robdaemon:
I have a few concerns with this:

1. Scaling. This is going to increase the number of concurrent users on the Rails app. Frankly, MySQL scales better than Rails in our current use-case. While it is possible to make a Rails app scale, it's a very involved process and we haven't even started it. We have no framework for scalability testing / tuning.
1. Performance / GC pressure. Rails creates a new instance of a controller for each request. This generates GC pressure on the MRI GC. 
1. APIs should be delivered spec first, using something like Swagger. This will help ensure that we properly version APIs and clients in many languages can easily generate stubs to interact. This is important for both internal and external APIs
1. There was no final decision to use Ruby for this API - it seems that we skipped a step? POC to final PR?
1. Is this intended for v2?

ereslibre:
> Scaling. This is going to increase the number of concurrent users on the Rails app. Frankly, MySQL scales better than Rails in our current use-case. While it is possible to make a Rails app scale, it's a very involved process and we haven't even started it. We have no framework for scalability testing / tuning.

The memory cache that has been added on this PR should help to not go to MySQL every time when nothing changed. I haven't tested Rails performance in our environment to assure how good/bad it scales for our use case.

> Performance / GC pressure. Rails creates a new instance of a controller for each request. This generates GC pressure on the MRI GC.

We'd need  to properly measure this in order to know if it's an issue for us.

> APIs should be delivered spec first, using something like Swagger. This will help ensure that we properly version APIs and clients in many languages can easily generate stubs to interact. This is important for both internal and external APIs

I agree Swagger is something we should provide for public API's, but please, keep in mind this `internal-api` is meant for being consumed from salt only (this `internal-api` won't be consumed by anything we don't maintain).

> There was no final decision to use Ruby for this API - it seems that we skipped a step? POC to final PR?

No decision has been made about the API (public nor internal-api). However, we have important blocked tasks because our current pillar override is very limited, and cannot provide certain complex structures required for certain use cases (insecure registries or registries with customer provided certificates). We needed to unblock this situation, and it's why this PR only includes this small bit.

> Is this intended for v2?

I fear it's too late for v2, but might make it for a maintenance update in the v2 lifetime.

robdaemon:
> The memory cache that has been added on this PR should help to not go to MySQL every time when nothing changed. I haven't tested Rails performance in our environment to assure how good/bad it scales for our use case.

This brings up cache invalidation issues. We must be thinking about an HA admin node, where there is more than once instance of Velum. Right now, the entire admin node is a single point of failure.

> I agree Swagger is something we should provide for public API's, but please, keep in mind this internal-api is meant for being consumed from salt only (this internal-api won't be consumed by anything we don't maintain).

This is not a distinction that should be made. An API is an API is an API. Versioning and proper support of said API is important no matter who the consumer of it is. This is making a long term assumption that Salt is the only consumer of this API, and that's not something I'm willing to place a bet on.

robdaemon:
> We needed to unblock this situation, and it's why this PR only includes this small bit.

Then let's have the discussion. If this merges, it circumvents any discussion.

ereslibre:
> This brings up cache invalidation issues. We must be thinking about an HA admin node, where there is more than once instance of Velum. Right now, the entire admin node is a single point of failure.

The invalidation is handled by the Pillar model. When a pillar is changed, its key is invalidated, so it will be read from MySQL and cached again until another change happens.

The memory cache is something that works wonders in our setup (even in the future with HA), as it's kept for each rails instance (each instance warms up at its on pace). So on HA deployment it would still work, and we wouldn't need an extra dependency like memcached (though, we could add it if we want).

> This is not a distinction that should be made. An API is an API is an API. Versioning and proper support of said API is important no matter who the consumer of it is. This is making a long term assumption that Salt is the only consumer of this API, and that's not something I'm willing to place a bet on.

I can agree on this, but shall we block this because of swagger missing? This PR (along with its dependencies) it's tackling a shortcoming on the pillar that we need to continue with really basic features. Would it suffice if we create an issue and work on it ASAP?

robdaemon:
> The invalidation is handled by the Pillar model. When a pillar is changed, its key is invalidated, so it will be read from MySQL and cached again until another change happens.

Only on a single instance of Velum. It's not going to scale once we make an HA admin node, which needs to happen very, very soon. In the case of multiple instances of velum (as in, multiple isolated containers), there is nothing to invalidate the cache on all running instances. You'll end up serving inconsistent data.

> I can agree on this, but shall we block this because of swagger missing? This PR (along with its dependencies) it's tackling a shortcoming on the pillar that we need to continue with really basic features. Would it suffice if we create an issue and work on it ASAP?

**Yes**. "Temporary" solutions have a way of becoming permanent.

ereslibre:
> Only on a single instance of Velum. It's not going to scale once we make an HA admin node, which needs to happen very, very soon. In the case of multiple instances of velum (as in, multiple isolated containers), there is nothing to invalidate the cache on all running instances. You'll end up serving inconsistent data.

Sorry, you are absolutely right.

ereslibre:
PR has been rebased.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+        group_dn:         pillar["ldap:group_dn"],
+        people_dn:        pillar["ldap:people_dn"],
+        base_dn:          pillar["ldap:base_dn"],
+        admin_group_dn:   pillar["ldap:admin_group_dn"],
+        admin_group_name: pillar["ldap:admin_group_name"],
+        tls_method:       pillar["ldap:tls_method"],
+        mail_attribute:   pillar["ldap:mail_attribute"]
+      }
+    }
+    # rubocop:enable Style/AlignHash
```
flavio: I would personally prefer to move that to a dedicated view instead of embedding a whole json object into the controller.

However this is just a matter of personal taste.

config/initializers/internal_api.rb
```diff
@@ -0,0 +1,15 @@
...
+  username_path: "/var/lib/misc/infra-secrets/velum-internal-api-username",
+  password_path: "/var/lib/misc/infra-secrets/velum-internal-api-password"
+}
+
+if File.exist?(Rails.application.config.internal_api[:username_path]) &&
+    File.exist?(Rails.application.config.internal_api[:password_path])
+  Rails.application.secrets.internal_api = {
+    username: File.read(Rails.application.config.internal_api[:username_path]).strip,
+    password: File.read(Rails.application.config.internal_api[:password_path]).strip
+  }
```
flavio: I'm fine reading the values from a text file, however I would prefer to have also a way to override them through an environment variable. I want to minimize the dependencies on files stored on the host and injected into the containers.

I think we should do something more similar to what Portus is doing.

ereslibre: Right, will fix.

app/controllers/internal_api/v1/pillars_controller.rb
MaximilianMeister: if it grows more I think this is a good idea. we could even use something like jbuilder, but for now i think it's still fine.

MaximilianMeister:
@ereslibre could you rebase to retrigger travis?

ereslibre:
Rebased.

ereslibre:
Rebased.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -0,0 +1,49 @@
+# Serve the pillar information
+class InternalApi::V1::PillarsController < InternalApiController
+  def show
+    pillar = Hash[Pillar.pluck :pillar, :value]
+    # rubocop:disable Style/AlignHash
+    ok content: {
```
MaximilianMeister: this structure could be derived from `Pillar.all_pillars` so we wouldnt have to duplicate and maintain this rather big structure 2 times, which is error prone:

```ruby
ok content: {}.tap do |hash|
  Hash[Pillar.pluck :pillar, :value].each do |key, value|
    hash.deep_merge! key.split(":").reverse.inject(value) { |a, n| { n => a } }
  end
end.deep_symbolize_keys
```

this however doesnt include nil values, in my case e.g. `proxy[:http]` will not be rendered. i just hacked it quickly on the console...

ereslibre: Yeah, I prefer to play safe at the moment with this, missing entries might lead to unexpected problems if the salt code doesn't have defaults...

app/models/pillar.rb
```diff
@@ -1,36 +1,5 @@
...
-#
-#   my_service:
-#     extra_ips:
-#       - 127.0.0.1
-#       - 127.0.0.2
-#
-# This is because how we configure our salt master using this table as an external pillar. For
-# further information you can visit:
-#   - https://github.com/kubic-project/velum/blob/master/docs/salt.md
-#   - https://github.com/kubic-project/salt/blob/master/config/master.d/50-returner.conf
```
bear454: Why is all this documentation removed?

ereslibre: The idea is that from now on we are not tied to the pillar model anymore, and the pillar API controller can render now information from other models. The whole idea of collisions as lists were just a way to represent certain structures on plain mysql information with keys because of how salt uses it, but for example you cannot represent a unitary list in any way (it would be a scalar).

Personally I'd like to see the Pillar model dying so we start to use more specific models (maybe a general configuration model still makes sense), but we are not tied to it anymore.

I can bring back this documentation anyways, but since the lists implementation was actually broken (you cannot represent a list of whatever size since a unitary list is not representable) I don't think collisions will work as expected right now (would have to test it).

PR #299, opened 2017-09-05
WIP: Use reactors to notify Velum of important events

ereslibre:
**Work in progress**, please do not review yet.

- [ ] Write consolidation job (even if we lose callbacks because Velum wasn't available at that moment we still have a job in the background making sure that everything is up to date)

Rebased on top of https://github.com/kubic-project/velum/pull/295, to show a better overview.

Depends on:
* https://github.com/kubic-project/salt/pull/209
* https://github.com/kubic-project/caasp-container-manifests/pull/100
* https://github.com/kubic-project/container-images/pull/21

PR #322, opened 2017-10-09
Collect the dashboard hosts external FQDN

kiall:
This should be used by Dex to redirect back to Velum, allowing for scenarios
where the admin node's private IP is not accessible to the outside world - e.g.
as is the case on OpenStack and possibly other cloud environments.

Part of bsc#1062291

Depends-On: https://github.com/kubic-project/salt/pull/251
Depends-On: https://github.com/kubic-project/automation/pull/164

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -65,6 +65,9 @@ class SetupController < ApplicationController
   def bootstrap
     @apiserver = Pillar.value(pillar: :apiserver) ||
       Minion.find_by(role: Minion.roles[:master]).fqdn
+    # TODO: Minion.find_by(role: Minion.roles[:admin]).fqdn ?
+    @dashboard_external_fqdn = Pillar.value(pillar: :dashboard_external_fqdn) ||
+      ""
   end
 
   def do_bootstrap
@@ -138,7 +141,7 @@ class SetupController < ApplicationController
     when "configure"
       [:dashboard]
     when "do_bootstrap"
-      [:apiserver]
+      [:apiserver, :dashboard_external_fqdn]
     end
   end
 
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -37,7 +37,7 @@ class Pillar < ApplicationRecord
 
   scope :global, -> { where minion_id: nil }
 
-  PROTECTED_PILLARS = [:dashboard, :apiserver].freeze
+  PROTECTED_PILLARS = [:dashboard, :apiserver, :dashboard_external_fqdn].freeze
 
   class << self
     def value(pillar:)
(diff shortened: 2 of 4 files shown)


app/controllers/setup_controller.rb
```diff
@@ -65,6 +65,9 @@ def set_roles
   def bootstrap
     @apiserver = Pillar.value(pillar: :apiserver) ||
       Minion.find_by(role: Minion.roles[:master]).fqdn
+    # TODO: Minion.find_by(role: Minion.roles[:admin]).fqdn ?
+    @dashboard_external_fqdn = Pillar.value(pillar: :dashboard_external_fqdn) ||
+      ""
```
ereslibre: The admin as a regular minion is included in https://github.com/kubic-project/velum/pull/299/files#diff-fa4c43e571bbfe7bd1ae914e5794a39fR13 (despite ignored in the UI, but important for checking its update status and other general status -- would allow us to do `Minion.admin.first.fqdn`). With our current code and in this specific version I'd set as placeholder `Pillar.value(pillar: :dashboard)` instead of `""`. In the general case it will be a sensible default, but you can still change it depending on what environment you are going to deploy for (e.g. openstack).

app/models/pillar.rb
```diff
@@ -37,7 +37,7 @@ class Pillar < ApplicationRecord
 
   scope :global, -> { where minion_id: nil }
 
-  PROTECTED_PILLARS = [:dashboard, :apiserver].freeze
+  PROTECTED_PILLARS = [:dashboard, :apiserver, :dashboard_external_fqdn].freeze
```
ereslibre: This is a leftover, `PROTECTED_PILLARS` can be completely removed.

app/models/pillar.rb
```diff
@@ -46,32 +46,33 @@ def value(pillar:)
...
-        ldap_domain:           "ldap:domain",
-        ldap_group_dn:         "ldap:group_dn",
-        ldap_people_dn:        "ldap:people_dn",
-        ldap_base_dn:          "ldap:base_dn",
-        ldap_admin_group_dn:   "ldap:admin_group_dn",
-        ldap_admin_group_name: "ldap:admin_group_name",
-        ldap_tls_method:       "ldap:tls_method",
-        ldap_mail_attribute:   "ldap:mail_attribute"
+        dashboard:               "dashboard",
+        dashboard_external_fqdn: "dashboard_external_fqdn",
```
ereslibre: I'd like to see some nesting here, but cannot come with proper names at the moment. `dashboard:internal` and `dashboard:external` or something like that?

We will have to move away from this flat pillar structure for storing pillar information in the future, maybe it's fine as it is.

ereslibre:
Also, the `oidc_controller.rb` will need some tweaking. Basically on the `client` method adapting the `redirect_uri` using this pillar value on the `url_for` helper, adding the `host` key.

app/models/pillar.rb
kiall: Yea, I originally had `dashboard:external_fqdn` - until I realised `dashboard` can't be both and string and a dict at the same time. Given this should be 2.0, I avoided that giant refactor.

PR #323, opened 2017-10-10
Wait for orchestration result for marking success on minions

ereslibre:
We are marking success depending on highstates, what makes a bit misleading
for the overall orchestration process. At this moment, orchestration will wait
for some tasks to have been completed (like some `kubectl apply` commands, which
apply immediately). We aren't waiting for the global orchestration result, and
this means that all minions might be marked as successful when we haven't really
waited for the overall orchestration to have finished.

This can have important implications, like not waiting for Dex to be completely
ready while the orchestration is waiting for it.

We just mark failed highstates as an early failure, but wait for the orchestration
result to finish and be succesful in order to mark pending nodes as successful.

This will also align with the current restritions in the UI to enable the `kubectl config`
button, since it will wait for the kubernetes master to be ready in order to be
enabled.

Fixes: bsc#1062542

### DIFF
--- a/app/models/salt_handler/minion_highstate.rb
+++ b/app/models/salt_handler/minion_highstate.rb
@@ -29,9 +29,10 @@ class SaltHandler::MinionHighstate
     data = JSON.parse(salt_event.data)
 
     highstate_succeeded = data["success"] && data["retcode"].zero?
-    new_highstate = Minion.highstates[highstate_succeeded ? :applied : :failed]
     # rubocop:disable SkipsModelValidations
-    minion.update_column(:highstate, new_highstate)
+    minion.update_column(:highstate, Minion.highstates[:failed]) unless highstate_succeeded
     # rubocop:enable SkipsModelValidations
+
+    true
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/salt_handler/minion_highstate.rb
```diff
@@ -29,9 +29,10 @@ def process_event
     data = JSON.parse(salt_event.data)
 
     highstate_succeeded = data["success"] && data["retcode"].zero?
-    new_highstate = Minion.highstates[highstate_succeeded ? :applied : :failed]
     # rubocop:disable SkipsModelValidations
-    minion.update_column(:highstate, new_highstate)
+    minion.update_column(:highstate, Minion.highstates[:failed]) unless highstate_succeeded
```
MaximilianMeister: i cant make the connection yet, if `highstate_succeeded` is `true` here, don't we need to `minion.update_column(:highstate, :applied)` here? what if the last `highstate` is still `:failed` ?

ereslibre: The highstate will only mark failure if the highstate failed. The responsible for marking all pending nodes as succeeded (applied) will be the orchestration result. Since we are lacking a proper overall orchestration result in the database structure, this is a way to make sure that nodes are marked as "applied" when the orchestration has finished and has finished successfully.

kiall:
Given Flavio +A'd, and the only diff between the last revision and this is the addition of a bug number to the commit message, we should go ahead and merge this.

PR #324, opened 2017-10-10
bsc#1062247 add new svg logo

MaximilianMeister:
with the full name SUSE® CaaS Platform

bsc#1062247

also the login page and the normal layout header will use the svg

~~there is only still one issue that the navbar in the app and in the login don't share the same css it seems, i have tweaked the image size in the application css but maybe that;'s not the best way to do it. @vitoravelino do you have an idea?~~

updated to `height: 60px` as per @vitoravelino's suggestion

logologin
logoapp

### DIFF
--- a/app/assets/images/logo.svg
+++ b/app/assets/images/logo.svg
@@ -1,42 +1,101 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!-- Generator: Adobe Illustrator 21.0.2, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-     viewBox="0 0 200 80" style="enable-background:new 0 0 200 80;" xml:space="preserve">
-<style type="text/css">
+
+<svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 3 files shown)


vitoravelino:
@MaximilianMeister It doesn't match the previous font family. I'm not sure if this is an issue or not. If not, great! For dex, what I did was convert the svg to png. Remember to set the height as 60px and it will work out of the box when we replace the existent one.

Thank you for this!

MaximilianMeister:
> It doesn't match the previous font family. I'm not sure if this is an issue or not. If not, great!

i think it's ok as it it unified now, before we had one unknown font in the login screen as svg, and another standard font as text in the app. but suggestions for another font are also welcome

> For dex, what I did was convert the svg to png. 

https://github.com/kubic-project/dex-branding/pull/2

> Remember to set the height as 60px and it will work out of the box when we replace the existent one.

thanks for the hint, i dropped the width and added the 60px height and it seems to work

vitoravelino:
> i think it's ok as it it unified now, before we had one unknown font in the login screen as svg, and another standard font as text in the app. but suggestions for another font are also welcome

I think it's okay to merge without matching the previous font-family but it would be nice to check with the branding team just to be sure. I'm not sure how strict SUSE is with this kind of stuff. We can always update it later.

I'll pull the changes locally and review now.

vitoravelino:
Weird... In your screenshots the font is completely different from the one rendered here.

vitoravelino:
> Weird... In your screenshots the font is completely different from the one rendered here.

I think I don't have Roboto installed in my computed and it renders with sans-serif. In the previous one there was no font-family thing. I think it's related to how the svg is exported. I don't understand these things, so can't help.

MaximilianMeister:
> I think I don't have Roboto installed in my computed and it renders with sans-serif. In the previous one there was no font-family thing. I think it's related to how the svg is exported. I don't understand these things, so can't help.

i dont know either, i thought svg is just paths and has no further dependencies...

PR #328, opened 2017-10-11
handle invalid sessions more user friendly

MaximilianMeister:
if the session is not valid anymore, show this information to the
user, and redirect him to the login instead of failing badly

we dont need to raise in the stored_nonce method, as a failed
deletion from the session hash will result in a nil value, meaning
a non matching nonce

@ereslibre this is one small thing i'd like to fix now, as i ran into it today. moving the logics into a library is the next step but can be done post 2.0

### before
invalid-sessionm
in production this will probably look different, but with no clues what went wrong

### after
invalid-nonce


MaximilianMeister:
to fix this, a user needs to log out and log in again, should we add this to the message as well?

MaximilianMeister:
included the patch in #331 as it happens in the same corner

MaximilianMeister:
reopening, as it still needs some rspec love to increase the coverage again. and after all i'd like to keep that separate from https://github.com/kubic-project/velum/pull/331

MaximilianMeister:
rebased

PR #331, opened 2017-10-12
bsc#1062728: correctly render redirect_url

MaximilianMeister:
a url_for renders a url with from wherever you have requested from

make sure to include the known dashboard url in the redirect_url
otherwise dex will render a Bad Request as it doesn't know the
dns entry

bsc#1062728

### DIFF
--- a/app/controllers/oidc_controller.rb
+++ b/app/controllers/oidc_controller.rb
@@ -15,6 +15,9 @@ class OidcController < ApplicationController
   # we'll need to add an "except" or an "only" clause to this.
   skip_before_action :verify_authenticity_token
 
+  # make sure that the request comes from a registered host
+  before_action :verify_host
+
   def new_nonce
     session[:nonce] = SecureRandom.hex(16)
   end
@@ -111,7 +114,16 @@ class OidcController < ApplicationController
       authorization_endpoint: config.authorization_endpoint,
       token_endpoint:         config.token_endpoint,
       userinfo_endpoint:      config.userinfo_endpoint,
-      redirect_uri:           url_for(controller: "oidc", action: "done")
+      redirect_uri:           oidc_done_url
     )
+  end
+
+  def verify_host
+    return true if accessible_hosts.include? request.host
+    sign_out(current_user)
+    redirect_to root_path,
+                alert: "You have been logged out as #{request.host} " \
+                       "is not within the registered hosts. Please access velum from either " \
+                       "#{accessible_hosts.join(" or ")}"
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,4 +21,11 @@ class ApplicationController < ActionController::Base
   def setup_done?
     Pillar.exists? pillar: Pillar.all_pillars[:apiserver]
   end
+
+  def accessible_hosts
+    [
+      Pillar.value(pillar: :dashboard),
+      Pillar.value(pillar: :dashboard_external_fqdn)
+    ]
+  end
 end
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -13,6 +13,8 @@ class DashboardController < ApplicationController
   skip_before_action :force_ssl_redirect, only: :autoyast
   skip_before_action :authenticate_user!, only: :autoyast
   skip_before_action :redirect_to_setup, only: :autoyast
+  # make sure that access comes from a registered host
+  before_action :host_warning
 
   # The index method is provided through the Discovery concern.
   alias index discovery
@@ -90,5 +92,12 @@ class DashboardController < ApplicationController
 
   def failed_assigned_nodes(assigned)
     assigned.select { |_name, success| !success }.keys.join(", ")
+  end
+
+  def host_warning
+    return true if accessible_hosts.include? request.host
+    flash[:alert] = "You are accessing velum from an unregistered host (#{request.host}). " \
+                    "It is advised to access velum via one of the registered hosts " \
+                    "#{accessible_hosts.join(" or ")}"
   end
 end
--- a/spec/controllers/oidc_controller_spec.rb
+++ b/spec/controllers/oidc_controller_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-# rubocop:disable ExampleLength,MultipleExpectations,MessageSpies
+# rubocop:disable ExampleLength,MultipleExpectations,MessageSpies,RSpec/AnyInstance
 RSpec.describe OidcController, type: :controller do
   let(:external_fqdn_pillar) { create(:external_fqdn_pillar) }
 
@@ -10,6 +10,9 @@ RSpec.describe OidcController, type: :controller do
     it "returns a 302 if the orchestration didn't yet finish" do
       VCR.use_cassette("kubeconfig/cluster_not_ready", record: :none) do
         setup_undone
+
+        allow_any_instance_of(OidcController).to receive(:verify_host)
+          .and_return(true)
 
         get :index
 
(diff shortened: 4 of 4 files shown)


MaximilianMeister:
any idea how this rspec error is related to this change?

```
Bootstrap cluster feature Nodes bootstrapping A user check all nodes at once to be bootstrapped
     Failure/Error: expect(page).to have_content("Summary")
       expected to find text "Summary" in "Logout Confirm bootstrap Cluster specific settings External Kubernetes API FQDN External Dashboard FQDN Back SUSE® CaaS Platform 2.0.0+dev | © SUSE Linux 2017"
```

MaximilianMeister:
now after trying to login there is still an issue

invalid-nonce

ereslibre:
> any idea how this rspec error is related to this change?

This is a flaky test, happens sometimes... :(

app/controllers/oidc_controller.rb
```diff
@@ -111,7 +113,7 @@ def client
       authorization_endpoint: config.authorization_endpoint,
       token_endpoint:         config.token_endpoint,
       userinfo_endpoint:      config.userinfo_endpoint,
-      redirect_uri:           url_for(controller: "oidc", action: "done")
+      redirect_uri:           dashboard + oidc_path
```
ereslibre: I'd better use: `url_for(controller: "oidc", action: "done", host: Pillar.value(pillar: :dashboard_external_fqdn))`. Let's use the helper with the `host` parameter, and also let's use the external fqdn for the dashboard, since this is where dex will redirect the user, and this host needs to be resolvable by the user.

ereslibre:
So on one side it's needed to point to a well known valid value or dex will reject to redirect (https://github.com/kubic-project/salt/blob/master/salt/dex/dex.yaml#L73-L76), but on the other side, if we are visiting Velum with other host and then dex redirects us with the dashboard public fqdn the cookie session won't exist in this new domain or host. You'll have to relogin in Velum, and I am not sure if you'll automatically get redirected to `/oidc/done`.

app/controllers/oidc_controller.rb
ereslibre: Or even better, we can use the implicit url helper: `oidc_done(host: Pillar.value(pillar: :dashboard_external_fqdn))`

MaximilianMeister: updated, thanks

flavio:
LGTM, trying to verify it

app/controllers/oidc_controller.rb
```diff
@@ -65,16 +63,13 @@ def done
...
-    if Rails.env.production?
-      response.headers["Content-Disposition"] = "attachment; filename=kubeconfig"
-    end
-    # :nocov:
-
-    render "kubeconfig.erb", layout: false, content_type: "text/yaml"
+    send_data(
+      render_to_string(template: "oidc/kubeconfig.erb", content_type: "text/yaml", layout: false),
+      filename: "kubeconfig"
+    )
```
ereslibre: I couldn't test this, but if this is served as an attachment (`Content-Disposition: attachment`), then phantomjs won't be able to fetch the kubeconfig file on automation project (https://github.com/kubic-project/automation/blob/master/velum-bootstrap/spec/features/02-bootstrap-cluster.rb#L158).

MaximilianMeister: for that reason i stubbed the `send_data` method in `oidc_controller_spec.rb`

haven't tested it though as i have problems with setting up a devenv

kiall: But, the stub won't have any effect when running velum-bootstrap? - The file will be served as an attachment, and the bootstrap won't be able to download it?

MaximilianMeister: yeah good point, so after all it was intended to just render and not to download, but just in development and test env ... i'll drop the commit again

MaximilianMeister:
rebased

MaximilianMeister:
jenkins depends on https://github.com/kubic-project/automation/pull/170

mjura:
This change works fine, but there is further problem
```
OpenIDConnect::Discovery::DiscoveryFailed (Connection refused - Connection refused - connect(2) for "api.test.nc" port 32000 (api.test.nc:32000)):
  app/controllers/oidc_controller.rb:99:in `oidc_config'
  app/controllers/oidc_controller.rb:104:in `client'
  app/controllers/oidc_controller.rb:89:in `authorization_uri'
  app/controllers/oidc_controller.rb:43:in `index'
```

This is because on Admin node
```
### service names ###
127.0.0.1 api api.infra.caasp.local api.test.nc
```
and there is missing haproxy pod. I have updated bug report

monstermunchkin:
[Bug 1062728 - Downloading the kubectl-config redirects to the kube-master and returns "Bad Request" on the second attempt](https://trello.com/c/FiB0kVL7/413-bug-1062728-downloading-the-kubectl-config-redirects-to-the-kube-master-and-returns-bad-request-on-the-second-attempt)

app/controllers/oidc_controller.rb
```diff
@@ -111,7 +111,7 @@ def client
       authorization_endpoint: config.authorization_endpoint,
       token_endpoint:         config.token_endpoint,
       userinfo_endpoint:      config.userinfo_endpoint,
-      redirect_uri:           url_for(controller: "oidc", action: "done")
+      redirect_uri:           oidc_done_url(host: request.base_url)
```
kiall: Isn't this change effectively a noop? I would assume `oidc_done_url()` under the covers does a `url_for(controller: "oidc", action: "done")`, and that `host` defaults to `request.base_url`?

Ideally, we should redirect to `request.base_url` IF `request.base_url  matches to either the dashboard internal fqdn (i.e. the `dashboard` pillar value) or the dashboard external fqdn (i.e. the `dashboard_external_fqdn` pillar value), If `request.host` does not match either of these, then we should redirect to `dashboard_external_fqdn`.

The reasoning here - is Dex will only accept those 2 names (as those are the 2 names we gave it). Anything else will be rejected, so we should try and use whichever name the user has browsed to - and fallback to the external_fqdn

kiall:
Also - the bug referenced here could easily have been caused by the kube-apiserver restarts - which would explain the "worked once, failed the second time" issues.

app/controllers/oidc_controller.rb
ereslibre: I agree, we should set `host` to whatever that was (if it's on the "whitelist" host -- also accepted by dex), otherwise fallback to `Pillar.value(:dashboard_external_fqdn)`.

The only problem I see with this approach and that I never tested is what happens if the session is missing (if we get to the fallback, the Velum session will probably not exist for `dashboard_external_fqdn` host), and I don't know if we will be redirecting automatically to whatever path was requested.

MaximilianMeister: i just tested this, and indeed the session gets invalid. so the user always needs to access from either the host of the `dashboard` or `dashboard_external_fqdn` pillar value.

i already tried to switch from sessions to rails cache but that didnt work that reliable (also in dev we have cache disabled)

should we add this as a warning in the branding, or should we log out the user and send a flash message with the information that it needs to be accessed via the registered one of the 2 fqdn's? or do you have any other idea that's better?

MaximilianMeister:
I have added a filter now that protects 
the user from invalid sessions, by logging out explicitly and explaining how it needs to be done in a flash message.

dexlogout

kiall:
Failing travis CI due to:

> Coverage report generated for RSpec to /home/travis/build/kubic-project/velum/coverage. 702 / 703 LOC (99.86%) covered.
> Coverage (99.86%) is below the expected minimum coverage (100.00%).

There is 1 line of code, somewhere, that's untested ;)

MaximilianMeister:
> There is 1 line of code, somewhere, that's untested ;)

should be fixed now

PR #336, opened 2017-10-13
fix travis

MaximilianMeister:


spec/features/bootstrap_cluster_feature_spec.rb
```diff
@@ -55,7 +55,9 @@
 
       # means it went to the confirmation page
       expect(page).to have_content("Confirm bootstrap")
-      click_button "Bootstrap cluster"
+      fill_in("External Kubernetes API FQDN", with: "some.url")
```
MaximilianMeister: in a dev setup i always saw a default value for the apiserver, so i think this is not strictly necessary, but i'd like to keep it for completeness should the default value change to be empty some day

spec/features/bootstrap_cluster_feature_spec.rb
```diff
@@ -55,7 +55,9 @@
...
+      fill_in("External Dashboard FQDN", with: "some.url")
+      click_on_when_enabled "#bootstrap"
```
MaximilianMeister: this is also not strictly necessary as the button is enabled by default, but i think (in another PR) we should disable it via js when a non-optional value is missing

MaximilianMeister: i have created https://github.com/kubic-project/velum/issues/337 now

PR #344, opened 2017-10-18
[2.0] Fixed not being able to accept new nodes on discovery page

vitoravelino:
Previously the next step after the discovery page was the cluster
bootstrap. It used to take a while and the user was still able to
accept new nodes while the bootstrap was happening. This behavior was
causing the bootstrap process to crash and we had to prevent the user
from accepting new nodes.

After we changed the behaviour of the setup wizard, user can now go
back and forth through the discovery page and change the nodes roles
for example. Now it's safe for the user to accept pending nodes to be
set for bootstrapping.

This fix only changes the behavior of accepting a new pending node
only on the discovery page. The previous rule of preventing a node
to be accepted is still valid for the dashboard page.

Fixes bsc#1063977

(cherry picked from commit e0535ca842862cb88b8cda2edd1fd95950aab642)

Backport of https://github.com/kubic-project/velum/pull/343

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -202,7 +202,7 @@ MinionPoller = {
   renderPendingNodes: function(pendingMinionId, hasPendingStateNode) {
     var acceptHtml;
 
-    if (hasPendingStateNode) {
+    if (hasPendingStateNode && !isSetRolesPage()) {
       acceptHtml = '';
     } else if (hasPendingAcceptance(pendingMinionId)) {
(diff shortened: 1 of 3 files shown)


monstermunchkin:
[Bug 1063977 — Unable to accept pending nodes after returning to bootstrap page](https://trello.com/c/JqRsm8NJ/419-bug-1063977-unable-to-accept-pending-nodes-after-returning-to-bootstrap-page)

vitoravelino:
Closing this in favor of a proper back-end solution.

PR #345, opened 2017-10-19
spec: make coverage folder public (in development)

MaximilianMeister:
this has the big advantage that you can check the results directly
in velum under http(s)://$velum/coverage/index.html

since we switched the official devenv to containers within VM's the
folder is not accessible that easy anymore

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,8 +3,10 @@ require "simplecov"
 require "webmock/rspec"
 require "vcr"
 
-SimpleCov.minimum_coverage 100
-SimpleCov.start "rails"
+SimpleCov.start("rails") do
+  minimum_coverage 100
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
still forgot to add the folder in `.gitignore` ... fixed now

spec/spec_helper.rb
```diff
@@ -4,6 +4,7 @@
 require "vcr"
 
 SimpleCov.minimum_coverage 100
+SimpleCov.coverage_dir "public/coverage"
 SimpleCov.start "rails"
```
ereslibre: Out of simplicity we could convert this three lines to:

```
SimpleCov.start("rails") do
  minimum_coverage 100
  coverage_dir "public/coverage"
end

MaximilianMeister: adapted

PR #346, opened 2017-10-19
When roles are set, set `not_applied` if it's a local assignment.

ereslibre:
If we are setting the role locally only (on the database), set the
minion status to `not_applied`. If we are also setting the remote role
(set on salt too), set the minion status to `pending`.

Since there's no Orchestration model yet, there's no entity to control
this assignments, and Minion model will set one or the other depending
on the `remote` attribute of `assign_role`.

Fixes: bsc#1063977

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -73,9 +73,10 @@ class Minion < ApplicationRecord
   # to salt subsystem if remote is true.
   def assign_role(new_role, remote: false)
     Minion.transaction do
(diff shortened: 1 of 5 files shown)


vitoravelino:
[Bug 1063977 — Unable to accept pending nodes after returning to bootstrap page](https://trello.com/c/JqRsm8NJ/419-bug-1063977-unable-to-accept-pending-nodes-after-returning-to-bootstrap-page)

PR #352, opened 2017-10-20
alert the user when bootstrap failed

MaximilianMeister:
otherwise the user is left without any clue what to do

@vitoravelino could you please have a look if this makes sense, or if it can be done better?

@kiall @ereslibre would it be easy to get the error(s) of the orchestration to be more specific in the flash message what happened? i believe it might be not that trivial to filter the responsible error and then condense it in a message that fits into a flash message...

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -247,6 +247,14 @@ MinionPoller = {
     $("#download-kubeconfig").attr("disabled", !enabled);
   },
 
+  alertFailedBootstrap: function(message) {
+    if ($('.alert-message span').is(':empty')) {
+      $('.alert-message span').append(message);
+    }
+    $('#float-alert').removeClass('collapse');
+    $('.alert-info').addClass('alert-danger').removeClass('alert-info');
+  },
+
   renderDashboard: function(minion) {
     var statusHtml;
     var checked;
@@ -265,6 +273,7 @@ MinionPoller = {
         break;
       case "failed":
         statusHtml = '<i class="fa fa-times-circle text-danger fa-2x" aria-hidden="true"></i>';
+        MinionPoller.alertFailedBootstrap('At least one of the nodes is in a failed state. Please run "supportconfig" on the failed node(s) to gather the logs.');
         break;
       case "applied":
         statusHtml = '<i class="fa fa-check-circle-o text-success fa-2x" aria-hidden="true"></i>';


app/assets/javascripts/dashboard.js
```diff
@@ -244,6 +244,13 @@ MinionPoller = {
     $("#download-kubeconfig").attr("disabled", !enabled);
   },
 
+  alertFailedBootstrap: function(message) {
+    if ($('.alert-message span').is(':empty'))
```
ereslibre: Is missing braces the recommended style?

ereslibre:
> @kiall @ereslibre would it be easy to get the error(s) of the orchestration to be more specific in the flash message what happened? i believe it might be not that trivial to filter the responsible error and then condense it in a message that fits into a flash message...

This is really tricky at the moment (to fetch the relevant information). We can do this when we have https://github.com/kubic-project/velum/pull/315, https://github.com/kubic-project/velum/pull/299 and https://github.com/kubic-project/velum/pull/295.

IMHO with all those tools we will be in a much better position to notify the user *what* went wrong in a very clear way.

app/assets/javascripts/dashboard.js
vitoravelino: Use curly braces always.

MaximilianMeister: fixed

vitoravelino:
I would probably use a red alert if all the nodes or at least 50% + 1 did not bootstrap. Not sure if it's necessary. Just an idea that came up. 😄

MaximilianMeister:
> I would probably use a red alert if all the nodes or at least 50% + 1 did not bootstrap. Not sure if it's necessary. Just an idea that came up.

good point, it depends if the cluster use usable or not i would say. but i think a generic warning that something needs to be investigated is already fine

vitoravelino:
> good point, it depends if the cluster use usable or not i would say. but i think a generic warning that something needs to be investigated is already fine

Makes sense. We can improve that after we get more info about the cluster as mentioned by @ereslibre.

LGTM 😃

flavio:
I feel the "look at at logs to see what went wrong" is a bit of a joke right now. It's really hard to get down them until all the PR mentioned by @ereslibre are merged (and some additional work is done).

However I don't see any harm in having this code merged.

flavio:
@eotchi what do you think about that?

MaximilianMeister:
can we improve the message a bit?

eotchi:
Just an idea, feel free to ignore. What if instead of saying to look at the logs we display instructions on how to run supportconfig and where to find the generated archives until we have something better?

PR #353, opened 2017-10-20
Notify user when not able to fetch data from Velum API

vitoravelino:
We are regenerating Velum's certificate by adding the external
dashboard fqdn/ip to the Velum certificate. For this certificate
to be reloaded we are currently doing a `docker restart` on Velum
while the orchestration is being executed.

This has a bad effect on the user session as he/she is waiting
for the polling to finish. Without them noticing, the polling will
start failing as the certificate changed under them, and they need
to accept it, but ajax requests will just fail because of this.

With this patch we are notifying the user that the browser cannot
connect with Velum API and suggests he/she to try to reload the session.

Fixes bsc#1064641

screenshot 2017-10-23 10 12 17

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -192,7 +192,10 @@ MinionPoller = {
 
         handleBootstrapErrors();
         toggleMinimumNodesAlert();
+        $('.connection-failed-alert').fadeOut(100);
       }
+    }).fail(function() {
+      $('.connection-failed-alert').fadeIn(100);
     }).always(function() {
       // make another request only after the last one finished
       MinionPoller.pollingTimeoutId = setTimeout(MinionPoller.request, 5000);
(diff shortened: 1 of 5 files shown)


ereslibre:
[Bug 1064641 - Velum polling breaks after clicking on "Bootstrap cluster"](https://trello.com/c/ys0gb7Z7/428-bug-1064641-velum-polling-breaks-after-clicking-on-bootstrap-cluster)

MaximilianMeister:
> I'm not very happy about the html_safe calls but...

even if it's not an obvious issue here, `html_safe` is always bad smell (it's similar to `eval`). we probably have under control which strings are displayed in a flash message, but i would also like to avoid it here, even if it means in the end that we don't have a convenient link in the flash message. could we mention to press `F5` instead? or do a window reload or something similar?

rest looks good to me :+1:

vitoravelino:
I've changed the approach of `html_safe`. I added only for this specific message and escaped the `default_url_host`.

monstermunchkin:
[Bug 1064641 — Velum polling breaks after clicking on "Bootstrap cluster"](https://trello.com/c/XUh1T9yM/430-bug-1064641-velum-polling-breaks-after-clicking-on-bootstrap-cluster)

PR #354, opened 2017-10-23
Reuse filter logic for known hosts

ereslibre:
Also, do not repeat the allowed hosts if they match.


vitoravelino:
Rename verify_host to host_warning?

ereslibre:
> Rename verify_host to host_warning?

Kept the name because it's not really a warning in that case, it actually signs you :)

vitoravelino:
> Kept the name because it's not really a warning in that case, it actually signs you out :)

I was on the phone and I didn't realize they were doing different things. Sorry! 😄

PR #355, opened 2017-10-23
[2.0] Notify user when not able to fetch data from Velum API

vitoravelino:
We are regenerating Velum's certificate by adding the external
dashboard fqdn/ip to the Velum certificate. For this certificate
to be reloaded we are currently doing a `docker restart` on Velum
while the orchestration is being executed.

This has a bad effect on the user session as he/she is waiting
for the polling to finish. Without them noticing, the polling will
start failing as the certificate changed under them, and they need
to accept it, but ajax requests will just fail because of this.

With this patch we are notifying the user that the browser cannot
connect with Velum API and suggests he/she to try to reload the session.

Fixes bsc#1064641

(cherry picked from commit d531ee3fa4c883edee9163e9be5452dccc8bbf4c)

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -192,7 +192,10 @@ MinionPoller = {
 
         handleBootstrapErrors();
         toggleMinimumNodesAlert();
+        $('.connection-failed-alert').fadeOut(100);
       }
+    }).fail(function() {
+      $('.connection-failed-alert').fadeIn(100);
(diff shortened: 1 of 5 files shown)


vitoravelino:
[Bug 1064641 - Velum polling breaks after clicking on "Bootstrap cluster"](https://trello.com/c/ys0gb7Z7/428-bug-1064641-velum-polling-breaks-after-clicking-on-bootstrap-cluster)

PR #356, opened 2017-10-25
Improved failure alert message

vitoravelino:
This is an attempt to make the experience better. We thought that
the previous message was too aggressive. Since this message will
always be shown for 2.0 at least, we want to tranquilize the user
that it's an expected behavior.

### DIFF
--- a/app/views/shared/_notifications.html.slim
+++ b/app/views/shared/_notifications.html.slim
@@ -4,4 +4,4 @@
         locals: { messages: value, alert: key, float: flash[:float], klass: "" }
 
 = render template: 'shared/_notification.html.slim',
-    locals: { messages: "Could not connect with Velum API. Please, try #{link_to "reloading the session", root_path}.".html_safe, alert: "alert", klass: "connection-failed-alert", float: false }
+    locals: { messages: "Could not connect with Velum API. If you're bootstrapping for the first time, this is expected due to an update of our certificate. Please, try #{link_to "reloading the session", root_path}.".html_safe, alert: "alert", klass: "connection-failed-alert", float: false }


ereslibre:
I would rephrase "this is expected due to an update of our certificate".

A certificate update will trigger this error, so will do a connection error with Velum (I as a user missed my connectivity while the dashboard is open, Velum went down for whatever reason, or the machine I'm working on is unreachable).

I would keep it if our current status is that something is being worked on (like, we have pending minions that are in pending status -- chances are the certificate was regenerated). Otherwise, I would not mention anything about certificates, only that we lost connection to the Velum API.

ereslibre:
We could check if all minions are in pending status, in which case we're bootstrapping, so we can show the certificate tip bit. Otherwise I'd say we don't.

app/views/shared/_notifications.html.slim
```diff
@@ -4,4 +4,4 @@
         locals: { messages: value, alert: key, float: flash[:float], klass: "" }
 
 = render template: 'shared/_notification.html.slim',
-    locals: { messages: "Could not connect with Velum API. Please, try #{link_to "reloading the session", root_path}.".html_safe, alert: "alert", klass: "connection-failed-alert", float: false }
+    locals: { messages: "Could not connect with Velum API. If you're bootstrapping for the first time, this is expected due to an update of our certificate. Please, try #{link_to "reloading the session", root_path}.".html_safe, alert: "alert", klass: "connection-failed-alert", float: false }
```
mjura: This sounds bad for me again. Instead presenting this message as error, we should ask user kindly to refresh his browser due to certs regeneration. If we can't handle this in other way, this error should stay as exception. It should be also print on more neutral background yellow or even green. Everything in this state makes bad impression during installation and it is too intrusive and from user experience feels really bad.

kiall: We can't, that I know of, make a determination between "certificate changed", and "you unplugged your ethernet cable". Both look identical to us, so the same error will be shown. Giving the user an indication as to why it may have occurred for the case we know everyone will hit is a good thing...

PR #361, opened 2017-10-27
enable remembering users

MaximilianMeister:
fixes bsc#1062184

### DIFF
--- a/db/migrate/20171027115228_add_remember_token.rb
+++ b/db/migrate/20171027115228_add_remember_token.rb
@@ -1,0 +1,5 @@
+class AddRememberToken < ActiveRecord::Migration
+  def change
+    add_column :users, :remember_token, :string, limit: 150, index: true, after: :reset_password_sent_at
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170911102046) do
+ActiveRecord::Schema.define(version: 20171027115228) do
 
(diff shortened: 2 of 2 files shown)


db/migrate/20171027115228_add_remember_token.rb
```diff
@@ -0,0 +1,5 @@
+class AddRememberToken < ActiveRecord::Migration
+  def change
+    add_column :users, :remember_token, :string, :limit => 150, :index => true
```
ereslibre: I think we could place the column in a better place: `after: :reset_password_sent_at`, what do you think?

MaximilianMeister: no strong opinion on that, i updated it

ereslibre: The schema has not been updated :)

MaximilianMeister: the schema is not used in production anyways. and i ran the migration again and the result is the same in `schema.rb`

ereslibre: If you had already ran it yes, schema won't change. If you rerun the migration you'll see the column at the right position in the schema.rb file. 

We do use it the first time Velum runs (we do a load schema).

ereslibre: BTW just saw that the syntax of the hash is not unified, maybe that's the issue.

MaximilianMeister: i rolled back and then re-ran it of course. but still there was no change. the hash syntax i dont think is the issue, but i can try again

PR #374, opened 2017-11-03
js: fixed undesired minimum nodes alert blink

vitoravelino:
Moved the calls of `toggleMinimumNodesAlert` and
`handleBootstrapErrors` to be done after the checkbox
changes to `true` instead of by any change (false/true).

With this change there will be no middle state that cause
the alert when `toggleMinimumNodesAlert` is called.

Fixes bsc#1066371


vitoravelino:
[Bug 1066371 - Warning message is displayed for split second](https://trello.com/c/Xtm9aSnK/438-bug-1066371-warning-message-is-displayed-for-split-second)

PR #377, opened 2017-11-03
ux: changed color of pending status icon

vitoravelino:
Changed from green to gray. The reason behind this is that green
is usually associated with a success operation and gray is a
more neutral color that doesn't add much meaning.

This change was a quick suggestion from Cynthia when I showed her Velum. 

### before

screenshot 2017-11-03 16 21 13

### after

screenshot 2017-11-03 16 19 41


flavio:
Heads up for @kravciak, this might break openQA.

PR #381, opened 2017-11-17
[devenv] we need to query production database in the devenv

MaximilianMeister:
in the devenv we set RAILS_ENV to development


MaximilianMeister:
not so sure anymore that this is really needed after all, as we point to the production db via `VELUM_DB_NAME` in https://github.com/kubic-project/velum/blob/master/config/database.yml#L17 ... closing

PR #384, opened 2017-11-21
[2.0] js: fixed alert color being overridden

vitoravelino:
The alert notifying the user that the bootstrap has failed was
overriding the type of any previous alert that were on the page.
This was causing some confusing to some users.

Now it creates an alert everytime we check if there's a failed
bootstrapped node without interfering other alerts.

Also removed unnecessary stuff from the notification partial.

Fixes bsc#1069258

(cherry picked from commit d63fba6cd8204238893b90bacd4d1eeceec61655)

Backport of #383

### DIFF
--- a/app/assets/javascripts/alert.js
+++ b/app/assets/javascripts/alert.js
@@ -1,0 +1,31 @@
+function showAlert(message, type, klass) {
+  var $container = $('.alerts-container');
+
+  klass = klass || '';
+  icon = type === 'alert' ? 'fa-exclamation-circle' : 'fa-info-circle';
+  type = type === 'alert' ? 'danger' : 'info';
+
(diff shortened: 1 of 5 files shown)


vitoravelino:
[Bug 1069258 - All flash messages switch to red color when bootstrap failed](https://trello.com/c/JBVkSyOP/459-bug-1069258-all-flash-messages-switch-to-red-color-when-bootstrap-failed)

vitoravelino:
I guess we can merge this one also since #383 was merged.

ereslibre:
> I guess we can merge this one also since #383 was merged.

We should discuss if we want this to be included as a maintenance upgrade, I'd say yes, but still needs to be discussed.

flavio:
Yes, we should get it pushed as part of a future maintenance request. Merging.

PR #385, opened 2017-11-22
Increase worker number

ereslibre:
Increase the number of workers. However, `puma` works better in the
backend with something in front of it.

To be applied with https://github.com/kubic-project/salt/pull/312


kiall:
re "To be applied with kubic-project/salt#312" - this to me is a good change anyway. Puma deadlocking it's workload processes to death has bit me a bunch of times... more workers, lasts longer.

Ideally we completely remove Puma ASAP.

ereslibre:
Let's close this one for now until we update puma and see how it behaves.

PR #390, opened 2017-11-29
Update rubocop

ereslibre:
Supersedes: https://github.com/kubic-project/velum/pull/382

PR #392, opened 2017-12-01
downgrade rake again

MaximilianMeister:
it must have accidentally be gotten updated via #390

rubocops Gemfile suggests gem 'rake', '~> 12.0' at this
point in time so we can safely stick to 12.0.0

https://github.com/bbatsov/rubocop/blob/082408/Gemfile#L9


MaximilianMeister:
this is needed otherwise velum is unresolvable in obs

PR #394, opened 2017-12-01
Remove dev vs prod kubeconfig download difference

kiall:
This was needed to workaround capybara + poltergeist being unable to download
files, this workarond is now down in the velum-interactions tooling, rather than
in a shiped part of the codebase.

Depends-On: https://github.com/kubic-project/automation/pull/198


kiall:
This should pass once the dependant PR is merged.

PR #396, opened 2017-12-07
Migrate LDAP passwords

ereslibre:
Fixes: bsc#1071023

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,9 +9,63 @@ class User < ApplicationRecord
 
   devise(*enabled_devise_modules)
 
+  before_create :encrypt_password
   before_create :create_ldap_user
 
-  protected
+  def after_ldap_authentication
+    return true if encrypted_password.present?
+    encrypted_password = BCrypt::Password.create current_password, cost: 11
+    # rubocop:disable Rails/SkipsModelValidations
+    update_column :encrypted_password, encrypted_password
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -136,7 +172,7 @@ def create_ldap_user
       cn:           "A User",
       objectclass:  ["person", "inetOrgPerson"],
       uid:          uid,
-      userPassword: (password.blank? ? "{CRYPT}#{encrypted_password}" : password),
+      userPassword: (Rails.env.test? ? password : "{CRYPT}#{encrypted_password}"),
```
ereslibre: This is something I haven't been able to solve on Travis with the `slapd` instance we are running there and its configuration. Using `{CRYPT}#{encrypted_password}` will make `spec/features/auth/login_feature_spec.rb` to fail on a couple of specs (the ones related to logging in correctly).

I am just workarounding it this way since I have been trying to find out what's going on for some time now.

MaximilianMeister: ideally this explanation would go into the commit message, maybe in an extra commit?

MaximilianMeister: or maybe even a comment inline... not sure. but we at least would have some clue next time we look at it and wonder

ereslibre:
@MaximilianMeister updated to include a comment.

ereslibre:
@MaximilianMeister rebased.

PR #400, opened 2017-12-14
Reject `nil` values on the Velum pillar.

ereslibre:
This gives a chance for default salt pillars to be used.


ereslibre:
@inercia approved too. I only did a rebase. Merging.

PR #405, opened 2017-12-21
log to disk as well

MaximilianMeister:
in case container logs are gone we still have a backup


config/puma.rb
```diff
@@ -38,3 +38,8 @@
 # end
 
 bind "unix:///var/run/puma/#{SOCKET_NAME}"
+
+stdout_redirect(
+  "/srv/velum/log/#{ENV.fetch("RAILS_ENV")}.log",
+  "/srv/velum/log/#{ENV.fetch("RAILS_ENV")}.log"
```
ereslibre: Maybe we can make this one `"/srv/velum/log/#{ENV.fetch("RAILS_ENV")}.err"` ?

MaximilianMeister: updated

ereslibre:
This will redirect the output completely right? So if we check the `stdout`/`stderr` of the dashboard its output will be blank? I think it's better to keep it as it is (without this patch) if that's the case. A file inside the container once it's dead it's of not much use while `docker logs` even when the container is dead is far more accessible.

PR #406, opened 2017-12-22
Consistent representation of public cloud instance types

bear454:
In order to present a reasonable set of choices for instance_type, acollection of instance types was hand-picked for each supported public cloud framework.

These will be presented in the bootstrap UI, based on design feedback from the SUSE UI/UX team.

### DIFF
--- a/lib/velum/instance_type.rb
+++ b/lib/velum/instance_type.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+require "json"
+
+module Velum
+  # Describe a public cloud instance type
+  class InstanceType
+    attr_reader :sort_value, :key, :category, :vcpu_count, :ram_bytes,
+      :ram_si_units, :details
+
+    def initialize(key, raw_data)
+      @key = key
+      @sort_value = raw_data["sort_value"]
+      @category = Velum::InstanceCategory.new(raw_data["category"])
+      @vcpu_count = raw_data["vcpu_count"]
+      @ram_bytes = raw_data["ram_bytes"]
+      @ram_si_units = raw_data["ram_si_units"]
+      @details = raw_data["details"]
+    end
+
+    def <=>(other)
+      sort_value <=> other.sort_value
+    end
+
+    class << self
+      def for(cloud)
+        raw_collection = get_raw_collection(cloud)
+        instance_type_collection_factory(
+          raw_collection["instance_types"],
+          raw_collection["categories"]
+        )
+      end
+
+      private
+
+      def get_raw_collection(cloud)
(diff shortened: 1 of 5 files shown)


bear454:
@ereslibre can you please look at the failures in Travis? They look completely unrelated to me...

vitoravelino:
@bear454 Lack of test coverage as stated in https://travis-ci.org/kubic-project/velum/jobs/324374573#L1405.

lib/velum/instance_type.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+require "json"
+require "filesize"
```
MaximilianMeister: is this used somewhere?

bear454:
@MaximilianMeister ... please review

lib/velum/instance_type.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def self.for(cloud)
+      raw_collection = get_raw_collection(cloud)
+      instance_type_collection_factory(
+        raw_collection["instance_types"],
+        raw_collection["categories"]
+      )
+    end
+
+    def self.get_raw_collection(cloud)
```
MaximilianMeister: i would not expose any other method than `self.for` through the interface. can we make them private?

bear454:
There is a user interface that uses these types; it's in the _next_ PR.

lib/velum/instance_type.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def self.for(cloud)
+      raw_collection = get_raw_collection(cloud)
+      instance_type_collection_factory(
+        raw_collection["instance_types"],
+        raw_collection["categories"]
+      )
+    end
+
+    private
```
MaximilianMeister: i think in you need to write it like this, otherwise the `private` has no effect:

```ruby
    class << self
      def for(cloud)
        raw_collection = get_raw_collection(cloud)
        instance_type_collection_factory(
          raw_collection["instance_types"],
          raw_collection["categories"]
        )
      end

      private

      def get_raw_collection(cloud)
        data_path = Rails.root.join("config", "data", "#{cloud}-types.json")
        JSON.parse(File.read(data_path))
      end

      def instance_type_collection_factory(raw_types, raw_categories)
        raw_types.collect do |key, values|
          values["category"] = raw_categories[values["category"]]
          InstanceType.new(key, values)
        end.sort!
      end
    end
```

bear454: Yeah, I saw the rubocop complaint; already pushed a change.

PR #407, opened 2017-12-27
[WIP] SUSE registry mirror setting

vitoravelino:


app/models/pillar.rb
```diff
@@ -41,7 +41,9 @@ def all_pillars
         ldap_admin_group_name:   "ldap:admin_group_name",
         ldap_tls_method:         "ldap:tls_method",
         ldap_mail_attribute:     "ldap:mail_attribute",
-        cloud_framework:         "cloud:framework"
+        cloud_framework:         "cloud:framework",
+        suse_mirror_url:         "cloud:suse_mirror_url",
+        suse_mirror_certificate: "cloud:suse_mirror_certificate"
```
vitoravelino: No idea if these names are good or if it makes sense here.

app/assets/javascripts/application.js
```diff
@@ -13,6 +13,7 @@
 //= require jquery
 //= require jquery_ujs
 //= require bootstrap.min
+//= require_tree ./setup
```
ereslibre: Doesn't `require_tree .` include this?

app/assets/javascripts/setup/mirror.js
```diff
@@ -0,0 +1,75 @@
+(function (window) {
+  var dom = {
+    MIRROR_URL_GROUP: '.mirror-url-group',
+    MIRROR_URL: '#settings_suse_mirror_url',
+    MIRROR_INVALID_INSECURE: '.invalid-insecure',
+    MIRROR_INVALID_URL: '.invalid-url',
+  };
+
+  var urlRegex = /regex/;
```
ereslibre: Leftover?

app/models/pillar.rb
ereslibre: I'd rename them to something like:

```
ldap_mail_attribute:              "ldap:mail_attribute",
suse_registry_mirror_url:         "suse_registry_mirror:url",
suse_registry_mirror_certificate: "suse_registry_mirror:certificate",
cloud_framework:                  "cloud:framework"
```

And also its callers in `setup_controller.rb`. I'd leave `cloud_framework` the last one on the list at least for now.

app/assets/javascripts/application.js
vitoravelino: Yes. I had to restart rails. Thanks!

app/models/pillar.rb
vitoravelino: Will do.

app/assets/javascripts/setup/mirror.js
vitoravelino: Yes.

vitoravelino:
[Add SUSE registry mirror option to the first page of the Velum setup](https://trello.com/c/7hEgGdFy/502-add-suse-registry-mirror-option-to-the-first-page-of-the-velum-setup)

config/initializers/assets.rb
```diff
@@ -9,4 +9,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
 # Rails.application.config.assets.precompile += %w( search.js )
-Rails.application.config.assets.precompile += ["authentication.css"]
+Rails.application.config.assets.precompile += ["authentication.css", "test.js", "test.css"]
```
MaximilianMeister: should we not make this dependent on `Rails.env.test?`

spec/controllers/setup_controller_spec.rb
```diff
@@ -173,6 +173,106 @@
       end
     end
 
+    context "when suse registry mirror is disabled" do
+      let(:no_registry_mirror_settings) do
+        s = settings_params.dup
```
MaximilianMeister: MINOR: could write as well:

```ruby
settings_params.dup.tap do |s|
  ...
end
```

...but it's a matter of taste i guess

config/initializers/assets.rb
ereslibre: Good catch, I think we can even remove this (during development and test there's no need to precompile anything).

vitoravelino: I think we could add this to `environments/test.rb`.

spec/controllers/setup_controller_spec.rb
vitoravelino: I like the block syntax...

vitoravelino:
> regarding the polyfill dependency, could we also use es6_promise_polyfill_rails ? I dont know if it's the same thing, but it would make maintenance easier to include this as a gem

Not the same thing. That's a polyfill only for the es6 promise. We don't have any issues with promise so far. 

The polyfill added is only for the `test` environment and it's a quick workaround that we won't need to maintain in the future when we move to API/SPA approach.

vitoravelino:
> also should we hide the certificate field when the user explicitly adds a http url?

IIRC @flavio mentioned a scenario that a `http` url needed a certificate. He can confirm that and give some insight if that's the case.

app/models/pillar.rb
```diff
@@ -14,34 +14,36 @@ def value(pillar:)
...
+        ldap_domain:                      "ldap:domain",
+        ldap_group_dn:                    "ldap:group_dn",
+        ldap_people_dn:                   "ldap:people_dn",
+        ldap_base_dn:                     "ldap:base_dn",
+        ldap_admin_group_dn:              "ldap:admin_group_dn",
+        ldap_admin_group_name:            "ldap:admin_group_name",
+        ldap_tls_method:                  "ldap:tls_method",
+        ldap_mail_attribute:              "ldap:mail_attribute",
+        suse_registry_mirror_url:         "suse_registry_mirror:url",
+        suse_registry_mirror_certificate: "suse_registry_mirror:certificate",
```
inercia: Please change it to `"suse_registry_mirror:cert"` instead of `"suse_registry_mirror:certificate"`

flavio:
> IIRC @flavio mentioned a scenario that a http url needed a certificate. He can confirm that and give some insight if that's the case.

Not really, only `https` needs a certificate to be specified.

flavio:
as @ereslibre said:

> Validating the mirror URL on the controller would be nice 

I agree with that. We should start to move the checks on the backend, that will make everything more robust when we move to a API/SPA mode.

flavio:
I would extend the UI to include some help/context messages.

As for the URL I would add something like:

> This is the URL of the local registry configured to be a mirror of the SUSE container registry.
> Use the `http://` or `https://` prefixes depending on how the local mirror has been deployed. Include also the port number if the mirror isn't using a standard one, eg: `https://suse-mirror.local.lan:5000`

We could also show a hint inside of the text box: `https://suse-mirror.local.lan`.

As for the "Certificate" section I would add something like:

> Use this option to provide the self-signed certificate used by your local mirror of the SUSE registry.

I don't know if it's better from a UI/UX POV to use an inline help (like we do on other parts of the page), to just show the text all the time or do a mix.

app/models/pillar.rb
```diff
@@ -14,34 +14,34 @@ def value(pillar:)
...
-        ldap_domain:             "ldap:domain",
-        ldap_group_dn:           "ldap:group_dn",
-        ldap_people_dn:          "ldap:people_dn",
-        ldap_base_dn:            "ldap:base_dn",
-        ldap_admin_group_dn:     "ldap:admin_group_dn",
-        ldap_admin_group_name:   "ldap:admin_group_name",
-        ldap_tls_method:         "ldap:tls_method",
-        ldap_mail_attribute:     "ldap:mail_attribute",
-        cloud_framework:         "cloud:framework"
+        dashboard:                        "dashboard",
```
vitoravelino: I need to revert this.

config/initializers/assets.rb
vitoravelino: What should I do with this?

app/models/docker_registry.rb
```diff
@@ -20,33 +20,36 @@ def apply(settings_params)
...
-      unless service.persisted?
+      unless registry.persisted?
         errors << "Registry url #{url} doesn't match a docker registry pattern"
       end
 
+      if cert.present?
+        certificate = Certificate.find_or_create_by(certificate: cert.strip)
+        errors << "Failed to validate certificate" unless certificate.persisted?
+
+        CertificateService.create(service: registry, certificate: certificate) if certificate.persisted?
```
vitoravelino: Any suggestion on how to fix rubocop offense line too long for this?

app/models/docker_registry.rb
```diff
@@ -20,33 +20,36 @@ def apply(settings_params)
...
-
-      service = DockerRegistry.find_or_create_by(url: url) do |r|
+      registry = DockerRegistry.find_or_create_by(url: url) do |r|
         r.mirror = mirror
-        r.certificate = certificate
       end
 
-      unless service.persisted?
+      unless registry.persisted?
         errors << "Registry url #{url} doesn't match a docker registry pattern"
```
ereslibre: We also probably want to return here, no?

app/models/docker_registry.rb
ereslibre: You can probably use:

```ruby
if certificate.persisted?
  CertificateService.create(service: registry, certificate: certificate)
end
```

config/initializers/assets.rb
ereslibre: I think we can remove both (`test.js` and `test.css`) and everything should work, can you double check that?

app/models/docker_registry.rb
```diff
@@ -20,33 +20,36 @@ def apply(settings_params)
...
       end
 
+      if cert.present?
+        certificate = Certificate.find_or_create_by(certificate: cert.strip)
+        errors << "Failed to validate certificate" unless certificate.persisted?
+
+        CertificateService.create(service: registry, certificate: certificate) if certificate.persisted?
+      elsif registry.certificate.present?
+        registry.certificate_service.destroy
+      end
```
ereslibre: Also, as an improvement, we could try to refactor this a bit to use a transaction and rollback everything if any of the steps breaks, that way we ensure our data is consistent.

app/models/docker_registry.rb
vitoravelino: Yep, to avoid creating certificate stuff.

app/models/docker_registry.rb
vitoravelino: Will try and see if rubocop will complain.

config/initializers/assets.rb
vitoravelino: We can't remove.

```
Asset was not declared to be precompiled in production.
Add `Rails.application.config.assets.precompile += %w( test.js )` to `config/initializers/assets.rb` and restart your server
```

vitoravelino:
> for whatever reason in jenkins, the bootstrap didnt create the registries table

Can someone help me with this?

> coverage has some small hole somewhere still

I ran here and it was 100%. I'll double check anyway.

> but it works just fine, and the spec tests are also great

Thanks!

ereslibre:
The weird thing regarding migrations is that the supportconfig from the admin states:

```
 Status   Migration ID    Migration Name
--------------------------------------------------
(snip)
   up     20171124131432  Create docker registries
   up     20180109145421  Set udp as default flannel backend
   up     20180110112210  Create certificate services
```

vitoravelino:
Fixed coverage.

app/controllers/setup_controller.rb
```diff
@@ -29,17 +31,22 @@ def welcome
     @services_cidr = Pillar.value(pillar: :services_cidr) || "172.24.0.0/16"
     @api_cluster_ip = Pillar.value(pillar: :api_cluster_ip) || "172.24.0.1"
     @dns_cluster_ip = Pillar.value(pillar: :dns_cluster_ip) || "172.24.0.2"
+    @suse_registry_mirror = DockerRegistry.first || DockerRegistry.new
```
flavio: Are we sure `DockerRegistry.first` is always going to return the SUSE registry mirror? Wouldn't be safer to use something like `find_or_initialize_by`?

app/views/setup/settings/_mirror.html.slim
```diff
@@ -0,0 +1,38 @@
...
+      .panel-body
+        .form-group.mirror-url-group
+          = f.label :url, "URL"
+          .input-group
+            = f.url_field :url, value: @suse_registry_mirror.url, class: "form-control", placeholder: "https://suse-mirror.local.lan:5000"
+            span class="input-group-addon"
+              a data-toggle="tooltip" data-placement="left" title="This is the URL of the local registry configured to be a mirror of the SUSE container registry. Use the http:// or https:// prefixes depending on how the local mirror has been deployed. Include also the port number if the mirror isn't using a standard one, eg: https://suse-mirror.local.lan:5000"
+                i class='glyphicon glyphicon-info-sign'
+          span.help-block.invalid-url.hide This is not a valid URL. Please try something like https://registry.local
+          span.help-block.invalid-insecure.hide You are using an insecure mirror address for a secure remote registry
```
flavio: Can you change the message to be: `<strong>Security warning:</strong> you are using...`.

app/controllers/setup_controller.rb
vitoravelino: At this point I don't think that we would have any registry stored in the database. So it would kind of be the same if I use by `find_or_initialize_by(mirror: SUSE_MIRROR_REGISTRY)`.

flavio: I was thinking about a user that goes back to the setup page after the cluster has been deployed and some registries have been added/removed. I would prefer to play safe and change the code to use `find_or_initialize_by`

vitoravelino: User can't go back to the setup pages after bootstrap unless they hack our code.

flavio:
I see you renamed `docker` into the backend (including the db tables) without doing a migration. I think this is fine because we just created these tables. What do the others think about it?

app/controllers/setup_controller.rb
flavio: I see. If you want to do the change go ahead, otherwise leave the code as it is.

ereslibre:
> I see you renamed docker into the backend (including the db tables) without doing a migration. I think this is fine because we just created these tables. What do the others think about it?

I don't see this change, but in any case as long as it wasn't released I think it's fine to mess with this kind of thing during the development phase (as long as there's no release in the meantime).

flavio:
Vitor reverted the change. I say let's make it with another PR, let's stay focused on merging this code.

config/initializers/assets.rb
ereslibre: Ouch, yes, we can move it to `environments/test.rb` then?

MaximilianMeister:
@vitoravelino could you add a filter for the certificate in https://github.com/kubic-project/velum/blob/master/config/initializers/filter_parameter_logging.rb ?

vitoravelino:
> @vitoravelino could you add a filter for the certificate in https://github.com/kubic-project/velum/blob/master/config/initializers/filter_parameter_logging.rb ?

Done.

vitoravelino:
@MaximilianMeister shoudn't the json output something like:

```json
{
  "registries": [
    {
      "Prefix": "https://registry.suse.com",
      "Mirrors": [
        {
          "URL": "https://docker-testing-registry.suse.de"
        }
      ]
    }
  ]
}
```
/cc @vrothberg @inercia

MaximilianMeister:
> @MaximilianMeister shoudn't the json output something like:

yes, i am looking at that right now, also the certificates were not written to disk, but it's a salt issue, this PR is fine!

PR #410, opened 2018-01-12
packaging: enhancements

MaximilianMeister:
* drop removal of .orig files
  we should keep patches rebased instead
* rebase patches
* drop empty scriptlets
  this was feedback from factory submission
* reduce verbosity
  this massively reduces the build log


flavio (commented):
LGTM

PR #414, opened 2018-01-15
WIP: follow up on mirrors

MaximilianMeister:


app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -22,9 +22,9 @@ def pillar_contents
   def registry_contents
     registries = DockerRegistry.all.map do |reg|
```
vitoravelino: You need to fetch all the registries with `mirror: nil`.

MaximilianMeister: i think here we need all registries. also the ones without a mirror, as this would be also a valid option on the `daemon.json` too

vitoravelino: Nope. IIRC you cannot have the same URL being used as `Prefix` and `URL` inside `daemon.json`. @vrothberg could confirm that.

vitoravelino:
The uniquess of `DockerRegistry` must be a combination of `url` and `mirror` plus it can't accept a registry with an url or mirror that was previously used by another `DockerRegistry`. The second case only applies if what I said earlier [here](https://github.com/kubic-project/velum/pull/414#discussion_r161559481) is true.

MaximilianMeister:
follow up https://github.com/kubic-project/velum/pull/415

PR #419, opened 2018-01-18
Add migration to change the pillar value column to text

ereslibre:
This migration changes the pillar `value` column from `string` to `text` so we keep in sync with the 2.0 branch where we need this column to allow bigger contents (to store certificates, for example).

### DIFF
--- a/db/migrate/20180118103201_change_pillar_value_to_text.rb
+++ b/db/migrate/20180118103201_change_pillar_value_to_text.rb
@@ -1,0 +1,8 @@
+class ChangePillarValueToText < ActiveRecord::Migration
+  def up
+    change_column :pillars, :value, :text
+  end
+  def down
+    change_column :pillars, :value, :string
+  end
+end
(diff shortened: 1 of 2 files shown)


ereslibre:
[SUSE registry mirror backport](https://trello.com/c/QmGHalOd/563-suse-registry-mirror-backport)

ereslibre:
[SUSE registry mirror backport](https://trello.com/c/QmGHalOd/563-suse-registry-mirror-backport)

ereslibre:
> lgtm but @ereslibre do you think a rebase of the schema.rb patch is needed again?

So, both on `big-value` and `release-2.0` the patch applies cleanly:

```
suse/velum@big-value patch -p1 < packaging/suse/patches/1_set_default_salt_events_alter_time_column_value.rpm.patch
patching file db/schema.rb
```

```
suse/velum@release-2.0 patch -p1 < packaging/suse/patches/1_set_default_salt_events_alter_time_column_value.rpm.patch
patching file db/schema.rb
```

So I think that's all we need, right? If it applies cleanly this way it's going to work when packaging.

PR #423, opened 2018-01-19
adapt devel libraries to tumbleweed

MaximilianMeister:
in tumbleweed it's named mariadb, in sle it's mysql


MaximilianMeister:
still needs to be tested

PR #424, opened 2018-01-22
relax kubeclient requirement

MaximilianMeister:
there is no reason now to restrict it at the moment


MaximilianMeister:
no longer needed anymore,a s we dropped kubeclient

PR #427, opened 2018-01-24
Add fallback link to fetch kubeconfig if redirect fails

colstrom:
This allows a user to proceed, even if the JS redirect does not execute.


kiall:
I've hit start on the CI job - assume it passes (I see no reason it won't) - I will merge..

PR #428, opened 2018-01-26
Fix packaging by forcing the subset of gems to be installed.

ereslibre:
Our `Gemfile` and `Gemfile.lock` should be the same under all environments,
development, test and production. Using an envvar on the Gemfile to decide
the dependency tree might cause deep problems and differences on the
dependency tree between development, test and production. We want to have
the same gem dependency tree on all possible environments, so we don't hit
bugs in production caused by a different dependency tree than the one used
in development and test, as this bugs could only be hit in production and
never reproduced in development and test environments.


packaging/suse/make_spec.sh
```diff
@@ -71,7 +71,7 @@ pushd build/$packagename-$safe_branch/
   export BUNDLE_GEMFILE=$PWD/Gemfile
   cp Gemfile.lock Gemfile.lock.orig
   bundle config build.nokogiri --use-system-libraries
-  PACKAGING=yes bundle install --retry=3 --no-deployment
+  PACKAGING=yes bundle install --retry=3 --no-deployment --without development test
```
MaximilianMeister: PACKAGING=yes can be dropped too

jordimassaguerpla:
I think I remember. The problem was with the assets. We need the gems needed for building the assets in the build time of the RPM, but we don' t need them in production (in order to not depend on nodejs).

If you can still build the RPM with this commit, then it is fine.

MaximilianMeister:
assets are still excluded, when generating the spec file. so the asset precompilation fails

ereslibre:
This is building locally and it simplifies the gem dependency logic tree differences. Now the `Gemfile` don't depend on environment variables to be set.

PR #429, opened 2018-01-27
Refresh salt pillar after changing external pillar.

bear454:
We've found this to be a necessary step before invoking `salt-cloud`.

### DIFF
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -1,9 +1,15 @@
+require "velum/salt"
+
 # Pillar represents a pillar value on Salt.
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
   validates :value, presence: true
 
   scope :global, -> { where minion_id: nil }
+
+  attr_accessor :refresh
+
+  after_save :refresh_external_pillar, if: :refresh
 
   PROTECTED_PILLARS = [:dashboard, :apiserver, :dashboard_external_fqdn].freeze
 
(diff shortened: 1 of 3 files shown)


ereslibre:
How is this triggered from within the cloud? I find it a bit weird to couple the `Pillar` model to a remote service? I wonder if there isn't a way that you can directly call `saltutil.refresh_pillar` with salt after you have created all the pillars that you need.

bear454:
I references the upstream documentation in [comments](https://docs.saltstack.com/en/latest/topics/development/external_pillars.html#reminder):
> external pillars must be refreshed in order for minions to see any fresh data

Maybe this issues hasn't bitten you guys for any other nodes because of order-of-operations, but in the case of bootstrapping a public cloud cluster the node utilizing the external pillar is the salt-master node. Instead of performing a manual refresh, it made sense to build that in as a behavior of the external pillar class.

The only reason I added a control switch was to prevent altering the default behavior in other use cases. If there is strong objection, I can move this to a manual operation during public cloud bootstrapping, but IMO this is a bug-waiting-to-happen, and some review should be done of all external pillars to see if refresh is an appropriate default behavior.

ereslibre:
@bear454 This is more a salt behavior, so what we do on our case when we orchestrate is to ensure to refresh the pillar as one of the very first steps of the orchestration:

https://github.com/kubic-project/salt/blob/dfd3b8a6a65c7d969466b09a1f20536a525ae42a/salt/orch/kubernetes.sls#L37-L42

This also allows us to decouple the Velum logic from the salt special behavior. That is, e.g. if you write a `Pillar`, trigger `saltutil.refresh_pillar`, restart the admin node, you'll be on the same situation but nothing will trigger this `saltutil.refresh_pillar`, so the safest bet is to ensure that it runs when we are about to orchestrate, hence why this logic is there instead of Velum, where we just save and serve this information.

PR #430, opened 2018-01-27
Update login feature spec

bear454:
I had issues with a failing test that used inconsitent access to the 'Log in' button, so I've updated it to match the rest of the spec.

Additionally, the descriptions didn't make sense in documentation format so I reworded them.

### DIFF
--- a/spec/features/auth/login_feature_spec.rb
+++ b/spec/features/auth/login_feature_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe "Login feature" do
+describe "Feature: login dialog" do
   let!(:user) { create(:user) }
 
   before do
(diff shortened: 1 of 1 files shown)


spec/features/auth/login_feature_spec.rb
```diff
@@ -22,26 +22,26 @@
...
   end
 
-  it "Wrong password results in an error message" do
-    pending("fix the validations")
+  it "shows an error message when using invalid credentials" do
+    # pending("fix the validations")
     fill_in "user_email", with: "foo"
     fill_in "user_password", with: "bar"
-    find("input[type=submit]", match: :first).click
+    click_button("Log in")
```
mjura: I think here is mistake? Space between "Log in"

bear454: No, it's "Log in".

https://github.com/kubic-project/velum/blob/master/spec/features/auth/login_feature_spec.rb#L20

https://github.com/kubic-project/velum/blob/master/app/views/devise/registrations/new.html.slim#L12

kiall: "Log in" is also more correct from an English point of view :)

PR #431, opened 2018-01-28
skip bundler during spec file creation

MaximilianMeister:
a bundler BuildRequire is already hardcoded

other than that it prevents a false strict dependency on bundler, e.g.
if i run make_spec.sh on openSUSE 42.1 it will render bundler 1.10.6
whatever is installed on the system. this happened in our concourse
build container that uses 42.1 as a base system

Follow up of https://github.com/kubic-project/velum/pull/428


MaximilianMeister:
merging, as this will fix the package build

PR #432, opened 2018-01-29
Add `assets` group that will only be loaded when we are compiling the assets

ereslibre:


Gemfile
```diff
@@ -26,7 +26,10 @@ gem "rake", "12.2.1"
 
 gem "sass-rails", "~> 5.0"
 gem "bootstrap-sass"
-gem "uglifier", ">= 1.3.0"
+
+group :assets do
+  gem "uglifier", ">= 1.3.0"
```
MaximilianMeister: what about the sass gems? shouldn't they also be in the assets group?

or if it's just about precompilation, would it make sense to give the group a less generic name like `group :asset_compile`?

ereslibre: Yes, let me improve it.

PR #433, opened 2018-01-29
js: added unique hostnames validation

vitoravelino:
Velum now validates if selected nodes to bootstrap and be added to the cluster
have unique hostnames. The error message appears in an error alert.

### Discover page

screenshot-20180129104830-401x177

### Unassigned page

screenshot-20180130144123-458x90

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -324,7 +324,7 @@ MinionPoller = {
       <tr> \
         <td class="status">' + statusHtml +  '</td>\
         <td><strong>' + minion.minion_id +  '</strong></td>\
-        <td>' + minion.fqdn +  '</td>\
+        <td class="minion-hostname">' + minion.fqdn +  '</td>\
         <td>' + masterHtml + minion.role + '</td>\
       </tr>';
   },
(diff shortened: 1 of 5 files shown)


vitoravelino:
[Unique hostname validation when bootstrapping/adding nodes](https://trello.com/c/EA0RxrpW/585-unique-hostname-validation-when-bootstrapping-adding-nodes)

vitoravelino:
I didn't add validation on the controller level because I didn't know the best way to do it for both bootstrap and unassigned nodes actions.

kiall:
Thanks @vitoravelino - this looks good at a glance, I'll test later today.

kravciak:
I think autoyast installations will be another issue here. IIRC by default hostname is "linux" for all autoyast installations. It might be useful to:
 - add information about new required HOSTNAME=<UNIQUE_HOSTNAME> parameter to bootstrap tips page - it's another linuxrc parameter, so ..netsetup=dhcp autoyast=*** hostname=***..
 - modify autoyast profile to get hostname from dhcp
 - fix documentation

For example in openQA we have ddns&dhcp setup and all ISO instalations get generated hostname, with exception of autoyast - https://openqa.suse.de/tests/1440451#step/stack_add_nodes/4

PR #434, opened 2018-01-30
Install `bundler` after `bundle clean`.

ereslibre:
`bundle clean` also removes stale executables under `bin`. It is fine that
bundler thinks that `bundler` executable is stale, but it's not, we are just
installing it as a separate gem, so do that at the end of the spec, after
we have executed `bundle clean`.

### DIFF
--- a/packaging/suse/velum.spec.in
+++ b/packaging/suse/velum.spec.in
@@ -72,9 +72,6 @@ __PATCHEXECS__
 install -d vendor/cache
 cp %{_libdir}/ruby/gems/%{rb_ver}/cache/*.gem vendor/cache
 
-# install bundler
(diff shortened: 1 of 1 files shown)


ereslibre:
[Fix `master` branch due to some changes on the velum spec file](https://trello.com/c/8dt8GAi9/587-fix-master-branch-due-to-some-changes-on-the-velum-spec-file)

jordimassaguerpla:
"gem install bundler" is needed in line 75 so we can call "bundler install" after that. We should not remove that. Is the package building?

ereslibre:
@jordimassaguerpla no, it's not required because of https://github.com/kubic-project/velum/blob/ab336014349961d3305ac516d678e4ded8a85600/packaging/suse/velum.spec.in#L54. On build time we have it. This gem install is just to let it installed afterwards. 

The package is building and everything is fine now.

PR #435, opened 2018-01-30
dont be strict with gems

MaximilianMeister:
this will help reduce the breakages in factory

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,33 +1,35 @@
 source "https://rubygems.org"
 
-gem "puma"
-gem "rails", "4.2.10"
+gem "puma", "~> 3.11"
+gem "rails", "~> 4.2.10"
+
+gem "rake", "~> 12.2"
+gem "minitest", "~> 5.10"
 
 gem "jbuilder", "~> 2.5"
-gem "jquery-rails"
-gem "slim"
-gem "font-awesome-rails"
+gem "jquery-rails", "~> 4.3"
+gem "slim", "~> 3.0"
+gem "font-awesome-rails", "~> 4.7"
 
 # NOTE: this is no longer needed in Rails 5. See
 # https://github.com/heroku/rails_stdout_logging#rails-5 for instructions on how
 # to transition.
-gem "rails_stdout_logging", group: [:development, :staging, :production]
+gem "rails_stdout_logging", "~> 0.0.5", group: [:development, :staging, :production]
 
 gem "bcrypt", "~> 3.1.7"
-gem "mysql2"
+gem "mysql2", "~> 0.4.10"
 
-gem "gravatar_image_tag"
-gem "devise"
-gem "devise_ldap_authenticatable"
-gem "net-ldap", require: "net/ldap"
+gem "gravatar_image_tag", "~> 1.2.0"
+gem "devise", ">= 4.3"
+gem "devise_ldap_authenticatable", "~> 0.8"
+gem "net-ldap", "~> 0.11", require: "net/ldap"
 
-gem "openid_connect"
-gem "rake", "12.2.1"
+gem "openid_connect", "~> 1.1"
 
 group :assets do
   gem "sass-rails", "~> 5.0"
-  gem "bootstrap-sass"
-  gem "uglifier", ">= 1.3.0"
+  gem "bootstrap-sass", "~> 3.3.7"
+  gem "uglifier", "~> 4.1"
 end
 
 group :development, :test do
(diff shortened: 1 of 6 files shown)


packaging/suse/bundler-dumpdeps
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/ruby
+# vim: set sw=2 sts=2 et tw=80 :
+require 'bundler'
+
+def expand_pessimistic(version)
+  splitted = version.to_s.split('.')
+  sub_version = splitted.slice(0,(splitted.length-1)).join('.')
+  ":#{sub_version} >= #{version}"
+end
```
ereslibre: I find this logic a bit weird, isn't `~> x.y.z` just meaning `>= x.y.z && < x.(y + 1)` ?

Gemfile
```diff
@@ -1,28 +1,28 @@
...
-gem "rails", "4.2.10"
+gem "puma", ">= 3"
+gem "rails", "~> 4.2.10"
+gem "rake", "~> 12.2.1"
 
 gem "jbuilder", "~> 2.5"
-gem "jquery-rails"
-gem "slim"
-gem "font-awesome-rails"
+gem "jquery-rails", ">= 4.3"
```
ereslibre: Where possible, I'd give priority to `~>` instead of `>=`.

packaging/suse/bundler-dumpdeps
jordimassaguerpla: I think expand_pessimistic will return ":x.y >= x.y.z". For example if you have "foo ~> 1.2.3" you will get "foo:1.2 >= 1.2.3"

jordimassaguerpla:
I like it.

We could improve it by using the macro

%__rubygems_bundled_provides

which will output a "Provides: bundled(rubygem(foo)) == version" for each gem since we are not longer explicitly requiring a specific version as we used to, and so this will help tracking what is inside.

However, this macro has not yet been approved:

https://build.opensuse.org/request/show/569569

So I would wait for that.

MaximilianMeister:
> So I would wait for that.

@jordimassaguerpla we could also add it later, i think we are not really blocked by this. what do you think?

also i've disabled rubocop on darix's script, at some point it'll go to https://github.com/openSUSE/ruby-packaging and we can then just fetch it from there

i've built successfully now for SLE and Tumbleweed, with some preferences in the prjconf

MaximilianMeister:
[(5) Relaxing the requirements](https://trello.com/c/hxPdjBAJ/582-5-relaxing-the-requirements)

MaximilianMeister:
@jordimassaguerpla @ereslibre btw having a patch to comment out the dev/test gems during rpmbuild is effectively the same as if we would re-add the `PACKAGING` env. nevertheless i think it's more "elegant" and visible ... (it's not really elegant in the first place i know, but there's no better way unfortunately as we discussed)

jordimassaguerpla:
@MaximilianMeister : Having a patch means having to review it everytime there is an update in the Gemfile. Parsing the Gemfile with the bundler libs or having an env variable will remove the need to review the patch.

MaximilianMeister:
> I think instead of patching the Gemfile to remove the development and test groups, we could try to write a ruby script that uses the bundler gem to get that list and write a new Gemfile from it.

@jordimassaguerpla makes sense, i'll give it a try. but tbh i think rebasing this patch is not a big deal. a rebase in this case is really quickly done (much better than rebasing Gemfile.lock). and we'd have it the same way as in hawk2, so at one point when we have a better rails packaging we are already on the same level for those 2 projects... but nevertheless i'll try sth with the bundler lib.

ereslibre:
> @jordimassaguerpla @ereslibre btw having a patch to comment out the dev/test gems during rpmbuild is effectively the same as if we would re-add the PACKAGING env. nevertheless i think it's more "elegant" and visible ... (it's not really elegant in the first place i know, but there's no better way unfortunately as we discussed)

Yes, in the end our `Gemfile.lock` does not reflect the reality as the real gem dependencies willl be sorted out by IBS/OBS, so the dependency tree changes will exist between development and production, unless we also fix the development environment and we start to use the gems in the images (so at least we develop to the closest form), instead of running a `bundle install` that will get gems from rubygems.

We'd have to do some tricks as we are vendoring the gems inside `/srv/velum/vendor` inside the container, and we mount `/srv/velum` to develop with the code in our machines, so if we could somehow just vendor somewhere else (`/var/lib/velum` or something similar) we would be developing with the same set of gems that will be used in production (sort of).

PR #437, opened 2018-02-05
mirrors and registries

MaximilianMeister:


MaximilianMeister:
[Endpoint to handle registries](https://trello.com/c/Svd3pSA6/558-endpoint-to-handle-registries)

db/schema.rb
```diff
@@ -29,7 +29,7 @@
     t.datetime "updated_at"
   end
 
-  create_table "docker_registries", force: :cascade do |t|
+  create_table "registries", force: :cascade do |t|
```
vitoravelino: Shouldn't this be done via a new migration? This will break people upgrading from 2.0, no?

db/migrate/20171124131432_create_registries.rb
```diff
@@ -1,6 +1,6 @@
-class CreateDockerRegistries < ActiveRecord::Migration
+class CreateRegistries < ActiveRecord::Migration
   def change
-    create_table :docker_registries do |t|
+    create_table :registries do |t|
```
vitoravelino: Shouldn't this be done in a new migration? Isn't it going to break customers upgrading from 2.0?

vitoravelino: Unless we didn't provide a maintenance to 2.0 yet.

MaximilianMeister: we dont have a registry model in 2.0, so i assumed i can still just update the migration

vitoravelino: But we were going to provide a maintenance update because this suse registry feature was needed by the CAP team, no? I'm just not sure if we still have time to make the change like this or not. If an update was already released, then we can't just change the name here.

MaximilianMeister: afaics it shouldnt affect the 2.0 code as we're writing directly to pillar, w/o any registry model

app/models/registry.rb
```diff
@@ -1,8 +1,6 @@
...
-class DockerRegistry < ActiveRecord::Base
-  has_one :certificate_service, as: :service, dependent: :destroy
-  has_one :certificate, through: :certificate_service
-
+# model that represents a registry
+class Registry < ActiveRecord::Base
+  validates :name, presence: true, uniqueness: true
   validates :url, presence: true, uniqueness: true, url: { schemes: ["https", "http"] }
 
   scope :is_mirror, -> { where.not(mirror: nil) }
```
ereslibre: This should be removed as we won't have `mirror` column here.

app/models/registry.rb
```diff
@@ -1,8 +1,6 @@
-# model that represents a docker registry
-class DockerRegistry < ActiveRecord::Base
-  has_one :certificate_service, as: :service, dependent: :destroy
-  has_one :certificate, through: :certificate_service
-
+# model that represents a registry
+class Registry < ActiveRecord::Base
+  validates :name, presence: true, uniqueness: true
```
ereslibre: This should `have_many :mirrors`, no?

ereslibre: It should also have:

```
has_one :certificate_service, as: :service, dependent: :destroy
has_one :certificate, through: :certificate_service
```

As one thing is the certificate the registry has, and another  one the certificate the mirror has.

db/migrate/20171124131432_create_registries.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRegistries < ActiveRecord::Migration
```
ereslibre: I'd rather modify the last migration, since if we go this way this PR is missing two migrations: one to remove the old table, and another one to migrate the SUSE registry data from the `pillars` table and removing it from there.

The first migration I think could be avoided by just modifying the `CreateDockerRegistries` migration (after all, this hasn't been released), and we need an extra migration for the `Pillar` to first-class citizen migration of the SUSE registry mirror.

app/models/mirror.rb
```diff
@@ -0,0 +1,8 @@
+class Mirror < ActiveRecord::Base
```
ereslibre: I don't know what do you think, but I'd prefer to rename them to `ImageMirror` and `ImageRegistry`. I don't have a strong opinion though, while `Registry` sounds good, `Mirror` sounds way too generic. Maybe `Registry` and `RegistryMirror`?

db/migrate/20171124131432_create_registries.rb
```diff
@@ -0,0 +1,9 @@
...
+  def change
+    create_table :registries do |t|
+      t.string :name
+      t.string :url, unique: true
+      t.timestamps
```
ereslibre: This also needs `certificate_service_id`.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -20,24 +20,22 @@ def pillar_contents
...
-          }
-        end
-      }
+    registries = []
+    Registry.all.map do |reg|
+      registry ||= {}
+      registry[:url]  = reg.url
+      registry[:cert] = reg.try(:certificate).try(:certificate)
+      mirrors = RegistryMirror.where(registry_id: reg.id)
+      mirrors.each do |mirror|
```
flavio: Given:
```ruby
class Registry < ActiveRecord::Base
  has_many :registry_mirrors, dependent: :destroy
```

Can this be replaced with `registry.registry_mirrors.each ...`

MaximilianMeister: fixed, thanks!

app/models/registry.rb
```diff
@@ -0,0 +1,61 @@
...
+    def configure_suse_registry(suse_registry_mirror_params)
+      if suse_registry_mirror_params.blank?
+        cleanup
+        return []
+      end
+
+      errors      = []
+      cert        = suse_registry_mirror_params["certificate"]
+      mirror      = suse_registry_mirror_params["mirror"]
+      name        = suse_registry_mirror_params["name"] || "suse"
```
MaximilianMeister: the codeclimate issue is caused by this line here. i suggest we ignore it (as i did with rubocop, which reported the same thing), as there is no better way to do it i think. can we configure codeclimate somehow to let that pass?

and the other Q: do we even need codeclimate when rubocop is testing the same thing?

vitoravelino: I think SUSE in uppercase is better since it will be displayed in a few pages.

app/models/registry.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+
+      errors
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+
+    private
+
+    # remove old registries and mirrors from the db if they were deleted in the UI
+    def cleanup
```
vitoravelino: No need to fetch registries here. We will have only SUSE registry at this point and time and we only need to destroy the associated mirror with it. We can simplify by calling just `RegistryMirror.destroy_all`. Is that right? Or did I miss something?

app/controllers/setup_controller.rb
```diff
@@ -31,17 +29,19 @@ def welcome
     @services_cidr = Pillar.value(pillar: :services_cidr) || "172.24.0.0/16"
     @api_cluster_ip = Pillar.value(pillar: :api_cluster_ip) || "172.24.0.1"
     @dns_cluster_ip = Pillar.value(pillar: :dns_cluster_ip) || "172.24.0.2"
-    @suse_registry_mirror = DockerRegistry.find_or_initialize_by(mirror: SUSE_REGISTRY_URL)
+    @suse_registry_mirror = RegistryMirror.find_or_initialize_by(
```
vitoravelino: You are assuming the `db:seed` will happen, right? Because if the seed doesn't happen, it will initialize a registry without a name. Might break for production env.

vitoravelino: Maybe instead of `registry_id` you could use `name` and `SUSE` because no other registry would exist and remove the db seed.

app/models/registry.rb
vitoravelino: Also, use a constant in `Registry` class and reuse everywhere instead of hard coded.

vitoravelino:
What about the pillars used in v2? How are we handling the upgrade from v2 to v3?

MaximilianMeister:
> What about the pillars used in v2? How are we handling the upgrade from v2 to v3?

good point, we might need to add some ruby migration code to https://github.com/kubic-project/caasp-container-manifests/blob/master/admin-node-setup.sh - i guess it can be as easy as deleting the old pillars and creating the mirror instances via `RegistryMirror.create`

app/models/registry.rb
```diff
@@ -0,0 +1,61 @@
...
+    # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+    def configure_suse_registry(suse_registry_mirror_params)
+      if suse_registry_mirror_params.blank?
+        cleanup
+        return []
+      end
+
+      errors      = []
+      cert        = suse_registry_mirror_params["certificate"]
+      mirror      = suse_registry_mirror_params["mirror"]
```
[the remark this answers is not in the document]
MaximilianMeister: which one? params or local variable?

app/models/registry.rb
MaximilianMeister: yes for now we dont need to be so intelligent, updated

app/models/registry.rb
MaximilianMeister: done

app/controllers/setup_controller.rb
MaximilianMeister: that's the thing, i used a seed because that's what they're there for in rails, do you think this could collide in some way? we run the seeds in `bin/init` so we can assume it ran

vitoravelino: I don't have a strong opinion about this. If we guarantee that the seed won't mess up with things, I guess that's fine.

db/seeds.rb
```diff
@@ -8,11 +8,11 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 def seed_development
```
vitoravelino: Seed is going to be ran every time the container is launched, right? If yes, we can't use `create` here because it will create unnecessary registries.

MaximilianMeister: we can use `first_or_initialize` instead, which is cleaner, but even `create` wont create more registries as there is the uniq constraint

vitoravelino: I didn't see the validation constraint. :sweat_smile:

MaximilianMeister: updated

app/models/registry.rb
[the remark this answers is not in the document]
MaximilianMeister: done

vitoravelino:
> good point, we might need to add some ruby migration code to https://github.com/kubic-project/caasp-container-manifests/blob/master/admin-node-setup.sh - i guess it can be as easy as deleting the old pillars and creating the mirror instances via RegistryMirror.create

Let's wait for @ereslibre review then.

app/models/certificate_service.rb
```diff
@@ -4,4 +4,27 @@ class CertificateService < ActiveRecord::Base
...
+  after_create :update_service_id
+  before_destroy :nullify_service_reference
+
+  private
+
+  def service_type_instance
+    service_type.constantize.find(service_id)
+  end
+
+  def update_service_id
```
vitoravelino: Is this necessary? When we create a model I think Rails is smart enough to associate all the needed ids, no? Which scenario is this needed?

app/models/certificate_service.rb
```diff
@@ -4,4 +4,27 @@ class CertificateService < ActiveRecord::Base
...
+    service_type_instance.tap do |s|
+      s.certificate_service_id = id
+      s.save
+    end
+  end
+
+  def nullify_service_reference
```
vitoravelino: Is this also necessary? If the service is destroyed, the certificate service will also be destroyed since is dependent.

app/models/certificate_service.rb
MaximilianMeister: unfortunately not, could be that it can be done with some model attribute/configuration but i couldnt find anything

app/models/certificate_service.rb
MaximilianMeister: didnt work out of the box so i added this hook

app/models/certificate_service.rb
vitoravelino: Really? Weird...

app/models/certificate_service.rb
vitoravelino: Really? Weird... [2]

app/models/registry.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  SUSE_REGISTRY_NAME = "SUSE".freeze
+  SUSE_REGISTRY_URL  = "https://registry.suse.com".freeze
+
+  class << self
+    # create or update suse Registry model
+    # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+    def configure_suse_registry(suse_registry_mirror_params)
+      if suse_registry_mirror_params.blank?
+        RegistryMirror.where(name: SUSE_REGISTRY_NAME).destroy_all
```
vitoravelino: You don't need to use `where` here but if you want to play safe you need to fetch the registry first, then call `destroy_all` on the `mirrors`.

vitoravelino: Just realized that you were also using `SUSE_REGISTRY_NAME` for the registry mirror, so this would also work.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -20,24 +20,21 @@ def pillar_contents
...
-            url:  reg.url,
-            cert: (reg.certificate.present? ? reg.certificate.certificate : nil)
-          }
-        end
-      }
+    registries = []
+    Registry.all.map do |reg|
+      registry ||= {}
+      registry[:url]  = reg.url
+      registry[:cert] = reg.try(:certificate).try(:certificate)
```
ereslibre: The first `certificate` should always be present, this can be `reg.certificate.try(:certificate)`.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -20,24 +20,21 @@ def pillar_contents
...
-        mirrors: DockerRegistry.where(mirror: remote_registry_url).map do |reg|
-          {
-            url:  reg.url,
-            cert: (reg.certificate.present? ? reg.certificate.certificate : nil)
-          }
-        end
-      }
+    registries = []
+    Registry.all.map do |reg|
+      registry ||= {}
```
ereslibre: It's safe to use `=` here, no?

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -20,24 +20,21 @@ def pillar_contents
...
+      registry[:cert] = reg.try(:certificate).try(:certificate)
+      reg.registry_mirrors.each do |mirror|
+        registry[:mirrors] ||= []
+        registry[:mirrors].push(
+          url:  mirror.url,
+          cert: CertificateService.find_by(id: mirror.certificate_service_id)
+                  .try(:certificate).try(:certificate)
+        )
+      end
+      registries.push(registry)
```
ereslibre: Nitpick, we can do it without having to append to `registries`, no?

```ruby
registries = Registry.all.map do |reg|
  registry = {}
  (snip)
  registry
end
```

app/controllers/setup_controller.rb
```diff
@@ -129,10 +129,12 @@ def suse_registry_mirror_params
     end
 
     if params["settings"]["suse_registry_mirror_enabled"] == "enable"
-      parameters["mirror"] = SUSE_REGISTRY_URL
+      parameters["mirror_url"] = params["settings"]["suse_registry_mirror"]["url"]
+      # not needed anymore
+      parameters.delete :url
```
ereslibre: Do we really need  to remove `url` from the parameters?

app/models/certificate_service.rb
ereslibre: I don't see why we need to do this. In fact, I don't understand why we need the `certificate_service_id` attribute on `Registry` or `RegistryMirror`, we should drop them. This is why this is messing up with us. The association is smart enough to use the proxy model.

app/models/certificate_service.rb
ereslibre: Same for this, we should remove this too.

db/schema.rb
```diff
@@ -79,6 +72,28 @@
   add_index "pillars", ["minion_id"], name: "index_pillars_on_minion_id", using: :btree
   add_index "pillars", ["pillar"], name: "index_pillars_on_pillar", using: :btree
 
+  create_table "registries", force: :cascade do |t|
+    t.string   "name",                   limit: 255
+    t.string   "url",                    limit: 255
+    t.integer  "certificate_service_id", limit: 4
```
ereslibre: This should be removed, the proxy model will take this into account by linking `certificate_id` with a `service_type` and a `service_type` (in this case `Registry` as type and `registries`.`id` as id).

db/schema.rb
```diff
@@ -79,6 +72,28 @@
...
+    t.datetime "updated_at"
+  end
+
+  add_index "registries", ["certificate_service_id"], name: "index_registries_on_certificate_service_id", using: :btree
+
+  create_table "registry_mirrors", force: :cascade do |t|
+    t.string   "name",                   limit: 255
+    t.string   "url",                    limit: 255
+    t.integer  "registry_id",            limit: 4
+    t.integer  "certificate_service_id", limit: 4
```
ereslibre: This should be removed, the proxy model will take this into account by linking `certificate_id` with a `service_type` and a `service_id` (in this case `RegistryMirror` as type and `registry_mirrors`.`id` as id).

db/schema.rb
```diff
@@ -126,4 +141,7 @@
   add_index "users", ["email"], name: "index_users_on_email", unique: true, using: :btree
   add_index "users", ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true, using: :btree
 
+  add_foreign_key "registries", "certificate_services"
+  add_foreign_key "registry_mirrors", "certificate_services"
+  add_foreign_key "registry_mirrors", "registries"
```
ereslibre: Why do we need foreign keys at all?

app/controllers/setup_controller.rb
MaximilianMeister: not technically

db/schema.rb
MaximilianMeister: good point, i dropped them now

ereslibre:
As a side note, do we want the suse registry entry always present? If not, we can remove the seeding and just add a migration to remove it from the `Pillar` and create a `Registry` and `RegistryMirror` of its own.

In any case I think we want to migrate the old mirror if exists on the pillar, and remove the old value from there.

MaximilianMeister:
> As a side note, do we want the suse registry entry always present?

maybe in kubic not, so this might be a good idea to have it in a migration.

app/controllers/setup_controller.rb
vitoravelino: Since you've removed the seeds, you need to change this to use `name` instead of `registry_id`.

MaximilianMeister:
@ereslibre any idea why the suse registry that got added during the last migration is not available in rspec? on the rails console with RAILS_ENV=test it's available

db/migrate/20180221142221_migrate_registry_mirror_from_pillar.rb
```diff
@@ -0,0 +1,43 @@
+class MigrateRegistryMirrorFromPillar < ActiveRecord::Migration
+  def up
+    registry = Registry.create(
```
ereslibre: This registry should only be created if the pillar previously existed, no?

MaximilianMeister: at least on v3 we depend on the suse registry entry to be present always, hence the mirror configuring possibility during bootstrap

MaximilianMeister:
regarding the codeclimate issue i've submitted https://github.com/kubic-project/velum/pull/445

app/controllers/setup_controller.rb
vitoravelino: Still missing this...

db/migrate/20180221142221_migrate_registry_mirror_from_pillar.rb
ereslibre: TBH I am a bit confused with this. I am OK with this `registry` entry being created from the very beginning, but I thought this was going to be created when on the UI we choose to mirror the SUSE registry on premise, instead of precreating this registry without knowing if this will be ever be used. I'm also thinking about kubic.

vitoravelino: AFIAK, user won't be able to do anything regarding the SUSE registry but only about its mirrors. So this might make sense but not sure about exporting it to the docker daemon.json.

app/controllers/setup_controller.rb
vitoravelino: If migration will happen, then we will always have a registry, right? Then this change is not necessary.

ereslibre: No, the migration won't happen on new installations, only on upgrades.

db/migrate/20180221142221_migrate_registry_mirror_from_pillar.rb
MaximilianMeister: hmm good point there, should we skip it in the pillars controller when there's no mirror?

MaximilianMeister: @ereslibre updated to only create the registry on demand

app/controllers/setup_controller.rb
MaximilianMeister: @vitoravelino sorry didnt see the comments....

@ereslibre not sure what you mean, all the migrations are run on new installations as well

ereslibre: No, they won't, and we've played with that in the past. A new installation will basically go through https://github.com/kubic-project/velum/blob/master/bin/init#L47 (`db:schema:load`, which will assume all migrations have been ran until the latest version of the (schema definition)[https://github.com/kubic-project/velum/pull/437/files#diff-1acd2e7e27a227829d5d14a91c863bb6L14]).

MaximilianMeister: maybe i don't get the idea here ;-) but at some point we need to run the migrations, no? if we wouldn't run them we wouldn't have any table available

ereslibre: `db:schema:load` will get the `db/schema.rb` and create all tables in the database, also assuming that all migrations until the version specified in `db/schema.rb` have been run. That's why on new installations migrations won't run (it's a feature, so you on new installations only load the schema from the `db/schema.rb` that is the reflection of the database status)

vitoravelino: I just realized that this is the mirror and not the registry. So this line is fine. Sorry for the confusion! :sweat_smile:

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -24,25 +24,21 @@ def pillar_contents
...
-      }
+    registries = Registry.all.map do |reg|
+      registry = {}
+      registry[:url]  = reg.url
+      registry[:cert] = reg.certificate.try(:certificate)
+      reg.registry_mirrors.each do |mirror|
+        registry[:mirrors] ||= []
+        registry[:mirrors].push(
+          url:  mirror.url,
+          cert: CertificateService.find_by(service_id: mirror.id)
```
ereslibre: I think `mirror.certificate.try(:certificate)` is more readable.

vitoravelino:
My only remaining concern is if the SUSE registry will be created in a new installation or an upgrade when no mirror was set. We need to ensure that it's created because user won't be able to do that via UI.

MaximilianMeister:
> My only remaining concern is if the SUSE registry will be created in a new installation or an upgrade when no mirror was set. We need to ensure that it's created because user won't be able to do that via UI.

it's done in the controller https://github.com/kubic-project/velum/pull/437/files#diff-31a60d13062acc90b6a78b0a563811a8R28

vitoravelino:
> it's done in the controller https://github.com/kubic-project/velum/pull/437/files#diff-31a60d13062acc90b6a78b0a563811a8R28\
>
> upgrade case: https://github.com/kubic-project/velum/pull/437/files#diff-f114bb7ad8e6620c8435ee93071cce31R6

In both cases the mirror was set. What about the case when the mirror was not set?

PR #438, opened 2018-02-06
Adapt velum to build a cluster in EC2

bear454:
* Present a cloud-framework-aware UI (designed with & approved by UI/UX team)
* Persist data back to salt-master via external pillars
* Trigger `salt-cloud` generation of the cluster


bear454:
Will someone please dismiss the [codeclimate issue](https://codeclimate.com/github/kubic-project/velum/pull/438) since it's just a method rename?

app/assets/javascripts/cloud/bootstrap.js
```diff
@@ -0,0 +1,64 @@
+$(function() {
+  // https://stackoverflow.com/questions/10420352/converting-file-size-in-bytes-to-human-readable-string
```
ereslibre: Is it fine to reuse code without an explicit license?

lib/velum/salt.rb
```diff
@@ -34,6 +35,27 @@ def self.update_status(targets: "*", cached: false)
...
+              client: "local_async",
+              tgt:    "admin",
+              fun:    "cloud.profile",
+              arg:    ["cluster_node", instance_name]
+            }
+          )
+        end
+      end
+      threads.each(&:join)
+      threads.collect { |t| t[:response] }
```
ereslibre: Is it really needed to thread this calls? After all we are joining them at the very end, so the request will not see any speedup. Since it's an async client, why not just loop-call `perform_request` without spawning threads?

ereslibre: Or even better, if the looping happened in `cloud.profile` we could just do one single call with `arg: ["cluster_node", list_of_machines]`

ereslibre:
@bear454 we can ignore the codeclimate issue, it's not a blocker for merging (jenkins is though).

app/assets/javascripts/cloud/bootstrap.js
bear454: According to StackOverflow, that code is licenced CC-BY-SA; and has been properly attributed by the comment.

https://meta.stackexchange.com/questions/271080/the-mit-license-clarity-on-using-code-on-stack-overflow-and-stack-exchange

bear454:
@ereslibre can you pinpoint the issue found via jenkins? The output is rather verbose, and contains hundreds of errors that appear (to my unfamiliar eye) to be unrelated.

lib/velum/salt.rb
bear454: I can't find any example of `cloud.profile` being called with a list of machine names, where did you find that usage?

https://docs.saltstack.com/en/latest/topics/cloud/salt.html#cloud-profile

MaximilianMeister:
adding screenshot for reference
ec2

app/models/cloud_cluster.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+require "velum/salt"
+# CloudCluster represents user-configured attributes of a cloud deployment.
+class CloudCluster
+  include ActiveModel::Model
+  attr_accessor :cloud_framework,
+    :instance_count, :instance_type, :instance_type_custom,
+    :subnet_id, :security_group_id # EC2
```
MaximilianMeister: do you think it makes sense to abstract the ec2 part out of this class into it's own mode, and use `CloudCluster` as a base class? i just imaging when adding more cloud providers this might be more organized in the long term, as (at least) the attributes will probably differ

app/models/pillar.rb
```diff
@@ -13,6 +13,10 @@ def value(pillar:)
     end
 
     def all_pillars
+      simple_pillars.merge(cloud_worker_pillars)
+    end
+
+    def simple_pillars
```
MaximilianMeister: MINOR: maybe `simple_pillars` could be called `local_pillars` or `non_pubcloud_pillars` or sth more descriptive?

config/routes.rb
```diff
@@ -37,7 +37,8 @@
   namespace :setup do
     get "/", action: :welcome
     match "/", action: :configure, via: [:put, :patch]
-    get :"worker-bootstrap"
+    get  "worker-bootstrap"
+    post "worker-bootstrap", action: :build_cloud_cluster
```
MaximilianMeister: maybe have a more descriptive route e.g. `trigger-cloud` or sth similar? i guess you could just tell the form to POST to another route via `url:` argument

app/assets/javascripts/cloud/bootstrap.js
ereslibre: Right, thank you.

lib/velum/salt.rb
ereslibre: You are right, didn't know where `cloud.profile` was coming from. Thanks for removing the thread spawning.

app/views/setup/_cluster_size_panel.html.slim
```diff
@@ -0,0 +1,19 @@
...
+    h3.panel-title Cluster size
+  .panel-body.cluster-size-selector
+    .col-md-8
+      .form-group
+        label for="cloud_cluster_instance_count" Number of instances
+        = form.text_field :instance_count, data: {slider_min: 3, slider_value: 5, slider_max: 50, slider_scale: "logarithmic", slider_tooltip: "always", slider_tooltip_position: "bottom", slider_ticks: [3, 5, 10, 20, 50], slider_ticks_snap_bounds: 0.5}
+    .col-md-2.text-right
+      .form-group.cluster-cpu-count
+        label Total vCPUs
+        h3#cluster-cpu-count = 0
```
vitoravelino: I wouldn't use a `h3` here. A single `span` with `display: block|inline-block` would do the trick.

app/views/setup/_cluster_size_panel.html.slim
```diff
@@ -0,0 +1,19 @@
+.panel.panel-default
+  .panel-heading
+    h3.panel-title Cluster size
+  .panel-body.cluster-size-selector
+    .col-md-8
+      .form-group
+        label for="cloud_cluster_instance_count" Number of instances
+        = form.text_field :instance_count, data: {slider_min: 3, slider_value: 5, slider_max: 50, slider_scale: "logarithmic", slider_tooltip: "always", slider_tooltip_position: "bottom", slider_ticks: [3, 5, 10, 20, 50], slider_ticks_snap_bounds: 0.5}
+    .col-md-2.text-right
+      .form-group.cluster-cpu-count
```
vitoravelino: Is `form-group` really necessary?

app/views/setup/_cluster_size_panel.html.slim
vitoravelino: You can use `.h3` class to reproduce the same font-size and color.

app/assets/javascripts/cloud/bootstrap.js
bear454: Just to be on the safe side, I've engaged our legal team to verify.

app/views/setup/_cluster_size_panel.html.slim
bear454: What benefit do you see from using
```
span.#cluster-cpu-count.h3 style="display: inline-block" = 0
```
over
```
h3#cluster-cpu-count = 0
```
?

app/views/setup/_cluster_size_panel.html.slim
bear454: It makes the bootstrap structure simpler, and avoids duplicating the related CSS.

app/models/pillar.rb
bear454: I chose "simple_pillars" because these pillars follow the simplest path from velum -> salt; no munging required like for the cloud or registry pillars.

I expect this name will be short-lived anyway... someone needs to refactor in order to shut up codeclimate ;-)

app/views/setup/_cluster_size_panel.html.slim
vitoravelino: But using this class here only adds an unnecessary margin-bottom that makes the layout a bit strange because there's more space in the bottom than in the top. You can see the difference  when you choose "Other type", this element gets hidden and the spacing gets the same as the top.

app/models/cloud_cluster.rb
bear454: I've been looking at both ways... I know we have a lot of duplication across all the providers we will support - I'm not going to be certain until I add Azure support. So, I'd like to leave it as-is, and make that decision when I hit the problem... for now that's just YAGNI.

app/views/setup/_cluster_size_panel.html.slim
vitoravelino: Visually speaking nothing changes but semantically speaking it's wrong to use header tags around just because it's a block element and has some style associated with it. Same applies for the `label` element above.

Ideally I would prefer a dedicated class with its own style. Maybe something like `cluster-cpu-count-value` since you already used `cluster-cpu-count` for the div that encapsulates it.

config/initializers/assets.rb
```diff
@@ -9,4 +9,7 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
 # Rails.application.config.assets.precompile += %w( search.js )
-Rails.application.config.assets.precompile += ["authentication.css"]
+Rails.application.config.assets.precompile += [
+  "authentication.css",
+  "cloud/bootstrap.js"
```
vitoravelino: Why not `//= require cloud/bootstrap` in application.js?

vitoravelino: Oh, because you are using it as an independent one in that specific page. Gotcha!

bear454: Didn't see the need to load it on other pages, and have to add code to not error out on other pages ;-)

kiall:
Looks like this needs to be rebased to pass CI, due to some other changes which have merged since this was last rebased/originally branched.

config/routes.rb
bear454: I changed it in d15b5f90.

app/views/setup/_cluster_size_panel.html.slim
bear454: Changed in d15b5f90.

app/views/setup/_cluster_size_panel.html.slim
bear454: Fixed in d15b5f90.

app/assets/javascripts/cloud/bootstrap.js
bear454: I've added a second comment with the license; legal has approved.

bear454:
I've addressed the issues that didn't already have a satisfactory discussion, and rebased on master. I _still_ don't understand the Jenkins failure, and need someone more familiar to highlight the issue, if any.

app/views/setup/_cluster_size_panel.html.slim
vitoravelino: 👍

app/views/setup/_cluster_size_panel.html.slim
vitoravelino: 👍

MaximilianMeister:
> I've addressed the issues that didn't already have a satisfactory discussion, and rebased on master. I still don't understand the Jenkins failure, and need someone more familiar to highlight the issue, if any.

@bear454 seems to happen only on jenkins, at least locally it went beyond the failure in the ci run. currently bootstrapping...

MaximilianMeister:
@bear454 could you rebase and push again? there were some ci fixes that need to be pulled in

PR #439, opened 2018-02-06
Use separate Dex clients for each actual client

kiall:
Previously Velum, CaaSP CLI, and Kubernetes all shared a single Dex
client. From a security perspective, this was far from ideal.

Update Velum to:

* Generate a unique secret for the Velum and Kubernetes client during setup
* Add a migration to generate secrets during upgrade
* Use the Velum client to auth with Dex
* Request a token from Dex which is valid for the kubernetes client

Depends-On: https://github.com/kubic-project/salt/pull/380
Depends-On: https://github.com/kubic-project/caasp-cli/pull/9


kiall:
This will fail CI without the dependant commits

app/controllers/setup_controller.rb
```diff
@@ -107,6 +107,9 @@ def settings_params
       settings["no_proxy"] = ""
     end
 
+    settings["dex_client_secrets_kubernetes"] = Pillar.value(pillar: :dex_client_secrets_kubernetes) || SecureRandom.uuid
+    settings["dex_client_secrets_velum"] = Pillar.value(pillar: :dex_client_secrets_velum) || SecureRandom.uuid
```
ereslibre: Can these cases really happen? We should have migrated the database, no?

kiall: The migrations don't run on a fresh install?

app/models/pillar.rb
```diff
@@ -14,34 +14,36 @@ def value(pillar:)
...
+        ldap_group_dn:                 "ldap:group_dn",
+        ldap_people_dn:                "ldap:people_dn",
+        ldap_base_dn:                  "ldap:base_dn",
+        ldap_admin_group_dn:           "ldap:admin_group_dn",
+        ldap_admin_group_name:         "ldap:admin_group_name",
+        ldap_tls_method:               "ldap:tls_method",
+        ldap_mail_attribute:           "ldap:mail_attribute",
+        cloud_framework:               "cloud:framework",
+        dex_client_secrets_kubernetes: "dex:client_secrets:kubernetes",
+        dex_client_secrets_velum:      "dex:client_secrets:velum",
```
ereslibre: Trailing comma, rubocop will complain probably.

kiall: Fixed

app/controllers/setup_controller.rb
ereslibre: Right.

PR #444, opened 2018-02-15
Reduce salt-api timeout from 60 sec, to 30 sec

kiall:
HAProxy has a 50 sec timeout, meaning HAProxy will kill the connection before Velum times out waiting for salt-api when there are issues.


kiall:
Fixed the rubocop fails.

kiall:
Updated commit message with the explanation - I had thought I already did that! Whoops.

Will merge - I believe this will help debugging the CI issues.

PR #447, opened 2018-02-26
feature: settings page

vitoravelino:
## Depends on
https://github.com/kubic-project/salt/pull/410

## Screenshots

screenshot-20180216120047-1233x516
screenshot-20180216154252-1243x526
screenshot-20180216155005-1252x498
screenshot-20180302102046-1220x458
screenshot-20180302102103-1216x505
screenshot-20180302102123-1205x454
screenshot-20180302102141-1205x449
screenshot-20180302102155-1210x486
screenshot-20180302102213-1208x402
screenshot-20180302102233-1206x431
screenshot-20180302102319-1203x391
screenshot-20180302102328-1205x622


MaximilianMeister:
[velum: allow users to configure on-premise registry mirrors](https://trello.com/c/VlUFFwDn/477-velum-allow-users-to-configure-on-premise-registry-mirrors)

app/controllers/settings/registries_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    @registry = Registry.new(registry_params.except(:certificate))
+    @cert = Certificate.find_or_initialize_by(certificate: certificate_param)
+
+    ActiveRecord::Base.transaction do
+      begin
+        @registry.save!
+        create_or_update_certificate! if certificate_param.present?
+        @created = true
+      rescue StandardError
+        raise ActiveRecord::Rollback
```
ereslibre: There's no need to do this, `.transaction` will rollback if it gets any exception: https://github.com/rails/rails/blob/05eccb116ded8fd1250524cc4314907ff77511cd/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L231-L236

app/controllers/settings/registry_mirrors_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    ActiveRecord::Base.transaction do
+      begin
+        @registry_mirror.save!
+
+        create_or_update_certificate! if certificate_param.present?
+
+        @created = true
+      rescue StandardError
+        raise ActiveRecord::Rollback
```
ereslibre: Same here.

app/controllers/application_controller.rb
```diff
@@ -29,4 +35,17 @@ def accessible_hosts
   def known_host?
     accessible_hosts.include?(request.host) || accessible_hosts.include?(request.host_with_port)
   end
+
+  def not_found
+    render_404
+  end
+
+  def render_404
```
ereslibre: Given `not_found` only calls `render_404` why not move the body of `render_404` inside `not_found` and remove `render_404`?

app/controllers/application_controller.rb
```diff
@@ -29,4 +35,17 @@ def accessible_hosts
...
+    respond_to do |format|
+      format.html do
+        render file: Rails.root.join("public", "404"), layout: false, status: :not_found
```
ereslibre: I'd prefer to not do this here. IMO we should just `render status: not_found, nothing: true` and then HAProxy should load the file to be shown in that case.

I'm not against merging this as it is though.

app/controllers/settings/registries_controller.rb
```diff
@@ -0,0 +1,103 @@
+# Settings::RegistriesController is responsibe to manage all the requests
+# related to the registries feature
+class Settings::RegistriesController < SettingsController
+  before_action :set_registry, only: [:show, :edit, :update, :destroy]
```
ereslibre: Maybe we can do it like the opposite, `except: :index`.

app/controllers/settings/registry_mirrors_controller.rb
```diff
@@ -0,0 +1,100 @@
+# Settings::RegistryMirrorsController is responsibe to manage all the requests
+# related to the registry mirrors feature
+class Settings::RegistryMirrorsController < SettingsController
+  before_action :set_registry_mirror, only: [:show, :edit, :update, :destroy]
```
ereslibre: Same here.

app/controllers/settings/registry_mirrors_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+
+        if certificate_param.present?
+          create_or_update_certificate!
+        elsif @registry_mirror.certificate.present?
+          @registry_mirror.certificate.destroy!
+        end
+
+        @updated = true
+      rescue StandardError
+        raise ActiveRecord::Rollback
```
ereslibre: Same with this block, we can remove the inner block and just let exceptions be raised without any `rescue` here.

app/controllers/settings/registries_controller.rb
vitoravelino: I read that it wasn't necessary but without that I get `ActiveRecord::RecordInvalid` exception on my face all the time instead of just rolling back and having page rendered with errors as expected. Tests also fail because nobody catches this exception.

Do you have any idea why?

app/controllers/application_controller.rb
vitoravelino: Is HAProxy using the 404 page in the public folder?

app/controllers/settings/registries_controller.rb
ereslibre: We talked about that on IRC, but writing here for the sake of completion. `.transaction` will rollback in case of an exception and will bubble up the exception again, so we also need to catch it on our controller.

app/controllers/application_controller.rb
ereslibre: Not at the moment, but we could make it do so.

vitoravelino:
Jenkins is green again! :smile:

MaximilianMeister:
> Jenkins is green again!

yeah, i just retriggered, then let's get it merged and work on the follow ups

PR #448, opened 2018-02-28
Be consistent and rename `upgrade` to `update`.

ereslibre:
I didn't refactor down the line because ultimately we cannot have an
enum with a :update status (rails will complain, so at least the enum
on the Orchestration model would need to be called :upgrade).

For that reason, keep everything as it is (call internally `upgrade`,
but just show "Retry cluster update" on the button label).

### DIFF
--- a/app/views/dashboard/index.html.slim
+++ b/app/views/dashboard/index.html.slim
@@ -55,7 +55,7 @@ h1 Cluster Status
 
       = link_to orchestrations_upgrade_path, method: :post, id: "retry-cluster-upgrade", class: "hidden btn btn-sm btn-primary pull-right" do
         i.fa.fa-refresh.fa-fw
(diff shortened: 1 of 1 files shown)


ereslibre:
[[QA] Test "Retry orchestration" feature](https://trello.com/c/zfAAGakY/670-qa-test-retry-orchestration-feature)

ereslibre:
> should we also update the Orchestration kind in the model? for internal consistency

I explain it on the commit message. On my first attempt I renamed everything (no `upgrade` at all in the source code). However this errors out because rails says that a new method `update!` is created because of the enum but that method already exists, and Velum doesn't start. I remember now why I picked up `upgrade` instead. So, let's keep `upgrade` everywhere internally except on the button caption...

PR #449, opened 2018-03-03
Azure integration

bear454:


bear454:
@MaximilianMeister said:
> could you still add a warning flash message when there are no values given in the form?
> and do you plan to add some input validations to all the fields at some point?

Yes, I plan to add validations both for presence, and content expectations, as we build up. We're moving forward with Azure ARM (a newer API), which we have only just begun using in the past month; we still have some learning here as well.

As an aside, moving forward, we expect to remove many of those form fields in a future version, when we implement support for VM-based security roles in Azure (as we do for EC2).

> also having some tooltips would be of great help for most of the fields. users might already know what they mean but i e.g. would have appreciated a small hint here and there ;-)

Yes, absolutely. We have a chicken-and-egg problem here, with needing a UI in order to test salt-cloud integration, and needing salt-cloud integration to provide proper feedback about some UI data (like bad input)... have to start somewhere, and this is the minimal viable interface.

THANK YOU for accepting, and expect continuous improvement!

PR #450, opened 2018-03-07
[wip] feature: node removal

vitoravelino:


app/views/dashboard/_warn_node_removal_modal.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+        <h4 class="modal-title" id="modal-label">
+          Invalid cluster topology
+        </h4>
+      </div>
+      <div class="modal-body">
+        <p>By removing the respective node you'll be breaking the following contrainsts that is needed to have a stable and supported topology:</p>
```
MaximilianMeister: `constraints that are needed`

app/views/dashboard/_warn_node_removal_modal.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+        <ul class="node-removal-contrainsts-list"></ul>
```
MaximilianMeister: constraints

app/models/orchestration.rb
```diff
@@ -57,6 +63,12 @@ def update_minions
       Minion.mark_pending_bootstrap
     when "upgrade"
       Minion.mark_pending_update
+    when "removal"
+      # rubocop:disable Rails/SkipsModelValidations
+      Minion.where(minion_id: params["target"]).update_all(
+        highstate: Minion.highstates[:pending_removal]
+      )
```
MaximilianMeister: should we have `Minion.mark_pending_removal` for consistency?

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -320,12 +353,23 @@ MinionPoller = {
         break;
     }
 
+    if (State.pendingRemovalMinionId || State.hasPendingStateNode) {
+      actionsHtml = '<a href="#" class="disabled remove-node-link" data-id="' + minion.minion_id + '">Remove</a>';
+    } else {
+      actionsHtml = '<a href="#" class="remove-node-link" data-id="' + minion.minion_id + '">Remove</a>';
```
MaximilianMeister: should we open a confirmation modal/dialog before actually performing the call? for such a destructive action IMO it's better to prevent accidental removes

app/controllers/minions_controller.rb
```diff
@@ -0,0 +1,15 @@
+# Allows to manage Minions
+class MinionsController < ApplicationController
+  before_action :fetch_minion
+
+  def destroy
+    Orchestration.run kind: :removal, params: { target: @minion.minion_id }
```
MaximilianMeister: even though it's prevented from the front-end side, should we make it safer on the api level by adding sth like:

```ruby
def destroy
  Orchestration.run kind: :removal, params: { target: @minion.minion_id }
rescue Orchestration::OrchestrationOngoing
  Rails.logger.debug "Orchestration already ongoing."
ensure
  render nothing: true, status: :no_content
end
```

app/assets/javascripts/dashboard/dashboard.js
vitoravelino: Yes, we should do the same as we did for the registries/mirrors.

app/controllers/minions_controller.rb
vitoravelino: The way that it is it's fine but If we change it, just remember to return a different status instead of no content every time. If an error happens, make it explicit to the client side.

vitoravelino: Hm... It would be a good idea to explicitly tell the user that another orchestration is going on and not show a generic error message.

MaximilianMeister:
[Node Remove: UI implementation](https://trello.com/c/2dLxFSpy/627-node-remove-ui-implementation)

ereslibre:
This is ready to be reviewed and complete.

app/views/dashboard/_warn_node_removal_modal.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="modal fade warn-node-removal-modal" tabindex="-1" role="dialog" aria-labelledby="modal-label">
```
MaximilianMeister: is there a reason why we dont use slim on the modal?

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180221142221) do
+ActiveRecord::Schema.define(version: 20180312112638) do
```
MaximilianMeister: you'll need to rebase the [patch](https://github.com/kubic-project/velum/blob/master/packaging/suse/patches/0_set_default_salt_events_alter_time_column_value.rpm.patch) otherwise it wont build (due to the strictness in our configuration)

MaximilianMeister:
could it be that some change around the accept minion has broken the `velum-bootstrap` script? looking at the jenkins failure, i think we might need to update [automation](https://github.com/kubic-project/automation/tree/master/velum-bootstrap) as well? can't figure out which change it was, but probably sth in the javascript changes regarding "accept" in this PR. @vitoravelino can you check?

app/views/dashboard/_warn_node_removal_modal.html.erb
vitoravelino: I copied and pasted from the other modal. We didn't have much time when I implemented that one and I wasn't aware of the html2slim converter. Lemme convert it.

vitoravelino:
> could it be that some change around the accept minion has broken the velum-bootstrap script? looking at the jenkins failure, i think we might need to update automation as well? can't figure out which change it was, but probably sth in the javascript changes regarding "accept" in this PR. @vitoravelino can you check? 

I'll take a look.

ereslibre:
@MaximilianMeister Based on the screenshot I think this is a random failure: http://jenkins.caasp.suse.net/job/velum/job/PR-450/20/artifact/screenshots/error_state.png

db/schema.rb
ereslibre: THE patch, right. I **always** forget.

MaximilianMeister:
> Based on the screenshot I think this is a random failure:

i have retriggered the ci and it's the same error again

MaximilianMeister:
bootstrap and feature worked perfectly fine on an older image `Build3.279`

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -119,41 +151,38 @@ MinionPoller = {
...
+            pendingRendered += MinionPoller.renderPendingNodes(pendingMinions[i]);
           }
 
           $(".pending-nodes-container tbody").html(pendingRendered);
         }
 
         // Show / Hide the table depending the presence of pending nodes
-        var acceptLinks = $('.pending-nodes-container td > a');
-        $(".pending-nodes-container #accept-all").prop('disabled', acceptLinks.length === 0)
+        checkAcceptAllAvaiability();
```
ereslibre: This looks like a typo.

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -744,3 +808,120 @@ $('body').on('click', '.role-btn-group .btn', function(e) {
...
+  var id = $this.data('id');
+
+  e.preventDefault();
+  removePendingAcceptance(id);
+
+  if ($this.hasClass('disabled')) {
+    return;
+  }
+
+  if (confirm('Are you sure?')) {
```
ereslibre: I think it wouldn't harm to also include the minion hostname and/or id on the confirmation message.

PR #451, opened 2018-03-08
Added CPI configuration

vitoravelino:
screenshot-20180312112905-1162x70
screenshot-20180312112856-1157x192
screenshot-20180312113157-1162x768

### DIFF
--- a/app/helpers/setup_helper.rb
+++ b/app/helpers/setup_helper.rb
@@ -1,0 +1,12 @@
+# SetupHelper contains all the view helpers for the setup process.
+module SetupHelper
+  def cloud_provider_options
+    [
+      ["OpenStack", :openstack]
+    ]
+  end
+
+  def cloud_providers_options_for_select
+    options_for_select(cloud_provider_options, selected: @cloud_provider)
+  end
+end
--- a/app/views/setup/cloud/_openstack_configuration.html.slim
+++ b/app/views/setup/cloud/_openstack_configuration.html.slim
@@ -1,0 +1,31 @@
+.openstack-settings class="#{'hidden' if @cloud_provider != 'openstack'}"
+  .form-group
+    = f.label :cloud_openstack_auth_url, "Keystone API URL"
+    = f.text_field :cloud_openstack_auth_url, value: @cloud_openstack_auth_url, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_domain, "Domain name"
+    = f.text_field :cloud_openstack_domain, value: @cloud_openstack_domain, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_project, "Project name"
+    = f.text_field :cloud_openstack_project, value: @cloud_openstack_project, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_region, "Region name"
+    = f.text_field :cloud_openstack_region, value: @cloud_openstack_region, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_username, "Username"
+    = f.text_field :cloud_openstack_username, value: @cloud_openstack_username, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_password, "Password"
+    = f.password_field :cloud_openstack_password, value: @cloud_openstack_password, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_subnet, "Subnet UUID for the CaaS Platform private network"
+    = f.text_field :cloud_openstack_subnet, value: @cloud_openstack_subnet, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_floating, "Floating network UUID"
+    = f.text_field :cloud_openstack_floating, value: @cloud_openstack_floating, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_lb_mon_retries, "Load balancer monitor max retries"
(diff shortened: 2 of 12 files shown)


mjura:
[Write user case for cloud provider UI part in Velum](https://trello.com/c/BoNizzYj/364-write-user-case-for-cloud-provider-ui-part-in-velum)

MaximilianMeister:
right now it tries to behave the same as the other cloud providers like ec2 and azure. meaning it expects a dedicated view for that in the worker-bootstrap page. similar to https://github.com/kubic-project/velum/blob/master/app/views/setup/worker_bootstrap_ec2.html.slim

but cant find `/srv/velum/config/data/openstack-types.json` when clicking next.

so i think you need to add an exception to https://github.com/kubic-project/velum/pull/451/files#diff-96d7538446a97258fe291a027a04c6baL57 regarding openstack, to not try to render a cloud worker view

vitoravelino:
> right now it tries to behave the same as the other cloud providers like ec2 and azure. meaning it expects a dedicated view for that in the worker-bootstrap page. similar to https://github.com/kubic-project/velum/blob/master/app/views/setup/worker_bootstrap_ec2.html.slim

You are right. We need to ask public cloud team what we should deliver in this pull request. But I'm aware that not all the options are functional, I was just playing around. I think it will only be EC2 and Azure for now but I also make it ready for openstack too on behind the scenes.

vitoravelino:
I'm wondering if these openstack fields shouldn't be in a dedicated page. I guess when the openstack integration is done, this will be handled. Anyway, only EC2 and Azure are enabled atm.

mjura:
> I'm wondering if these openstack fields shouldn't be in a dedicated page. I guess when the openstack integration is done, this will be handled. Anyway, only EC2 and Azure are enabled atm.

Yes, you are right OpenStack fields are dedicated only for openstack driver, Azure, AWS and GCE have different fields.

vitoravelino:
> Yes, you are right OpenStack fields are dedicated only for openstack driver, Azure, AWS and GCE have different fields.

What I wanted to know if it should be in a dedicated page as it was done for EC2 and Azure or if it's in the same page as shown in the screenshots. Let's wait for some feedback.

app/helpers/setup_helper.rb
```diff
@@ -0,0 +1,16 @@
+# SetupHelper contains all the view helpers for the setup process.
+module SetupHelper
+  def cloud_provider_options
+    [
+      ["Azure", :azure],
+      ["Amazon EC2", :ec2]
+      # Still missing templates / business logic around it
+      # ["Google Compute Engine", :gce],
+      # ["OpenStack", :openstack]
```
mjura: "OpenStack" has to be un-commented, you can comment "Azure" an "Amazon"

vitoravelino:
> Why OpenStack integration is disabled?

Because there's no integration ready for it afaik. The integrations submitted by the public cloud team so far were only AWS and Azure. Is OpenStack our responsibility? I thought that the requirements for this were just adding a simple select with the options based on what had been done. I might be wrong, so guide me here if that's the case.

What does OpenStack integration mean? Only the option available or something else (e.g.: instance types)? Because if we enable OpenStack there it will throw an exception because there's stuff missing that I'm not aware of. That's why I added the guy from public cloud team to also review this.

vitoravelino:
Confusions were cleared. I'm going to proceed with the expected changes.

vitoravelino:
@mjura This is good to be tested again.

bear454:
Public Cloud environments will be automatically detected; our test images already provide appropriate selection of the UI workflow, by injecting the cloud:framework pillar at boot in EC2, Azure. There's no need for a dropdown to select, for example, EC2 as a framework.

aosthof:
I agree with @bear454 . By adding this dropdown an additional (unnecessary) step would get added for the user. Let's remove the dropdown step of the UI unless there's an important reason to keep it.

mjura:
@aosthof @bear454: This change is about enabling Kubernetes Cloud Provider Integration driver, it doesn't have anything common with Public Cloud integration yet. If you would like to get more details please check it out https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/

vitoravelino:
@mjura I've reverted the rename that I did before. Please review this again.

app/views/setup/cloud/_openstack_configuration.html.slim
```diff
@@ -0,0 +1,31 @@
...
+    = f.text_field :cloud_openstack_project, value: @cloud_openstack_project, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_region, "Region name"
+    = f.text_field :cloud_openstack_region, value: @cloud_openstack_region, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_username, "Username"
+    = f.text_field :cloud_openstack_username, value: @cloud_openstack_username, class: "form-control"
+  .form-group
+    = f.label :cloud_openstack_password, "Password"
+    = f.text_field :cloud_openstack_password, value: @cloud_openstack_password, class: "form-control"
```
MaximilianMeister: this should not be a `text_field`

mjura: Yes, I agree

MaximilianMeister:
@vitoravelino @mjura i have added the cloud provider to the pillars api in a separate commit, otherwise salt wouldn't set the pillars

vitoravelino:
> @vitoravelino @mjura i have added the cloud provider to the pillars api in a separate commit, otherwise salt wouldn't set the pillars

Thanks!

PR #452, opened 2018-03-09
prevent multiple orchestrations at the same time

MaximilianMeister:

### DIFF
--- a/app/models/orchestration.rb
+++ b/app/models/orchestration.rb
@@ -4,6 +4,7 @@ require "velum/salt_orchestration"
 # Orchestration represents a salt orchestration event
 class Orchestration < ApplicationRecord
   class OrchestrationAlreadyRan < StandardError; end
+  class OrchestrationOngoing < StandardError; end
 
   enum kind: [:bootstrap, :upgrade]
   enum status: [:in_progress, :succeeded, :failed]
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
[How to apply the changes?](https://trello.com/c/O2BGEKMv/688-how-to-apply-the-changes)

app/models/orchestration.rb
```diff
@@ -26,9 +27,14 @@ def run
   # rubocop:enable Rails/SkipsModelValidations
 
   def self.run(kind: :bootstrap)
+    raise OrchestrationOngoing unless runnable?
     Orchestration.create!(kind: kind).tap(&:run)
   end
 
+  def self.runnable?
+    Orchestration.all.none?(&:in_progress?)
```
ereslibre: It's always better to use the scopes:

```
irb(main):005:0> Orchestration.all.none?(&:in_progress?)
  Orchestration Load (0.4ms)  SELECT `orchestrations`.* FROM `orchestrations`
=> true
```

as opposed to

```
irb(main):004:0> Orchestration.in_progress.empty?
   (3.5ms)  SELECT COUNT(*) FROM `orchestrations` WHERE `orchestrations`.`status` = 0
=> true
```

With `Orchestration.in_progress.empty?` we don't need to load the records in memory just to check if there's any, and the SQL generated sentence is smarter.

ereslibre (changes requested):
Except for the comment, LGTM.

app/models/orchestration.rb
MaximilianMeister: updated, thx

ereslibre (approved):
LGTM

vitoravelino (approved):
:+1:

PR #454, opened 2018-03-12
[WIP] Refuse to start Velum if a command fails during init

kiall:
If a command in the Velum container init script fails, we should refuse
to start. Right now, we'll silently ignore failures such as seeding
the database.

Additionally, Use production DB seeds in development - if a seed is useful in
production, we should have the same seed in development, while also allowing for
development specific seeds if necessary. This reduces the likelihood of a
bad change merging and only affecting non-devenv/CI environments.

--

WIP as 1) untested and 2) this will fail CI as the production seeds are broken - see https://github.com/kubic-project/velum/pull/453

### DIFF
--- a/bin/init
+++ b/bin/init
@@ -2,7 +2,7 @@
 
 # This script will setup the database and then start the rails application
 
-set -e
+set -euo pipefail
 
 : ${VELUM_PORT:=80}
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,20 +7,24 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-def seed_development
-  # nothing yet
-end
-
 def seed_production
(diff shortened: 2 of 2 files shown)


bin/init
```diff
@@ -2,7 +2,7 @@
 
 # This script will setup the database and then start the rails application
 
-set -e
+set -euo pipefail
```
kiall: Chances are, something in this script is expected to fail - and we'll need top update more code to make it work with the stricter environment.

kiall:
And... Despite `bundle exec rake db:seed` clearly failing in my dev env, it exits with a 0 code.

ereslibre:
This has conflicts now @kiall, I think we want to keep the main part of this patch, no?

dannysauer:
The ruby changes are all incorporated into master now, and there would be little benefit to adding `nounset` or `pipefail` to the current script.  There are four variables, all of which are not going to cause unnoticed bad results if they get typo'd, and there's only one pipeline which is inside of a `set +e` anyway.

Closing this PR as a result.

PR #457, opened 2018-03-16
Add rake task to import CPI options

mjura:
This task can be called:
```
docker exec -it $(docker ps | grep "velum-dashboard" | awk '{print $1}') entrypoint.sh bash -c "export RAILS_ENV=production; bundle exec rake velum:cpi_openstack['/etc/caasp/openstack.conf']
```

Example `openstack.conf` file which can be inject by cloud-init
```
[Global]
auth-url="http://prvcld.caasp.suse.net:5000/v3"
username="test"
password="secure"
tenant-name="caasp"
domain-name="caasp"
region="az"
[LoadBalancer]
lb-version=v2
subnet-id="e4f92380-4189-4dab-84c9-82830fe45522"
floating-network-id="e4f92380-4189-4dab-84c9-82830fe43311"
create-monitor=yes
monitor-delay=1m
monitor-timeout=30s
monitor-max-retries=3
[BlockStorage]
trust-device-path=false
bs-version=v3
```

### DIFF
--- a/lib/tasks/cpi.rake
+++ b/lib/tasks/cpi.rake
@@ -3,7 +3,7 @@
 namespace :cpi do
   desc "Import OpenStack Cloud Proivider options"
   task :openstack, [:config] => :environment do |_, args|
-    config_file = args[:config] || "/etc/caasp/openstack.conf"
+    config_file = args[:config] || "/etc/caasp/cpi/openstack.conf"
     unless File.exist?(config_file)
       puts "OpenStack Cloud Provider config file doesn't exist"
       exit(1)
--- a/lib/tasks/cpi.rake
+++ b/lib/tasks/cpi.rake
@@ -1,0 +1,38 @@
+# rubocop:disable Metrics/BlockLength
+
+namespace :cpi do
+  desc "Import OpenStack Cloud Proivider options"
+  task :openstack, [:config] => :environment do |_, args|
+    config_file = args[:config] || "/etc/caasp/openstack.conf"
+    unless File.exist?(config_file)
+      puts "OpenStack Cloud Provider config file doesn't exist"
+      exit(1)
+    end
+    cfg = {}
+    File.open(config_file, "r").each do |line|
+      cfg["cloud:provider"] = "openstack"
+      line.chomp!
+      key, value = line.delete('"').split("=")
+      case key
+      when /^[\[#]/ then              puts "Skipping the line"
+      when "auth-url" then            cfg["cloud:openstack:auth_url"] = value
+      when "domain-name" then         cfg["cloud:openstack:domain_name"] = value
+      when "tenant-name" then         cfg["cloud:openstack:tenant_name"] = value
+      when "region" then              cfg["cloud:openstack:region"] = value
+      when "username" then            cfg["cloud:openstack:username"] = value
+      when "password" then            cfg["cloud:openstack:password"] = value
+      when "subnet-id" then           cfg["cloud:openstack:subnet_id"] = value
+      when "floating-network-id" then cfg["cloud:openstack:floating_id"] = value
+      when "monitor-max-retries" then cfg["cloud:openstack:lb_mon_retries"] = value
+      when "bs-version" then          cfg["cloud:openstack:bs_version"] = value
+      when /^./ then                  puts "Unknown option: #{key}"
+      end
+    end
+    cfg.each do |pillar, value|
+      Pillar.find_or_create_by!(pillar: pillar) do |p|
+        p.value = value
+      end
+    end
+  end
+end
+# rubocop:enable Metrics/BlockLength
--- a/bin/init
+++ b/bin/init
@@ -61,6 +61,16 @@ setup_database() {
   set -e
 }
 
+setup_cpi() {
+  # Check if Cloud Provider config exists
+  CPI_CONFIG="/etc/caasp/cpi/openstack.conf"
+  if [ -f "${CPI_CONFIG}" ]; then
+    # Import Cloud Provider config
+    bundle exec rake cpi:openstack['${CPI_CONFIG}']
+  fi
+}
+
 setup_root_ca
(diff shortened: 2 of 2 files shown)


mjura:
[CaaS Platform on OpenStack instead of Magnum](https://trello.com/c/T1y2NObB/547-caas-platform-on-openstack-instead-of-magnum)

ereslibre:
I think it would be worth having a generic way to import external pre-settings. Something like:

`rake db:preload_settings` or something similar.

This task could retrieve all `*.json` files in a given directory (mounted from the host). I'd prefer to use json than a set of regexps/splits. Also, this could be completely agnostic in terms of what pillars it imports. If a json inside that directory looks something like:

```json
{
  "some:pillar": "some value",
  "other_pillar": "other value"
}
```

The logic could just:

```ruby
cfg.each do |pillar, value|
  Pillar.find_or_create_by!(pillar: pillar) do |pillar|
    pillar.value = value
  end
end
```

mjura:
@ereslibre: I would like to keep it in ini format. With it user can provide `openstack.conf` file based on Kubernetes documentation https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/ 

We can import this option and validate them in this rake file, after then they will be imported to Velum and used for bootstraping

MaximilianMeister:
> I see. This is an INI file format. Always my concerns with this kind of processing of INI files is what happens when on two different sections you have the same key, with the current way of processing we'd just pick the latest one and the way we process the file is unaware of sections.

although this is a valid concern i think this is not a big issue in our case, if there are the same keys in different sections it's misconfiguration anyways, and section awareness could be easily implemented in plain ruby with a simple helper method

> 
> I'd prefer to add inifile gem and process the INI file with that.

from a maintenance point of view this should possibly be avoided. IMO it would only make sense to introduce a new gem if it's a critical component of velum. and if we'd go that way we should probably use a gem that can parse more than one config file format, if that exists

lib/tasks/cpi.rake
```diff
@@ -0,0 +1,38 @@
+# rubocop:disable Metrics/BlockLength
+
+namespace :cpi do
+  desc "Import OpenStack Cloud Proivider options"
+  task :openstack, [:config] => :environment do |_, args|
+    config_file = "/etc/caasp/openstack.conf"
+    config_file = args[:config] if args[:config]
```
MaximilianMeister: you could combine the 2 lines

```ruby
config_file = args[:config] || "/etc/caasp/openstack.conf"
```

lib/tasks/cpi.rake
```diff
@@ -0,0 +1,38 @@
...
+    exit(1) unless File.exist?(config_file)
```
MaximilianMeister: maybe give a message on what failed before exit

lib/tasks/cpi.rake
```diff
@@ -0,0 +1,38 @@
...
+    file = File.open(config_file, "r")
```
MaximilianMeister: this could be done within one block, which also closes the file at the end without an explicit `.close` call:

```ruby
...
    cfg = {}
    File.open(config_file, "r").each do |line|
      cfg["cloud:provider"] = "openstack"
      line.chomp!
      key, value = line.delete('"').split("=")
      case key
        when /^[\[#]/; # ignore line
        when "auth-url" then            cfg["cloud:openstack:auth_url"] = value
        when "domain-name" then         cfg["cloud:openstack:domain_name"] = value
        when "tenant-name" then         cfg["cloud:openstack:tenant_name"] = value
        when "region" then              cfg["cloud:openstack:region"] = value
        when "username" then            cfg["cloud:openstack:username"] = value
        when "password" then            cfg["cloud:openstack:password"] = value
        when "subnet-id" then           cfg["cloud:openstack:subnet_id"] = value
        when "floating-network-id" then cfg["cloud:openstack:floating_id"] = value
        when "monitor-max-retries" then cfg["cloud:openstack:lb_mon_retries"] = value
        when "bs-version" then          cfg["cloud:openstack:bs_version"] = value
        when /^./ then                  puts "#{key}: unknown option"
      end
    end
    cfg.each do |pillar, value|
...
```

ereslibre:
> although this is a valid concern i think this is not a big issue in our case, if there are the same keys in different sections it's misconfiguration anyways, and section awareness could be easily implemented in plain ruby with a simple helper method

We cannot control how this file will be modified over time, and since it's an INI file it would be perfectly valid for upstream to add a duplicated key in two different sections.

Aside from that, while I can understand we can go without proper ini parsing (I really dislike the idea but if you want to go that way I won't oppose), I'd rather see a generic task that loads all `*.conf` in a given directory, so it's faster to raise the environment once and load N files at once with that same env rather than starting N environments (rails initialization is really slow, even in production).

I presume we'll call this task from an init container or the `bin/init` script. If we do that, I'd like to see the environment only load once in order to import all the different configurations in different files rather than slowing down the Velum startup because they are splitted in different rake tasks.

lib/tasks/cpi.rake
```diff
@@ -0,0 +1,38 @@
+# rubocop:disable Metrics/BlockLength
+
+namespace :cpi do
+  desc "Import OpenStack Cloud Proivider options"
+  task :openstack, [:config] => :environment do |_, args|
+    config_file = "/etc/caasp/openstack.conf"
```
inercia: This hardcoded path here looks too _hard_ for me...

mjura:
It depends on https://github.com/kubic-project/caasp-container-manifests/pull/167

jordimassaguerpla:
@mjura : Please add this commit to fix the rubocop issues: https://github.com/mjura/velum/commit/5317972e13da9bae18333bfee66d1cc60c8a0d18

jordimassaguerpla:
Jenkins tests fail. @mjura : Please rebase with master branch and let the tests run again

mjura:
Rebased

PR #458, opened 2018-03-16
Add cloud:provider to cloud_provider_contents for Salt API call

mjura:
Without this change `cloud:provider` won't be saved in Salt as pillar.

### DIFF
--- a/app/controllers/internal_api/v1/pillars_controller.rb
+++ b/app/controllers/internal_api/v1/pillars_controller.rb
@@ -112,6 +112,7 @@ class InternalApi::V1::PillarsController < InternalApiController
   def openstack_cloud_contents
     {
       cloud: {
+        provider:  "openstack",
         openstack: {
           auth_url:       Pillar.value(pillar: :cloud_openstack_auth_url),
           username:       Pillar.value(pillar: :cloud_openstack_username),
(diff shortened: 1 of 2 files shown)


app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -112,6 +112,7 @@ def azure_cloud_contents
   def openstack_cloud_contents
     {
       cloud: {
+        provider: "openstack",
```
vitoravelino: Instead of hardcoded value I would use `Pillar.value(pillar: :cloud_provider)`.

ereslibre: Given that the method is called `openstack_cloud_contents` I don't think it's much of an issue, no?

vitoravelino: Gotcha. Just saw that `cloud_provider_contents` calls `openstack_cloud_contents` when it's `cloud_provider` is `openstack`. Then it's fine to be like this. 😅

vitoravelino:
Just fix the test and this should be good to merge.

app/controllers/internal_api/v1/pillars_controller.rb
mjura: Yes, this should stay hardcoded.

mjura:
I've updated tests

vitoravelino:
Rubocop is not happy with the changes...

mjura:
Hopefully this will fix rubocop complains

mjura:
[Write user case for cloud provider UI part in Velum](https://trello.com/c/BoNizzYj/364-write-user-case-for-cloud-provider-ui-part-in-velum)

ellisab:
tested it and works fine

PR #464, opened 2018-03-21
[test] reduce workers/threads

MaximilianMeister:
in smaller scale environments, choosing such a high number of default
workers and threads might lead to resource overconsumption, as regarding
puma (rule of thumb) nr of cpu core == nr of workers

under normal circumstances it's enough to handle the workloads of
the rails app with 2 workers and the default set of threads (5)

this needs to be tested with a big amount of nodes to be sure it can
handle the workload

ideally we would have a way to dynamically set this configuration before
starting the rails app

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,6 +1,6 @@
-WORKERS = ENV.fetch("VELUM_WORKERS", 8).freeze
-MIN_THREADS = ENV.fetch("VELUM_MIN_THREADS", 8).freeze
-MAX_THREADS = ENV.fetch("VELUM_MAX_THREADS", 32).freeze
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
rebased, to retrigger CI

MaximilianMeister:
after retriggering 2 times it always fails during the node tests, i'll let it run once more and see. maybe we need to adjust the values further

ereslibre:
@MaximilianMeister what do you think about this patch, should we keep it, or drop it?

MaximilianMeister:
> @MaximilianMeister what do you think about this patch, should we keep it, or drop it?

drop (for now)

PR #466, opened 2018-03-21
ui: allowed master node to be added to the cluster

vitoravelino:
So far we only allowed workers to be added to the cluster through the
unassigned nodes page.

With this patch, we are now allowing the users to select the roles of the
nodes they want to add to the cluster. Very similar to what is available
from the discovery bootstrap page.

This feature keep forcing the user to respect the contrainst of having an
odd number of master nodes so we have a stable topology.

feature#add_master_node

### Screenshots

screenshot-20180321150526-1203x624
screenshot-20180321150512-1216x603
screenshot-20180321150504-1206x710
screenshot-20180321150547-1201x718
screenshot-20180321150903-1203x712

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -1,6 +1,7 @@
 State = {
   nextClicked: false,
   bootstrapErrors: [],
+  assignableErrors: [],
 },
 
 MinionPoller = {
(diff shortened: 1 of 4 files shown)


vitoravelino:
[Allow master nodes to be added through the UI](https://trello.com/c/WKueiTqu/736-allow-master-nodes-to-be-added-through-the-ui)

Issue #456, opened 2018-03-12
discovery/dashboard page times out when a minion is down


jgleissner:
The dashboard page appears to query all minions for grains, which, in case a minion is unresponsive, can lead to a salt API call time-out. A Net::ReadTimeout exception is thrown as the log shows, and I assume not caught, and results in the generic "We're sorry, something went wrong" error message.

It seems the timeout for waiting for a response from the minion is 30 seconds:
```
bash-4.3# time salt '*' grains.get tx_update_failed
ca:
admin:
testnode1:
    Minion did not return. [No response]

real    0m30.671s
user    0m0.804s
sys     0m0.118s
```
This happens to be the same as the API call timeout, which was recently reduced to 30 secs, and hence is probably now triggered, unlike before. Slightly increasing the API timeout again, or reducing the timeout for the minion query could work around this. About the latter, if I reduce the timeout to 1 second, it still takes about 11 seconds to time out:
```
bash-4.3# time salt -t 1 '*' grains.get tx_update_failed
admin:
ca:
testnode1:
   Minion did not return. [No response]

real    0m11.863s
user    0m0.567s
sys     0m0.131s
```
I'm not sure where that 10 second base delay comes from. There's none when all minions are responsive.

ereslibre:
![](https://github.trello.services/images/mini-trello-icon.png) [discovery/dashboard page times out when a minion is down](https://trello.com/c/L5R2MD8j/717-discovery-dashboard-page-times-out-when-a-minion-is-down)

PR #468, opened 2018-03-23
Extend salt-api timeout as long as possible.

bear454:
Allow enough time for the salt timeout, and a minion timeout before cutting off the API call.

Should resolve https://github.com/kubic-project/velum/issues/456 introduced in 6189bcad .


bear454:
[[Issue] discovery/dashboard page times out when a minion is down](https://trello.com/c/L5R2MD8j/717-issue-discovery-dashboard-page-times-out-when-a-minion-is-down)

PR #470, opened 2018-03-26
Hide CPI panel when no cloud framework

vitoravelino:
When no cloud:framework pillar is set, it means that CaaSP is not
running in a cloud environment. So it doesn't make sense to show that
option in that scenario.

feature#cpi


vitoravelino:
[Hide OpenStack CPI options unless running on OpenStack](https://trello.com/c/bF1BBz6P/751-hide-openstack-cpi-options-unless-running-on-openstack)

PR #479, opened 2018-03-27
Fixed # of masters warning when removing worker

vitoravelino:
This patch fixes the warning modal that was being shown when the cluster
was in an unsupported configuration (even # masters) and user was trying
to remove a normal worker node. This scenario should trigger that
warning modal.

bsc#1087123


vitoravelino:
[Bug 1087123 - Removing worker node warns about breaking odd masters constraint](https://trello.com/c/szMIuoYA/759-bug-1087123-removing-worker-node-warns-about-breaking-odd-masters-constraint)

PR #481, opened 2018-03-28
Fixed # of masters warning when adding a new master

vitoravelino:
This patch fixes the warning that was being shown when the user selected
a new master node to be added and it wasn't breaking the odd number
constraint. Before this, the current assigned masters were not being
considered to show/not show the warning.

Now the warning will only be shown if the number of masters to be added
breaks the constraint as it's expected to be.

bsc#1087267

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -116,6 +116,7 @@ MinionPoller = {
           minions = allMinions;
           break;
         case "Unassigned":
+          State.assignedMinions = minions;
           minions = unassignedMinions;
           break;
         default:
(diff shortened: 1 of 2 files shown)


vitoravelino:
[Bug 1087267 - Scaling to 3 > 2 > 3 masters is not possible](https://trello.com/c/1FaClOHQ/761-bug-1087267-scaling-to-3-2-3-masters-is-not-possible)

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -561,10 +562,12 @@ function healthCheckToReload() {
 // begin unassigned
 function isAssignable() {
   var errors = [];
+  var masters = State.assignedMinions.filter(function (m) { return m.role === 'master' });
 
   // We need an odd number of masters
-  if (selectedMastersLength() % 2 !== 0) {
-    errors.push('The number of masters to be added has to be an even number in order to maintain the odd constraint number in the cluster');
+  if ((masters.length + selectedMastersLength()) % 2 === 0 &&
+      selectedMastersLength() > 0) {
```
MaximilianMeister: for optimization i would switch the logic here to first check for `selectedMastersLength() > 0` - this saves the heavyer calculation part in case `selectedMastersLength()` is 0

vitoravelino: Not a big deal during these days but you are right. I'll change it. Thanks!

vitoravelino: Done!

PR #485, opened 2018-04-02
Properly create and update orchestrations that have been executed out of Velum

ereslibre:
It's more correct for the removal orchestration to read the parameters that were
passed to the orchestration rather than to rely on the current status of the
minions on the database.

### DIFF
--- a/app/models/salt_handler/orchestration_result.rb
+++ b/app/models/salt_handler/orchestration_result.rb
@@ -19,11 +19,15 @@ class SaltHandler::OrchestrationResult < SaltHandler::Orchestration
     when "orch.kubernetes", "orch.update"
       update_minions orchestration_succeeded: orchestration_succeeded
     when "orch.removal"
+      # Since this orchestration is parametrized is more correct to only update the minions that
+      # are specified in the orchestration parameters.
       if orchestration_succeeded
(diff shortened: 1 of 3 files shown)


app/models/salt_handler/orchestration_result.rb
```diff
@@ -19,11 +19,15 @@ def process_event
     when "orch.kubernetes", "orch.update"
       update_minions orchestration_succeeded: orchestration_succeeded
     when "orch.removal"
+      # Since this orchestration is parametrized is more correct to only update the minions that
+      # are specified in the orchestration parameters.
       if orchestration_succeeded
-        Minion.pending_removal.destroy_all
+        Minion.pending_removal.where(minion_id: (orchestration.params || {})["target"]).destroy_all
```
MaximilianMeister: as we currently don't have a way in velum to remove multiple minions at the same time, a more direct and explicit `find_by(minion_id: (orchestration.params || {})["target"]).destroy` would make it more readable

rest LGTM :+1:

ereslibre: It was in purpose indeed, as this will work with a single target (as of now), and in the future if we change `target` to `targets` (array containing minions to remove) it will also work without problem with a minimum change.

PR #488, opened 2018-04-05
Public Cloud fixes

bear454:
* An pillar for Azure storage account name was described as optional in the documentation, but in practice is required.
* Improve the layout of inputs in public cloud forms.
* Add an explanatory link on Azure service principal attributes.

### DIFF
--- a/app/assets/stylesheets/pages/instance_type.scss
+++ b/app/assets/stylesheets/pages/instance_type.scss
@@ -30,6 +30,8 @@
             background: $table-bg;
             color: $brand-success;
 
+            cursor: pointer;
+
             margin: 10px;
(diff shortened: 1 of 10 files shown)


bear454:
[Public Cloud fixes](https://trello.com/c/di0Qokry/778-public-cloud-fixes)

vitoravelino:
Since you are fixing smaller things, what about adding the `cursor: pointer` to `.instance-type-box` that I've mentioned in the past? ;)

bear454:
@vitoravelino if Jenkins fails for some reason, because I added 16 chars of CSS... I'm blaming you. :stuck_out_tongue:

PR #490, opened 2018-04-12
[2.0] Avoid the event processor from crashing if it cannot interpret the event arguments

ereslibre:
If we executed: `salt-run state.orchestrate 2` we get an error that crashes the event
processor:

`TypeError: no implicit conversion of String into Integer`

This happens because of `fun_args.first["mods"]` being `2["mods"]`, what makes it
crash with the previous exception. `2.respond_to?(:[])` is `true`, so I think the
best thing we can do here is to swallow the exception, and ignore the event if we
cannot even process its arguments.

Fixes: bsc#1088597

Backport from https://github.com/kubic-project/velum/pull/489

### DIFF
--- a/app/models/salt_handler/minion_orchestration.rb
+++ b/app/models/salt_handler/minion_orchestration.rb
@@ -18,7 +18,12 @@ class SaltHandler::MinionOrchestration
     fun_args = parsed_event_data["fun_args"]
 
     orchestrations.each do |o|
-      return true if fun_args.first == o || fun_args.first["mods"] == o
+      # rubocop:disable Lint/HandleExceptions
+      begin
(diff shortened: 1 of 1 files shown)


ereslibre:
[Bug 1088597 - L3: Unable to add new node](https://trello.com/c/CnySLTo9/788-bug-1088597-l3-unable-to-add-new-node)

ereslibre:
> MINOR: but is this really a backport? maybe the label is wrong here

Yeah, it's a backport, not a clean one though :)

jordimassaguerpla:
Please backport this to 2.1. Thanks.

jordimassaguerpla:
https://github.com/kubic-project/velum/pull/491

PR #499, opened 2018-04-19
node removal: improved way of showing removal links

vitoravelino:
'Remove' and 'Force remove' links were being shown at the same time so
far. But 'Force remove' only makes sense to be shown when a previous
removal has failed.

With this we are showing only one link at a time. When the highstate of
a minion is fine, we show 'Remove'. Whenever it comes 'removal_failed',
we show 'Force remove'.

enhancemente#node-removal

screenshot-20180419144636-1161x377
screenshot-20180419144651-1152x380
screenshot-20180419144846-1156x331

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -303,6 +303,7 @@ MinionPoller = {
     var checked;
     var masterHtml;
     var actionsHtml;
+    var removalFailedClass = '';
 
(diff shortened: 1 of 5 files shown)


vitoravelino:
[Improve force / force remove links](https://trello.com/c/oeEa4zSp/814-improve-force-force-remove-links)

vitoravelino:
Fixed rubocop offense.

PR #503, opened 2018-05-02
Fixed misleading text about nodes selection

vitoravelino:
Small fix of a misleading text that referred to an old behavior of the
nodes discovery page.

bsc#1059787

### DIFF
--- a/app/views/setup/discovery.html.slim
+++ b/app/views/setup/discovery.html.slim
@@ -36,7 +36,9 @@ h1 Select nodes and roles
       i.fa.fa-times.fa-fw
       | Deselect all nodes
   .panel-body
-    p After choosing the master and clicking "Next" all the other selected nodes will be set to the worker role.
+    p
+      | Please choose an odd number of <em>Master</em> nodes, then click "Select remaining nodes"
+      |  to assign the remaining nodes to the <em>Worker</em> role.
 
     = form_tag(setup_discovery_path, method: "post")
       .nodes-container data-url=setup_discovery_path


vitoravelino:
[Bug 1059787 - Misleading help text on deployment](https://trello.com/c/7Y98uSOJ/834-bug-1059787-misleading-help-text-on-deployment)

vitoravelino:
As soon as we approve this, I'll create a backport for 2.1.

app/views/setup/discovery.html.slim
```diff
@@ -36,7 +36,7 @@ h1 Select nodes and roles
       i.fa.fa-times.fa-fw
       | Deselect all nodes
   .panel-body
-    p After choosing the master and clicking "Next" all the other selected nodes will be set to the worker role.
+    p After choosing the nodes and clicking "Next", all the selected nodes will be set to bootstrap.
```
rjschwei: I think this  still does not reflect what we want the user to do. Maybe something like this.

"Select 1 node as "Master" node and one or more of the remaining nodes as "Worker" nodes to bootstrap the cluster"

vitoravelino: Maybe `Select at least one node as master and one or more of the remaining nodes as worker to  set up the cluster to bootstrap`? 

My English is not good, so... Let's wait for other opinions.

lproven: Suggestion:

«
After selecting at least one node to be master, click "Select remaining nodes" and then "Next" to make all the others into workers.
»

vitoravelino: > After selecting at least one node to be master, click "Select remaining nodes" and then "Next" to make all the others into workers.

Doesn't it sound confusing having `to make all the others into workers` after `"Next"`?

Trying a quick fix...

`After selecting at least one node to be master, click "Select remaining nodes" to make all the others into workers and then "Next" to set up nodes to bootstrap.`

bear454: Consider:

```
    p
      | Please choose an odd number of
      i<> Master
      | nodes, then click
      i<> "Select remaining nodes"
      | to assign the remaining nodes to the
      i<> Worker
      | role.
```

flavio: I like @bear454 proposal. Let's go with it

lproven: I like that. I was trying to minimize alteration, but this is better.

vitoravelino:
Good to go!

flavio:
Feel free to merge **after** beta 1 is released.

vitoravelino:
@bear454 Correct. Already done with #511.

bear454:
Whoops - didn't even think to check if it was already merged. THANK YOU!

PR #504, opened 2018-05-03
Initial implementation of a system wide certificates.

bergmannf:
The implementation follows the registry & registry mirror style and
adds a simple (name, certificate) model.


bergmannf:
[Providing self-signed CA for CaaSP](https://trello.com/c/nNv85jtT/808-providing-self-signed-ca-for-caasp)

app/assets/javascripts/settings/index.js
```diff
@@ -9,4 +10,8 @@ $(function() {
   if ($registryForm.length) {
     new RegistryForm($registryForm);
   }
-});
\ No newline at end of file
+
+    if ($systemCertificateForm.length) {
+        new SystemCertificateForm($systemCertificateForm);
+    }
+});
```
vitoravelino: Could you use 2 instead of 4 spaces? :)

db/schema.rb
```diff
@@ -29,6 +29,12 @@
     t.datetime "updated_at"
   end
 
+  create_table "global_certificates", force: :cascade do |t|
+    t.string   "name",       limit: 255
+    t.datetime "created_at",             null: false
+    t.datetime "updated_at",             null: false
+  end
```
vitoravelino: Leftover?

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180406080400) do
+ActiveRecord::Schema.define(version: 20180427124945) do
```
MaximilianMeister: can you please rebase this [patch](https://github.com/kubic-project/velum/blob/master/packaging/suse/patches/0_set_default_salt_events_alter_time_column_value.rpm.patch) ? otherwise it will fail due to our automated and strict packaging.

this is one thing we always need to do when we update the `schema.rb`

MaximilianMeister:
> maybe i dont ahve the full picture, don't we want to use them for registries/mirrors e.g.? then we'd need to reference them, and change the UI workflow a bit to be able to e.g. first add a cert and then choose it when creating a registry or a mirror

i've just seen https://github.com/kubic-project/salt/pull/512 - adding the reference here

i am only still wondering if we should expose the registry/mirror certs in the same way in a generic certificates menu where we can set a type (system or registry/mirror) and add/change/remove certs in one single place, and choose the cert from a dropdown

right now we already have a `Certificate` model. can't we extend this with a type (system/registry) instead of creating a new independent `SystemCertificate` model?

bergmannf:
I think we will and up still having two different types of certificates:

1. `registries` and `mirrors` only install the certificates to `/etc/docker/certs.d/{{ name }}/ca.crt` and then are only used by `docker` and not by any other tool.

2. The `system-wide` certificates which go into `/etc/pki/anchors/` (or any other path valid for those certificates) and should be used by *all* tools in the system.

That's why I added them separate from the others.

I like the idea of having a single view to add certificates, but right now it seemed to me, that for `registries` and `mirrors` the certificate is optional (as it can be not self-signed) and for the system-wide one it is mandatory. I am currently looking at extracting some of the duplicated code into another controller, so I can also give a unified view a shot and see how that ends up looking.

app/controllers/setup_controller.rb
```diff
@@ -51,6 +52,11 @@ def welcome
 
     # container runtime setting
     @cri = Pillar.value(pillar: :container_runtime) || "docker"
+
+    # allow adding system certificate: required if a user uses CPI with a
+    # self-signed certificate
+    @system_certificate = SystemCertificate.new
+    @cert = Certificate.new
```
vitoravelino: Be aware that the user can go back and forth through the pages before the bootstrap itself and any previous value set should be there when the user goes to that specific page.

app/views/settings/system_certificates/_form.html.slim
```diff
@@ -0,0 +1,6 @@
+= form_for [:settings, @system_certificate], html: { class: "system-certificate-form"} do |f|
+  = render :partial '_fields', locals { f: f, required: true }
```
MaximilianMeister:
```ruby
render partial: 'fields', locals: { f: f, required: true }
```

app/views/settings/system_certificates/_form.html.slim
```diff
@@ -0,0 +1,6 @@
...
+
+.form-actions.clearfix
+  = f.submit "Save", class: "btn btn-primary action"
+  = link_to "Cancel", settings_system_certificates_path, class: "btn btn-default action"
```
MaximilianMeister: this 3 lines need to be indented with 2 spaces

MaximilianMeister: hmm not entirely sure after all, but as it is now, it doesnt render

app/controllers/setup_controller.rb
bergmannf: Good point, I updated it to use the `session` to store a entered `form`.

I only tested it with the browser's back, which worked, not with our `back` buttons.

app/controllers/setup_controller.rb
```diff
@@ -1,3 +1,4 @@
+require "pp"
```
MaximilianMeister: debugging remainder?

app/controllers/setup_controller.rb
```diff
@@ -64,10 +78,12 @@ def configure
                        unprotected_pillars: unprotected_pillars)
     registry_errors = Registry.configure_suse_registry(suse_registry_mirror_params)
 
-    if res.empty? && registry_errors.empty?
+    certificate_errors = create_system_certificate
+
+    if res.empty? && registry_errors.empty? && certificate_errors.empty?
```
MaximilianMeister: one way of shortening this would be

```ruby
if [res, registry_errors, certificate_errors].all?(:empty?)
```

app/controllers/settings/system_certificates_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def create
+    @system_certificate = SystemCertificate.new(system_certificate_params.except(:certificate))
+    @cert = Certificate.find_or_initialize_by(certificate: certificate_param)
+
+    ActiveRecord::Base.transaction do
+      @system_certificate.save!
+      create_or_update_certificate! if certificate_param.present?
+      @created = true
```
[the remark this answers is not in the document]
bergmannf: Oh yes, that was some remnants of debugging code.

app/views/setup/welcome.html.slim
```diff
@@ -156,5 +156,20 @@ h1 Initial CaaS Platform Configuration
   = render partial: 'setup/cloud/settings', locals: { f: f }
   = render partial: 'setup/settings/container_runtime', locals: { f: f }
 
+  .panel.panel-default
```
vitoravelino: To avoid growing this file more and more, extract this to a partial like we did for some of the other configs.

app/views/setup/welcome.html.slim
```diff
@@ -156,5 +156,20 @@ h1 Initial CaaS Platform Configuration
...
+    .panel-heading.clearfix
+      h3.panel-title
+        | Certificate
```
vitoravelino: I was wondering.... Maybe it would be a good idea to rename this to "System wide certificate". Or at least be more explicit on the description of this config?

vitoravelino:
Besides the concerns I've mentioned on the comments, LGTM! Really nice work, @bergmannf! :clap:

bergmannf:
@MaximilianMeister @vitoravelino Thanks for all the reviews.

I just pushed the last set of changes that should fix the latest comments. Unfortunately it changed a bit more logic, as I extracted a lot more logic into the `BaseCertificateController` that was shared between the `registry`, `registry_mirror` and `system_certificates`.

vitoravelino:
@bergmannf This needs rebase.

vitoravelino:
This one LGTM, is there anything blocking us to merge it besides the conflicts?

PR #505, opened 2018-05-03
Use `update_columns` to update minion attributes with the minion reconciler

ereslibre:
This will avoid strange issues when Rails caches the attributes of the `minions`
table (we also avoid the need to `reload` explicitly for this corner cases).

Fixes: bsc#1091843

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -15,17 +15,18 @@ class Minion < ApplicationRecord
 
   # Update all minions grains
   def self.update_grains
-    # rubocop:disable Lint/HandleExceptions
+    # rubocop:disable Lint/HandleExceptions, SkipsModelValidations
     Minion.all.find_each do |minion|
       begin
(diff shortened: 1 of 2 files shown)


ereslibre:
[Bug 1091843 - When updating from 2.0 the `update all nodes` link disappears](https://trello.com/c/DnZfae7N/842-bug-1091843-when-updating-from-20-the-update-all-nodes-link-disappears)

ereslibre:
We need superpowers to merge this one. @flavio, @kiall?

kiall:
> We need superpowers to merge this one. @flavio, @kiall?

@ereslibre Humm - why? It doesn't seem to depend on other changes etc, why does it need to be force merged?

ereslibre:
@kiall I messed up with this PR, this doesn't need a force merge at all, in fact is green, sorry for the confusion.

PR #515, opened 2018-05-09
Apply patch

bear454:
Why are we carrying patch files in git?


bear454:
[Apply patches](https://trello.com/c/sCVocosm/862-apply-patches)

flavio:
I think @jordimassaguerpla and @MaximilianMeister know more about that.

ereslibre:
I cannot really tell about the `Gemfile.lock` one, but losing the schema patch kind of breaks our workflow and can introduce error-prone behavior. If you modify the schema and then `rake db:schema:dump`, the schema rails generates will miss the patch, and you need to modify it manually in order to keep it as it was.

I would really keep the schema patch, because it allows us to use the rails tooling normally, namely: `db:schema:dump`.

bear454:
MaximilianMeister reviewed 2 hours ago:
> you still need to remove the patch from packaging/suse/patches, otherwise the package won't build

@MaximilianMeister I've updated the commits to include removing the patches

bear454:
ereslibre commented 4 hours ago:
> I cannot really tell about the Gemfile.lock one, but losing the schema patch kind of breaks our workflow and can introduce error-prone behavior. If you modify the schema and then rake db:schema:dump, the schema rails generates will miss the patch, and you need to modify it manually in order to keep it as it was.
> 
> I would really keep the schema patch, because it allows us to use the rails tooling normally, namely: db:schema:dump.
> 
> Failing to set "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP" on the database will effectively not allow salt creating events on the salt_events table, because datetime column is not provided by the salt insert operation, and rails 4 does not allow to set a default value for a column as a function (introduced in rails 5). That's why we need this patch.

I disagree, and think the benefits of merging outweigh the benefits of *not*:

1. You shouldn't be committing a new schema dump to an existing project; any change to schema should be done through migration; if you do generate a new schema dump, unexpected changes should be noticed in code review.
2. This creates an artificial divide between development and production; I haven't seen where this patch is applied in any development environment - merging it in helps close the devel/prod gap.

bear454:
Can someone please add a label that this is for release-2.1 ? I don't have permission to add labels.

jordimassaguerpla:
@bear454 : I think this is the problem https://github.com/rails/rails/issues/27077 

I have not read through the details of it. If you can provide a migration that sets the alter-time as CURRENT_TIMESTAMP as default, then we could get rid of the patch.

bear454:
Since the intended purpose of this PR was to merge existing patches, and one cannot be merged, but needs to be reconstructed, I'm going to remove the commit applying
`1_set_default_salt_events_alter_time_column_value.rpm.patch`, and will submit a separate PR for that when I've obsoleted the patch.

bear454:
MaximilianMeister approved these changes 6 hours ago:
> @bear454 thanks, can you please forward-port this also to 3.0 and to master ?

@MaximilianMeister - this doesn't apply to release-3.0 or master. The patch was already removed before the 3.0 fork.

MaximilianMeister:
> this doesn't apply to release-3.0 or master. The patch was already removed before the 3.0 fork.

oops, right :+1:

MaximilianMeister:
can't see how the jenkins failure is related, but it failed 2 times like that already

@bear454 as we dropped the patch, could you just rename https://github.com/kubic-project/velum/blob/release-2.1/packaging/suse/patches/1_set_default_salt_events_alter_time_column_value.rpm.patch to `0_set_default_salt_events_alter_time_column_value.rpm.patch` - we have some packaging magic there with ascending patch numbers

jordimassaguerpla:
@bear454 : I see there is a conflict with a file. Could you fix it?

MaximilianMeister:
> Conflicting files
> packaging/suse/patches/1_set_default_salt_events_alter_time_column_value.rpm.patch

how can that be when the file doesnt exist anymore? :confused: 

i think we can ignore that one

bear454:
@jordimassaguerpla @MaximilianMeister it may be that the patch changed; and therefore my rename introduces an inconsistency, but I do not have sufficient permissions on this project  to even see the conflict. :/

screenshot from 2018-05-16 07-41-54

bear454:
**DO OVER**

flavio:
Done!

PR #520, opened 2018-05-11
Mark admin as updated when we are going to reboot it

ereslibre:
This ensures that when the admin is back, the status of the minion on
the database will match the grains from the very beginning, without
having to rely on the background task updating all minions (including
the admin) update status.

Fixes: bsc#1092910


ereslibre:
[Bug 1092910 - When updating the admin node, it might show that it has updates after it reboots](https://trello.com/c/pfuc4U97/872-bug-1092910-when-updating-the-admin-node-it-might-show-that-it-has-updates-after-it-reboots)

PR #522, opened 2018-05-11
[bsc#1087447] validate pillars

MaximilianMeister:
invalidpillar

### DIFF
--- a/app/models/pillar.rb
+++ b/app/models/pillar.rb
@@ -2,7 +2,7 @@
 # rubocop:disable Metrics/ClassLength
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
-  validates :value, presence: true
+  validates :value, presence: true, format: { with: /\A[\S]+(?: \S+)*\z/ }
 
   scope :global, -> { where minion_id: nil }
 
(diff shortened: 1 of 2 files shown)


app/models/pillar.rb
```diff
@@ -2,7 +2,7 @@
 # rubocop:disable Metrics/ClassLength
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
-  validates :value, presence: true
+  validates :value, presence: true, format: { with: /\A[\S]+\z/ }
```
bergmannf: Wouldn't this regular expression force the value to not contain any spaces at all?

That might be problematic for the `username` and `password` of the cloud configurations.

MaximilianMeister: right, i'll update it

app/models/pillar.rb
```diff
@@ -2,7 +2,7 @@
 # rubocop:disable Metrics/ClassLength
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
-  validates :value, presence: true
+  validates :value, presence: true, format: { with: /\A[\S].*[\S]\z/ }
```
bergmannf: The second `\[S]` forces every entry to have at least 2 characters, which currently breaks on the pillar `lb_mon_retries: "3"` I think.

vitoravelino:
In the past Rails projects I've worked on we always used https://github.com/rmm5t/strip_attributes to solve this issue. But I'm not sure if it would be fine for Velum since is another gem.

jordimassaguerpla:
I don't see this gem in obs://devel:languages:ruby:extensions. If you want to use that gem, we need to package that one.

I don't see any issue using that, but this will get blocked until we have the rpm.

app/models/pillar.rb
MaximilianMeister: updated, can you test again?

MaximilianMeister:
> I don't see any issue using that, but this will get blocked until we have the rpm.

probably there's no issue, but i think there is also no benefit in functionality. if we use this validator for other fields as well, we might reconsider, but i'd say let's do the minimal now, and save us some gem maintenance

PR #523, opened 2018-05-15
emphasize settings apply workflow

MaximilianMeister:
the warning that any changes are not immediately applied should be
more visible

a blue flash message background with a blue button is not visible
enough

settings-ui#apply

### before

apply-before

### after

apply-danger

### alternatively

apply-warning

apply-red-blue

warning

### DIFF
--- a/app/assets/stylesheets/velum_bootstrap-variables.sass
+++ b/app/assets/stylesheets/velum_bootstrap-variables.sass
@@ -21,7 +21,6 @@
 $brand-success:         #04a36f
 // $brand-info:            #5bc0de
 // $brand-warning:         #f0ad4e
-$brand-warning:         #d9534f
 // $brand-danger:          #d9534f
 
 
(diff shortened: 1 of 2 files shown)


vitoravelino:
I personally don't like the red, maybe the yellow color [0]? Remember to change the button color too.

[0] https://getbootstrap.com/docs/3.3/components/#alerts

MaximilianMeister:
warning

this is with `alert-warning` and `btn-warning` not sure why the button is red

MaximilianMeister:
> not sure why the button is red

had to do some css fixes locally, but havent pushed them yet

apply-warning

flavio:
I don't like the "all red" one.

I would choose one between the "yellow + orange" and "yellow + red"

flavio:
Let's go with yellow + red as discussed on irc with @MaximilianMeister. Apparently that's also @vitoravelino's favorite

MaximilianMeister:
updated, @vitoravelino please have another look

PR #528, opened 2018-05-15
Allow users to choose flannel's backend

flavio:
Starting with v3 we changed the default flannel backend from UDP to
VXLAN.

The VXLAN backend does not work when users have segmented networks,
while the UDP one does.

The salt states are already capable of switching between the two
backends, we just needed to expose the right pillars from Velum.

This commit changes the setup UI to allow users to choose the backend of
and the port used by flannel.

Mandatory screenshot

screenshot-2018-5-15 velum

feature#flannel-backend

### DIFF
--- a/app/controllers/setup_controller.rb
+++ b/app/controllers/setup_controller.rb
@@ -49,6 +49,10 @@ class SetupController < ApplicationController
     @cloud_openstack_lb_mon_retries = Pillar.value(pillar: :cloud_openstack_lb_mon_retries) || "3"
     @cloud_openstack_bs_version = Pillar.value(pillar: :cloud_openstack_bs_version) || "v2"
 
+    # flannel settings
+    @flannel_backend = Pillar.value(pillar: :flannel_backend) || "vxlan"
+    @flannel_port = Pillar.value(pillar: :flannel_port) || "8472"
+
     # container runtime setting
     @cri = Pillar.value(pillar: :container_runtime) || "docker"
   end
--- a/app/views/setup/welcome.html.slim
+++ b/app/views/setup/welcome.html.slim
@@ -70,6 +70,24 @@ h1 Initial CaaS Platform Configuration
 
         hr
 
+        .form-group
+          = f.label :flannel_backend, "Flannel backend"
+          = f.select :flannel_backend, options_for_select([["VXLAN", "vxlan"], ["UDP", "udp"]], @flannel_backend), {}, {class: "form-control", 'aria-describedby' => "flannel_backend_help"}
+          small.form-text.text-muted#flannel_backend_help
+            | The backend used by Flannel to encapsulate network traffic. VXLAN is the recommended choice.
+
+        .form-group
+          = f.label :flannel_port, "Flannel port"
+          = f.text_field :flannel_port, value: @flannel_port, class: "form-control", required: true, 'aria-describedby' => "flannel_port_help"
+          small.form-text.text-muted#flannel_port_help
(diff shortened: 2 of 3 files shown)


flavio:
CC @ajaeger - as you know this was requested by some customers

inercia:
What about adding some other flannel backends? Maybe we could have [the three recommended backends](https://github.com/coreos/flannel/blob/master/Documentation/backends.md#recommended-backends), `VXLAN`, `UDP` and `host-gw`...

app/controllers/setup_controller.rb
```diff
@@ -49,6 +49,10 @@ def welcome
     @cloud_openstack_lb_mon_retries = Pillar.value(pillar: :cloud_openstack_lb_mon_retries) || "3"
     @cloud_openstack_bs_version = Pillar.value(pillar: :cloud_openstack_bs_version) || "v2"
 
+    # flannel settings
+    @flannel_backend = Pillar.value(pillar: :flannel_backend) || "vxlan"
```
inercia: Is `"vxlan"` a port number?

app/views/setup/welcome.html.slim
```diff
@@ -70,6 +70,24 @@ h1 Initial CaaS Platform Configuration
...
+          = f.label :flannel_backend, "Flannel backend"
+          = f.select :flannel_backend, options_for_select([["VXLAN", "vxlan"], ["UDP", "udp"]], @flannel_backend), {}, {class: "form-control", 'aria-describedby' => "flannel_backend_help"}
+          small.form-text.text-muted#flannel_backend_help
+            | The backend used by Flannel to encapsulate network traffic. VXLAN is the recommended choice.
+
+        .form-group
+          = f.label :flannel_port, "Flannel port"
+          = f.text_field :flannel_port, value: @flannel_port, class: "form-control", required: true, 'aria-describedby' => "flannel_port_help"
+          small.form-text.text-muted#flannel_port_help
+            | The port used by Flanne to encapsulate network traffic.
```
MaximilianMeister: Flannel

app/views/setup/welcome.html.slim
```diff
@@ -70,6 +70,24 @@ h1 Initial CaaS Platform Configuration
 
         hr
 
+        .form-group
+          = f.label :flannel_backend, "Flannel backend"
+          = f.select :flannel_backend, options_for_select([["VXLAN", "vxlan"], ["UDP", "udp"]], @flannel_backend), {}, {class: "form-control", 'aria-describedby' => "flannel_backend_help"}
```
MaximilianMeister: i just looked at the screenshot, it doesnt seem to be a bootstrap-styled drop down, or is that only your browser?

flavio:
> What about adding some other flannel backends? Maybe we could have the three recommended backends, VXLAN, UDP and host-gw...

@inercia we can definitely do that, but afaik we only tested the vxlan and udp ones in the past. If we want to, we can do that when there's no "beta-induced" pressure :)

app/views/setup/welcome.html.slim
flavio: Good catch. Fixed

app/views/setup/welcome.html.slim
flavio: On my computer looks like the select box from [here](http://getbootstrap.com/docs/4.1/components/forms/#form-controls)

flavio:
I did a forced merge to fix the typo pointed out by @MaximilianMeister. Merging

PR #529, opened 2018-05-16
Add self-signed certificate form to bootstrap setup

bergmannf:
Based on #504.

This adds a new form to the initial bootstrap page that allows the user to add a system-wide certificate.

I put a general form on the top-level, but maybe it should be nested under the `CPI` form?


bergmannf:
[[P1] Bug 1090067 - Providing self-signed CA for CaaSP](https://trello.com/c/nNv85jtT/808-p1-bug-1090067-providing-self-signed-ca-for-caasp)

bergmannf:
Merged changes into #504.

PR #542, opened 2018-05-21
js: removal of pending acceptance state

vitoravelino:
We are storing the minion id in session storage while it goes under
pending acceptance state. However we were not handling its removal if
the acceptance succeeded or if an error happened after the request.

This patch fixes those scenarios avoiding unwanted user experiences.

bsc#1093869

(cherry picked from commit a660c5b273befe3427f04036719c035fcf0635e1)


vitoravelino:
[Bug 1093869 - New nodes never leave "Acceptance in progress"](https://trello.com/c/V98G6tdo/907-bug-1093869-new-nodes-never-leave-acceptance-in-progress)

PR #547, opened 2018-05-22
ui: fixed unwanted acceptance failure alert 

vitoravelino:
Also replaced deprecated ajax callback method.

bsc#1093869 follow-up

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -501,8 +501,9 @@ function requestMinionApproval(selector, minionIds) {
   $.ajax({
     url: '/accept-minion.json',
     method: 'POST',
-    data: { minion_id: selector }
-  }).error(function () {
+    data: { minion_id: selector },
+    dataType: 'text'
+  }).fail(function () {
     $alert.remove();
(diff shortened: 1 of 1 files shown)


vitoravelino:
[Bug 1093869 - New nodes never leave "Acceptance in progress"](https://trello.com/c/V98G6tdo/907-bug-1093869-new-nodes-never-leave-acceptance-in-progress)

app/controllers/salt_controller.rb
```diff
@@ -17,7 +17,7 @@ def accept_minion
 
     respond_to do |format|
       format.html { redirect_to root_path }
-      format.json { head :ok }
+      format.json { render nothing: true, status: :accepted }
```
ereslibre: I think `:ok` resembles better what happens here in reality than `accepted`. `accepted` has the connotation of having added the task to a job queue. In this case, the accept process happens synchronously, and we are actually waiting for the `salt-api` to return to us.

vitoravelino: I thought that it would be fine to use since we delegate the task to salt-api and return always a success response. I'll revert the change since it won't break the fix anyway.

vitoravelino: Done.

PR #550, opened 2018-05-24
read distribution from a file

MaximilianMeister:
this enables branding for kubic and caasp

feature#branding

~~still depends on the availability of the [velum-branding package](https://github.com/kubic-project/velum-branding) in the product~~ should be in the devel image now

## needs to be merged with

https://github.com/kubic-project/caasp-container-manifests/pull/183


app/views/dashboard/index.html.slim
```diff
@@ -1,7 +1,7 @@
 .alert.alert-warning.unsupported-alert role="alert" hidden="true"
   i.fa.fa-4x.pull-left aria-hidden="true"
   span
-    | A supported deployment of SUSE CaaS Platform requires&nbsp;
+    | A supported deployment of #{Rails.root.join("PRODUCT").read} requires&nbsp;
```
vitoravelino: To avoid reading the file every time it would be nice to cache the info and extract it to a helper method.

vitoravelino:
I was wondering if a file with a single info there enough... Some places we use `CaaS Platform` and others 
`SUSE CaaS Platform`. In the footer of the login page we use the `®` symbol. Initially I thought about an yaml file to have more flexibility since we don't know exactly what our needs are gonna be.

app/views/dashboard/index.html.slim
MaximilianMeister: updated

MaximilianMeister:
> I was wondering if a file with a single info there enough... Some places we use CaaS Platform and others
> SUSE CaaS Platform. In the footer of the login page we use the ® symbol. Initially I thought about an yaml file to have more flexibility since we don't know exactly what our needs are gonna be.

good point, i think for now it's fine from my point of view. but this is easy to adapt once we have a different requirement

for the symbol i opened https://github.com/kubic-project/velum-branding/pull/1

MaximilianMeister:
> There still some #{Rails.root.join("PRODUCT")} around.

thanks, it was wrong anyways, and so my `sed` regex didnt catch it :-( fixed now

app/helpers/application_helper.rb
```diff
@@ -45,4 +45,12 @@ def error_messages_for(model, field)
...
     end
   end
+
+  def product_name
+    @product_name ||= Rails.root.join("PRODUCT").read
+  end
+
+  def product_version
+    @product_version ||= Rails.root.join("VERSION").read
+  end
```
ereslibre: Instead of reading this on a helper every time I think we should save this using an initializer (http://guides.rubyonrails.org/configuring.html#custom-configuration), so we only read this once when we start the process.

README.md
```diff
@@ -4,17 +4,17 @@
...
 - Bootstrap a Kubernetes cluster with a simple click.
 - Manage your Kubernetes cluster: adding and removing nodes from your cluster,
   monitoring faulty nodes, etc.
-- Setup an update policy that suits your needs. SUSE CaaS Platform already provides a
+- Setup an update policy that suits your needs. Kubic/SUSE CaaS Platform already provides a
   transparent and sensible procedure for updates that guarantees no downtime,
   but with Velum you will be able to further tune this.
 
-The architecture of CaaS Platform uses [Salt](https://saltstack.com/) quite heavily,
+The architecture of Kubic/CaaS Platform uses [Salt](https://saltstack.com/) quite heavily,
```
ereslibre: Can't we use `#{product_name}` here too? I guess `PRODUCT` file should contain `SUSE CaaS Platform` inside caasp right?

MaximilianMeister: Is that possible with markdown to read file content and expand it in the rendered view?

app/helpers/application_helper.rb
MaximilianMeister: good point, updated

README.md
ereslibre: I completely missed this was the README file, I'm sorry.

MaximilianMeister:
just tested it in combination with https://github.com/kubic-project/caasp-container-manifests/pull/183 and it looks good now that we have the branding in the devel images installed by default

PR #557, opened 2018-05-24
ux: toggle domain/project id/name inputs

vitoravelino:
Whenever the user wants to configure openstack domain/project id/name,
they cannot fill both cases. To prevent that we are disabling the
respective opposite fields when one of the fields is filled.

bsc#1091809 ui improvement

peek 2018-05-24 15-35

### DIFF
--- a/app/assets/javascripts/setup/openstack.js
+++ b/app/assets/javascripts/setup/openstack.js
@@ -1,0 +1,36 @@
+(function (window) {
+  var dom = {
+    NAME_INPUTS: '#settings_cloud_openstack_domain, #settings_cloud_openstack_project',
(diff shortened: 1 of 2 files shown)


vitoravelino:
I think it would be nice to add some brief explanation on the top of the openstack settings panel. Help with that is appreciated since I don't know well the terms.

vitoravelino:
[[P1] Bug 1091809 - openstack CPI authentication with tenant-name and domain-name fails on engcloud.prv.suse.net](https://trello.com/c/zRnJUMrG/841-p1-bug-1091809-openstack-cpi-authentication-with-tenant-name-and-domain-name-fails-on-engcloudprvsusenet)

PR #560, opened 2018-05-28
Revert "Allow users to choose flannel's backend"

flavio:
This reverts commit 99550aacb6b7c5bfa48545005039c743aa2ee853.

It turns out it's possible to use the VXLAN backend to create clusters
over segmented networks.

feature#revert-UDP-selection-UI


flavio:
Yes, we want to revert that on `master` as well

PR #564, opened 2018-05-30
Some fixes on the eviction hard validation controller

mssola:
This PR includes two fixes:

1. A fix for bsc#1093894: it adds a validator which is checked only for `kubelet:eviction-hard` pillars.
2. A fix for the `create` method in which the transaction was simply being ignored.


mssola:
Note that this is still on progress since there are some tests failing

app/controllers/settings/kubelet_compute_resources_reservations_controller.rb
```diff
@@ -18,27 +18,18 @@ def create
...
+      @kube_reservations.update_attributes!(kube_reservation_params)
+      @system_reservations.update_attributes!(system_reservation_params)
+      @eviction_hard.update_or_remove!(eviction_hard_param)
     end
 
     redirect_to settings_kubelet_compute_resources_reservations_path,
       notice: "kubelet resource reservations successfully saved."
-  rescue ActiveRecord::RecordInvalid
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.tagged(:kubelet_compute_resources_reservaction) { Rails.logger.error e.message }
```
ereslibre: Typo? `reservaction`?

app/models/pillar.rb
```diff
@@ -2,12 +2,18 @@
...
 
   scope :global, -> { where minion_id: nil }
 
   PROTECTED_PILLARS = [:dashboard, :apiserver, :dashboard_external_fqdn].freeze
 
+  # update_or_remove! updates the current pillar with the given value. That
+  # being said, if the given value is blank, then it will destroy this pillar.
+  def update_or_remove!(val)
+    val.blank? ? destroy : update_attributes!(value: val)
+  end
```
ereslibre: Isn't this method overlapping logic with https://github.com/mssola/velum/blob/ce4da5af9c710cf4068985a3b1b69488936e3a9f/app/models/pillar.rb#L149-L150?

app/models/pillar.rb
```diff
@@ -2,12 +2,18 @@
 # rubocop:disable Metrics/ClassLength
 class Pillar < ApplicationRecord
   validates :pillar, presence: true
-  validates :value, presence: true
+  validates :value, presence: true, eviction: true
```
ereslibre: This is why I never liked this flat model that we needed when the `pillars` table was shared with salt (nowadays this is no longer the case). I'd say moving this to an explicit model would be cleaner so we can properly add the validation for that model specifically, but to be honest... at this point this is still fine, and probably not worth the refactor.

app/models/pillar.rb
mssola: Not exactly. This one will always raise an exception on error, while the other simply calls `save` which does not raise an exception. This is important because this way we'll rollback the transaction on the controller.

app/controllers/settings/kubelet_compute_resources_reservations_controller.rb
mssola: Fixed!

app/models/pillar.rb
ereslibre: Thank you!

flavio:
@mssola yes, please backport that to the 3.0 branch

PR #565, opened 2018-05-30
ui: show proper update failed message and reboot button

vitoravelino:
Changed update failed error message stating a generic state instead of
saying that the update failed itself.

Update admin node link only showing if `tx_update_reboot_needed` flag is
truthy. Otherwise keeps it hidden.

bsc#1082786


vitoravelino:
[Bug 1082786 - Wrong information in error case in velum about update](https://trello.com/c/zMzHEpjG/974-bug-1082786-wrong-information-in-error-case-in-velum-about-update)

vitoravelino:
>  I think we want a backport to 3.0 too.

Yes. I was waiting an approval. 

Going to fix rubocop offenses and backport it.

PR #574, opened 2018-06-04
[3.0] ui: do not count failed removal as minion for removal

vitoravelino:
We should not consider a failed removal minion as a valid removal when
we decide to show/hide the removals link.

In this case, if we are in a scenario 1+2 and one of the workers is in
removal failed highstate, we should not show the 'Remove' link for the
other worker.

bsc#1095205

(cherry picked from commit 1137f43166f137408e76219c6e544e14b7d2510e)

Backport of #573

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -467,8 +467,14 @@ MinionPoller = {
 };
 
 function isTheLast(minion, role) {
+  var remainingValid = State.minions.filter(function (m) {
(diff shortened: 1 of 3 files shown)


vitoravelino:
[Bug 1095205 - Do not offer removal in the UI if one of the 2 remaining workers/masters is in falied state](https://trello.com/c/ZTL9cDm4/986-bug-1095205-do-not-offer-removal-in-the-ui-if-one-of-the-2-remaining-workers-masters-is-in-falied-state)

PR #575, opened 2018-06-05
Fix bsc#1072242: Use keyboard layout from admin installation.

bergmannf:
The keyboard layout the user was using during the installation will be
used in the autoyast generated to bootstrap additional nodes.

Requires kubic-project/caasp-container-manifests#191

Signed-off-by: Florian Bergmann <fbergmann@suse.de>

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -18,6 +18,9 @@ class DashboardController < ApplicationController
 
   before_action :redirect_to_dashboard, only: :unassigned_nodes
 
+  DEFAULT_KEYBOARD_LAYOUT = "english-us".freeze
+  YAST_KEYBOARD_KEY = "YAST_KEYBOARD".freeze
+
   # Return the autoyast XML profile to bootstrap other worker nodes. They will read this response in
   # order to start an unattended cluster installation.
   #
@@ -48,6 +51,7 @@ class DashboardController < ApplicationController
       ssh_key_file = "/var/lib/misc/ssh-public-key/id_rsa.pub"
       # rubocop:disable Style/RescueModifier
       @ssh_public_key = File.read(ssh_key_file) rescue nil
+      @keyboard_layout = read_keyboard_layout
       # rubocop:enable Style/RescueModifier
 
       # proxy related settings
@@ -58,6 +62,20 @@ class DashboardController < ApplicationController
 
       render "autoyast.xml.erb", layout: false, content_type: "text/xml"
     end
+  end
+
+  # Read the keyboard layout set by Yast during installation.
+  #
+  # @return [String] the keyboard set from Yast, or 'english-us' as
+  # default
+  def read_keyboard_layout(keyboard_config_file: "/var/lib/misc/keyboard")
+    return DEFAULT_KEYBOARD_LAYOUT unless File.file?(keyboard_config_file)
+    yast_layout = File.readlines(keyboard_config_file).select do |line|
+      line =~ /^#{YAST_KEYBOARD_KEY}=/
+    end.first
+    return DEFAULT_KEYBOARD_LAYOUT unless yast_layout
+    layout, _char_map = yast_layout.split("=")[1].delete('"').split(",")
+    layout ? layout : DEFAULT_KEYBOARD_LAYOUT
   end
 
   # GET /assign_nodes
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,4 +1,5 @@
 require "velum/suse_connect"
+require "velum/autoyast_values"
 
 # DashboardController shows the main page.
 class DashboardController < ApplicationController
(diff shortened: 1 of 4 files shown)


bergmannf:
[Bug 1072242 - Keyboard selection not reflected in AutoYast file](https://trello.com/c/PCwK3MZz/999-bug-1072242-keyboard-selection-not-reflected-in-autoyast-file)

app/controllers/dashboard_controller.rb
```diff
@@ -60,6 +61,22 @@ def autoyast
     end
   end
 
+  # Read the keyboard layout set by Yast during installation.
+  #
+  # @return [String] the keyboard set from Yast, or 'english-us' as
+  # default
+  def read_keyboard_layout(keyboard_config_file: "/var/lib/misc/keyboard")
+    default_keyboard_layout = "english-us"
+    yast_keyboard_key = "YAST_KEYBOARD"
```
flavio: I would make both `yast_keyboard_key` and `default_keyboard_layout` constants and move them at the top of the class definition.

app/controllers/autoyast_values.rb
```diff
@@ -0,0 +1,47 @@
+# Load values that will be used to populate the autoyast profile
+class AutoyastValues
```
flavio: That's not a controller. I propose to move that to [lib/velum](https://github.com/kubic-project/velum/tree/master/lib/velum).

What does others like @ereslibre, @MaximilianMeister think about?

lib/velum/autoyast_values.rb
```diff
@@ -0,0 +1,47 @@
+# Load values that will be used to populate the autoyast profile
+class AutoyastValues
```
MaximilianMeister: let's wrap that into a `Velum` namespace like in all the other libs?

```ruby
module Velum
  class AutoyastValues
    ..
  end
end
```

app/controllers/dashboard_controller.rb
```diff
@@ -29,33 +30,11 @@ class DashboardController < ApplicationController
   # fields that customer uses) and also skips the redirection to the setup process (when a worker
   # asks for the autoyast profile we will either serve it, or return an error).
   def autoyast
-    @controller_node = Pillar.value pillar: :dashboard
+    autoyast = AutoyastValues.new
```
MaximilianMeister: we might not need to cache this in a variable do we? `pollute_variables!(AutoyastValues.new)`

app/controllers/dashboard_controller.rb
```diff
@@ -97,4 +76,12 @@ def host_warning
                     "It is advised to access velum via one of the registered hosts " \
                     "#{accessible_hosts.join(" or ")}"
   end
+
+  # Pollutes the instance variables of this object by copying over all the
+  # variables of the parameter
+  def pollute_variables!(other)
+    other.instance_variables.each do |var|
+      instance_variable_set(var, other.instance_variable_get(var))
```
MaximilianMeister: could you clarify a bit more what this is supposed to do? to me it looks like it's reading the instance variables and sets them again to the same value. is this to "bring them over" from the lib to the controller namespace? i dont know if that's needed, or if they're accessible during runtime anyways without setting them again. i'd add a clearer comment because this is meta hacking ;-)

the parameter name `other` is also a bit unclear to me still, maybe it's a bit too generic

MaximilianMeister: `var` might also need to be symbolized, looking at https://apidock.com/ruby/Object/instance_variable_set but i havent tested it yet

bergmannf: `object.instance_variables` already returns the symbolized form, so it should be fine:

```ruby
[1] pry(main)> class A
[1] pry(main)*   def initialize
[1] pry(main)*     @a = "a"
[1] pry(main)*     @b = "b"
[1] pry(main)*   end  
[1] pry(main)* end  
=> :initialize
[2] pry(main)> a.instance_variables
=> [:@a, :@b]
```

And I really dislike the `meta-programming` copying - but there are some tests that really expect the templates to render with only the variable names. I added this in a comment, but maybe it is not worth to hack it this way, just for those tests and we could remove them?

Then I would just assign `@autoyast = ...` and update the template to render the variables with an `autoyast` prefix.

lib/velum/autoyast_values.rb
bergmannf: Forgot that, when moving the file - I added the `module`.

app/controllers/dashboard_controller.rb
```diff
@@ -97,4 +75,21 @@ def host_warning
...
+
+  # This method will copy *all* instance variables of the given object to the
+  # current instance. No checks will be performed if a variable already exists -
+  # the current value will be overwritten in that case.
+  #
+  # This copying right now is used to prevent nesting values in the template:
+  # instead of writing `<%= other.var %>` it is possible to just write `<%= var
+  # %>` again. It was solely added as there are tests that expect this syntax,
+  # when rendering a template, and would fail otherwise:
+  # /spec/views/dashboard/autoyast.xml.slim_spec.rb
```
MaximilianMeister: the way you do it is correct i guess, and does not only apply to the test, but also to the rendering itself (e.g. when visiting `/autoyast`). so i think i'd remove the bit about the test again

MaximilianMeister:
dependent PR is merged, merging this so we'll likely have both changes in the new devel image

PR #579, opened 2018-06-07
Fix public assets not in iso

jordimassaguerpla:
This way this error message can be used both in kubic and in caasp

This is needed because we need to revert a change in the velum-branding
package and the caasp-container-manifests that was mounting the public
directory from the host.

Mounting the public directory from the host, was "hidding" the assets files
present in the public directory and so our images were broken.

This was only happening in the iso and not in our dev environments
because in production we have the assets precompiled and we don't
compile them on the fly, since we don't have a js engine installed.

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>


vitoravelino:
I guess this page is not being used anymore after we moved to haproxy. There are already 503 files inside velum-branding repo.

vitoravelino:
Forget my comment. I just read your message explaining why. My bad.

jordimassaguerpla:
this goes with https://github.com/kubic-project/caasp-container-manifests/pull/192 and https://github.com/kubic-project/velum-branding/pull/5

public/503_database_not_ready.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>SUSE CaaS Platform is initializing</title>
+  <title>Platform is initializing</title>
```
bergmannf: Maybe we can put `Platform still initializing` instead of `is` - I feel it sounds better?

public/503_database_not_ready.html
```diff
@@ -58,10 +58,10 @@
 </head>
 
 <body>
-  <!-- This file lives in public/500.html -->
+  <!-- This file lives in public/503.html -->
   <div class="dialog">
     <div>
-      <h1>SUSE CaaS Platform is initializing</h1>
+      <h1>Platform is initializing</h1>
```
bergmannf: I would also put `still initializing` here.

bergmannf:
To make the tests work, the image files need to be moved:

```
renamed    app/assets/images/logo-footer.png -> app/assets/images/branding/logo-footer.png
renamed    app/assets/images/logo.svg -> app/assets/images/branding/logo.svg
```

public/503_database_not_ready.html
MaximilianMeister: i guess we can safely drop the static pages/assets from velum entirely, as we have them in `velum-branding` so we dont need to duplicate them

flavio: I would use `Platform is being initialized`

MaximilianMeister:
what i still dont undestand is why are the compiled assets in `/public` ? the default configuration should put them under `assets` according to the [docs](http://guides.rubyonrails.org/asset_pipeline.html#precompiling-assets) and we haven't changed the default via `config.assets.prefix`. i haven't looked yet at an installation from ISO but i believe the public folder should only contain static files, and no compiled assets

the static error pages should not be an issue because they're static and not precompiled

we might need to fix this in the velum packaging, possibly by requiring `velum-branding` and move the branded assets (logos) to the right place before running `rake assets:precompile`

MaximilianMeister:
also we compile during runtime as well, for whatever reason... see https://github.com/kubic-project/velum/blob/master/config/environments/production.rb#L29

so effectively the precompilation is useless because of this option. we should probably set that to false as well

MaximilianMeister:
a suggestion for both the static pages, and the assets. we could:

### static error pages

* set `paths["public"] = "static-pages"` in https://github.com/kubic-project/velum/blob/master/config/application.rb to register a new subfolder which we could then mount under `/srv/velum/public/static-pages`
* drop the static pages in git/velum and let `velum-branding` only provide them

### assets (images)

* BuildRequire `velum-branding` in `velum`
* copy the branded assets (only logos for now) to the right place (`app/assets/images`) during the build
* and then run `rake assets:precompile`
* drop the logos from `app/assets/images` in git/velum and let only `velum-branding` provide them

flavio:
I guess with "drop" you mean remove these files from git. If that's the case I would be a bit concerned about how to keep track of their changes. OBS doesn't have the best interface to collaborate on that IMHO

MaximilianMeister:
> I guess with "drop" you mean remove these files from git. If that's the case I would be a bit concerned about how to keep track of their changes. OBS doesn't have the best interface to collaborate on that IMHO

@flavio they'd still be tracked in git, just in a different repo (velum-branding). but thinking about it again, we might keep them also in velum with generic content as @jordimassaguerpla proposed, as they are not really essential pieces of the stack, and therefore don't need too much branding anyways

PR #583, opened 2018-06-12
make 503 page more generic

MaximilianMeister:
bsc#1080636

### DIFF
--- a/public/503.html
+++ b/public/503.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>Platform is being initialized</title>
+  <title>Platform is unavailable</title>
   <meta name="viewport" content="width=device-width,initial-scale=1">
   <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
   <style>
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
Could this be backported to release-3.0 branch? Even if it has to be released as a maintenance update.

MaximilianMeister:
> Could this be backported to release-3.0 branch? Even if it has to be released as a maintenance update.

@jordimassaguerpla underlying bug is a P2, and showing a branded page instead of just a 503 Service Unavailable text I think is more of a cosmetic issue. we'd need to backport a few commits from velum/salt/manifests and the branding package itself. do you think it's worth it?

jordimassaguerpla:
Can  we fix the bug for 3.0 without the branding package? I would rather prefer not to add new packages in 3.0

MaximilianMeister:
> Can we fix the bug for 3.0 without the branding package? I would rather prefer not to add new packages in 3.0

@jordimassaguerpla i think there's no clean way, as we need to have the files installed on the admin host. we could install the static pages via another existing package but i dont know which one would fit.

PR #584, opened 2018-06-13
Add a feature spec for building cluster in Azure

bear454:
Most of the content is covered in the cloud_cluster model, but there are a few differentiators that should be checked.


bear454:
[Add a feature spec for building cluster in Azure](https://trello.com/c/ah3QAPqa/1019-add-a-feature-spec-for-building-cluster-in-azure)

PR #585, opened 2018-06-13
Add dex connector ldap table, models, and pillar output

nanoscopic:
Added a new table "dex_connectors_ldap" and migration file to the db schema and migrate files.
Altered pillar controller to output this new data.
Added a basic/empty app model for the new table so that the data can be fetched through it.
Updated pillar rspec to test newly output connectors as well as to allow the empty connectors in the other pillar tests.

### DIFF
--- a/app/controllers/internal_api/v1/pillars_controller.rb
+++ b/app/controllers/internal_api/v1/pillars_controller.rb
@@ -1,3 +1,5 @@
+require "velum/dex/ldap"
+
 # Serve the pillar information
 # rubocop:disable Metrics/ClassLength
 class InternalApi::V1::PillarsController < InternalApiController
@@ -12,6 +14,8 @@ class InternalApi::V1::PillarsController < InternalApiController
       kubelet_contents
     ).merge(
       system_certificate_contents
+    ).deep_merge(
+      dex_connectors_as_pillar
     )
   end
 
@@ -165,6 +169,12 @@ class InternalApi::V1::PillarsController < InternalApiController
         "eviction-hard"     => eviction_hard.value || ""
       }
     }
+  end
+
+  def dex_connectors_as_pillar
+    connectors = []
+    connectors.concat(Velum::Dex.ldap_connectors_as_pillar)
+    { dex: { connectors: connectors } }
   end
 end
 # rubocop:enable Metrics/ClassLength
--- a/lib/velum/dex/ldap.rb
+++ b/lib/velum/dex/ldap.rb
@@ -1,0 +1,68 @@
+require "base64"
+
+module Velum
+  # This class offers the integration between ruby and the Saltstack API.
+  module Dex
+    class << self
+      def ldap_connectors_as_pillar
+        ldap_connectors = DexConnectorLdap.all.map do |con|
+          {
+            type:            "ldap",
+            id:              con.id,
+            name:            con.name,
+
+            # Combine host and port since they ultimately
+            #   feed into a single line of config for dex
+            server:          "#{con.host}:#{con.port}",
+            start_tls:       con.start_tls,
+            root_ca_data:    Base64.encode64(con.certificate.try(:certificate) || ""),
+            bind:            generate_bind_block(con), # Place basic bind information together
+            user:            generate_user_block(con), # Place user stuff together
+            group:           generate_group_block(con), # Place group stuff together
+            username_prompt: con.username_prompt
+          }
+        end
+        ldap_connectors
+      end
+
+      private
+
+      def generate_user_block(con)
+        {
+          base_dn:  con.user_base_dn,
+          filter:   con.user_filter,
+          attr_map: {
+            username: con.user_attr_username,
+            id:       con.user_attr_id,
+            email:    con.user_attr_email,
+            name:     con.user_attr_name
+          }
+        }
+      end
+
+      def generate_bind_block(con)
+        bind = {}
+        if con.bind_anon
+          bind[:anonymous] = true
+        else
+          bind[:anonymous] = false
+          bind[:dn] = con.bind_dn
+          bind[:pw] = con.bind_pw
+        end
+        bind
+      end
+
+      def generate_group_block(con)
+        {
+          base_dn:  con.group_base_dn,
+          filter:   con.group_filter,
+          attr_map: {
+            user:  con.group_attr_user,
+            group: con.group_attr_group,
+            name:  con.group_attr_group
+          }
+        }
+      end
+    end
+  end
+end
--- a/app/models/dex_connector_ldap.rb
+++ b/app/models/dex_connector_ldap.rb
@@ -1,0 +1,6 @@
+# Model that represents a dex authentication connector for LDAP
+class DexConnectorLdap < ActiveRecord::Base
+  has_one :certificate_service, as: :service, dependent: :destroy
+  has_one :certificate, through: :certificate_service
+  self.table_name = "dex_connectors_ldap"
+end
--- a/db/migrate/20180530105000_create_dex_connectors_ldap.rb
+++ b/db/migrate/20180530105000_create_dex_connectors_ldap.rb
@@ -1,0 +1,26 @@
+class CreateDexConnectorsLdap < ActiveRecord::Migration
+  def change
+    create_table :dex_connectors_ldap do |t|
+      t.timestamps
+      t.string    :name,               limit: 255
+      t.string    :host,               limit: 255
+      t.integer   :port,               limit: 2,   default: 636
+      t.boolean   :start_tls,                      default: false, null: false
+      t.boolean   :bind_anon,                      default: false, null: false # bind_dn and bind_pw ignored if true
+      t.string    :bind_dn,            limit: 255, default: "uid=someuid,cn=users,dc=somedomain,dc=com"
+      t.string    :bind_pw,            limit: 255
+      t.string    :username_prompt,    limit: 255, default: "Username"
+      t.string    :user_base_dn,       limit: 255, default: "cn=users,dc=somedomain,dc=com"
+      t.string    :user_filter,        limit: 255, default: "(objectClass=person)"
+      t.string    :user_attr_username, limit: 255, default: "uid"
+      t.string    :user_attr_id,       limit: 255, default: "uid"
+      t.string    :user_attr_email,    limit: 255, default: "mail", null: false
+      t.string    :user_attr_name,     limit: 255, default: "name"
+      t.string    :group_base_dn,      limit: 255, default: "cn=groups,dc=somedomain,dc=com"
+      t.string    :group_filter,       limit: 255, default: "(objectClass=group)"
+      t.string    :group_attr_user,    limit: 255, default: "uid"
+      t.string    :group_attr_group,   limit: 255, default: "member"
+      t.string    :group_attr_name,    limit: 255, default: "name"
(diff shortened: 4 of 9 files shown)


nanoscopic:
I have updated the code to use dex namespace properly everywhere ( including to merge it properly into the existing dex output without output json )

Additionally I added 'do' blocks in the rspec which seems to correct the concurrency issues that were happening earlier and cuasing CI to fail.

MaximilianMeister:
> I have updated the code to use dex namespace properly everywhere ( including to merge it properly into the existing dex output without output json )

seemed that worked, now you just need to find the small hole in the coverage to make it green

```
Coverage (99.68%) is below the expected minimum coverage (100.00%).
```

you can go to `https://admin-ip.url/coverage` after running the full rspec testsuite **locally** and you'll see what is missing

nanoscopic:
I have dropped the 'configure_connector' class function from the DexConnectorLdap class. It was not being used by anything, and was unneeded boilerplate that does not belong. That was part of the missing coverage.

I additionally modified the second test case to use anonymous bind to test the single other line not being tested that was added.

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -166,5 +170,60 @@ def kubelet_contents
...
+      bind[:anonymous] = true
+    else
+      bind[:anonymous] = false
+      bind[:dn] = con.bind_dn
+      bind[:pw] = con.bind_pw
+    end
+    bind
+  end
+
+  def generate_group_block(con)
```
MaximilianMeister: can we put `generate_group_block`, `generate_bind_block`, `generate_user_block` into a library e.g. under either `lib/velum/dex/ldap.rb`, `lib/velum/dex_connectors.rb` or even to https://github.com/kubic-project/velum/blob/master/lib/velum/ldap.rb ?

i think we can expect more connectors to be added soon (SAML e.g.), therefore we could put the dex connector business logic already in a good place for not having to move it again later, what do you think?

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -166,5 +170,60 @@ def kubelet_contents
...
+    dex_connectors = DexConnectorLdap.all.map do |con|
+      {
+        "type"            => "ldap",
+        "id"              => con.id,
+        "name"            => con.name,
+
+        # Combine host and port since they ultimately feed into a single line of config for dex
+        "server"          => "#{con.host}:#{con.port}",
+        "start_tls"       => con.start_tls,
+        "root_ca_data"    => Base64.encode64(con.certificate.try(:certificate)),
```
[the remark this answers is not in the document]
nanoscopic: The UI portion that will be stacked on top of this should be created so as to make it unable to save a connector without having a certificate in place. That said, I agree it is a good change to make the rendering code resilient to the case anyway since it is possible in the database. I will make this so.

A bigger issue that is related that I see in the dev environment is that I have to manually create the table for the dex_ldap_connectors to bootstrap my dev cluster. My current understanding is that this is due to the way a static iso is used combined only partially with my local dev velum. I am not clear on this though, as it is weird to me that the table does not get automatically created in dev. I was told that changing the db models file and adding the migration db file should be sufficient when adding a new table. If there is something else that needs to be done I don't know what it is.

I cannot tell what is making the CI test fail now, as I don't know how to get the salt execution results out of a CI run. I suspect it is failing in the same way as my dev cluster; getting stuck because the DB table does not exist. I could handle this by adding guard code around the pillar output able to handle the table not existing, but it should be created. Handling the situation where the table doesn't exist seems like a hack.

app/controllers/internal_api/v1/pillars_controller.rb
nanoscopic: Agreed that it would be cleaner to have the various dex related code pulled out and be separate. It does make sense as well to place them in different files per type of connector, for ease of navigating to the relevant code. I will split it up that way and update this PR.

Yes, we are working on adding OIDC and SAML support as well.

nanoscopic:
Moved dex ldap specific helper functions out of pillar code and into lib/velum/dex. Added a catch block into pillar output to handle non-existence of dex_ldap_connectors table which was breaking CI. Added a default onto certificate handling so that a lack of a certificate doesn't crash the output of pillars.

app/controllers/internal_api/v1/pillars_controller.rb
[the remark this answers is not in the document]
MaximilianMeister: when you create a devenv from scratch, you'll see right at the beginning that a velum development container will be built. there we create the database as well with the new tables. so in theory that should work... but the thing is, you need to create it from scratch

the other solution would be to run the migrations inside the velum container, and then commit that container so the change persists in your env, but i haven't done that yet as it's easier to create all from scratch

CI is green now :+1:

lib/velum/dex/ldap.rb
```diff
@@ -0,0 +1,74 @@
...
+              "username_prompt" => con.username_prompt
+            }
+          end
+        rescue ActiveRecord::StatementInvalid => e
+          # Silently ignore non existence of table so that cluster can start without it
+          # :nocov:
+          raise e unless e.message.include? "dex_connectors_ldap' doesn't exist"
+          # :nocov:
+        end
+        ldap_connectors
```
MaximilianMeister: you could save a few lines of code and dont need to cache the connectors like this:

```ruby
def ldap_connectors_as_pillar
  DexConnectorLdap.all.map do |con|
￼    {
￼       ...
    }
  end
rescue ActiveRecord::StatementInvalid => e
  ...
end
```

but i wonder do we need to have the `rescue` block at all? or is it still a remainder of your issue that the table wasn't create automatically? it's also only catching a very specific exception, which normally is good, but i think the table should always be there, so it's not really needed

nanoscopic: Without this rescue the code fails CI. The table is not created automatically. I don't know why. Once it is known why the table isn't created, and how to make it so that it is, this rescue can be dropped.

I added this rescue primarily to prove that CI was failing specifically because of this, rather than only speculating about it.

Using a rescue within a function without begin is interesting; I agree it would be cleaner. As you state though, this code is entirely unneeded when everything works correctly. It doesn't so it is a hack at the moment.

nanoscopic: Great, that was my speculation that this was true. ( that everything will be fine once the PR is merged )

How am I supposed to get it to pass CI without a rescue block? If it doesn't pass CI the checks on Github will fail, and then merging will be prevented. This is why I added the rescue block.

Once it is accepted once, an additional PR to remove the rescue block would then be able to succeed, correct?

kiall: We can either leave it in, and remove in another PR.. or better yet, fix the Dev tooling to apply the PRs schema.rb and re-run the jobs once that's merged.

I'd be fine either way, but obviously, it'll be best to fix the tooling issue :)

kiall: @MaximilianMeister thoughts? How much effort is fixing the tooling?

(I don't want to block progress on tooling issues for the sake of an excess rescue block!)

nanoscopic:
I have:
* rebased on the latest master updates from velum upstream
* accepted your merge commit and rebase merged it ( so that the UI shows it is merged )
* pulled that done, and rebase squashed the patch into the change

lib/velum/dex/ldap.rb
nanoscopic: Alright; I will drop the rescue block once I see the latest rebase and patch addition pass CI again.

MaximilianMeister: created https://trello.com/c/p9ZNCCss/1054-velum-devenv-enable-testing-database-changes to track this

nanoscopic: Am trying to get a successful integration test run of this. It has failed twice so far. It looks random. Am trying a third time.

nanoscopic:
Passed both CI checks. After passing, dropped the unneeded rescue from the code. I notably will not be able to merge this once CI fails again due to the rescue block / db date difference / etc.

lib/velum/dex/ldap.rb
```diff
@@ -0,0 +1,66 @@
+require "base64"
+
+module Velum
+  # This class offers the integration between ruby and the Saltstack API.
+  module Dex
+    class << self
+      def ldap_connectors_as_pillar
+        ldap_connectors = DexConnectorLdap.all.map do |con|
+          {
+          "type"            => "ldap",
```
MaximilianMeister: the body of the block needs to be indented 2 more spaces, and i also was wondering if we could use the 'new' ruby hash syntax here:

```ruby
{
  type: "ldap",
  id:   con.id,
  ...
}
```

the old way `key => value` is generally outdated and only used when there e.g. is a `-` in the key and the parser can't do it, e.g. this doesn't work:

```ruby
{ symbolized-key: "value" }
```

nanoscopic: I do prefer the new hash syntax, but the rest of velum is written with the old syntax. A quick "grep ' =>' * -RIi | wc -l" across the velum codebase reveals 324 instances of the old syntax.

If we are going to prefer the new style syntax, then I would recommend that all of them be changed simultaneously in a single commit to be rid of the old style everywhere possible.

If we do not want to change them everywhere, why do we want to only use the clean syntax in some places? I feel that it makes the code more confusing if there is a random mixture of both old and new style.

nanoscopic:
Corrected indent such that rubocop passes with no issues.
Additionally changed the ldap.rb file further to use the new hash syntax, enhancing not just the function in question but cleaning up the later functions in the same file as well.

Made the new alterations to use the 201817... date in both schema and patches so that deployment should now pass CI. ( as well as merge cleanly )

nanoscopic:
I have increased the db schema version by 1 to ....34. Accepted the merge commit of that change here in Github.

I think there should be a follow up PR to correct the file with the bad date, rather than continuing to use a corrupt date and incrementing it by 1 for the rest of the year. That seems... very confusing.

I'm not clear on why the other accepted commit used ...33 to begin with. That does not seem like a clean solution. Why not just change the filename then? If it is because existing systems will have a higher number, then could we add a 1 at the far left and still correct the filenames otherwise? I think that would be preferable.

nanoscopic:
Am replacing this PR with a new new to prevent the merge. I don't see how to prevent the additional merge commit no matter what I do at this point without using a different branch. Created a new branch dex_connectors_2 that has it cleanly rebased and will submit that as a new PR.

kiall:
> I'm not clear on why the other accepted commit used ...33 to begin with. That does not seem like a clean solution. Why not just change the filename then?

The risk of breaking customers systems is pretty high if we try and mess about the database migration number. It obviously can be done safely - but the value isn't IMO all that high.

> If it is because existing systems will have a higher number, then could we add a 1 at the far left and still correct the filenames otherwise? I think that would be preferable.

This would lock us into a weird pattern forever, instead of only until the end of the year :)

kiall:
> Am replacing this PR with a new new to prevent the merge. I don't see how to prevent the additional merge commit no matter what I do at this point without using a different branch. Created a new branch dex_connectors_2 that has it cleanly rebased and will submit that as a new PR.

`git config --global pull.rebase true` helps avoid accidentally creating merge commits when you pull.

`git fetch origin && git rebase origin/master` is handy for updating your local branch without ending up with a merge.

PR #586, opened 2018-06-14
Expose pillars managing kubernetes extra args

flavio:
The salt backend already has pillars that can be used to provide extra arguments to the different kubernetes components (api-server, controller-manager, scheduler, kubelet and proxy). See [here](https://github.com/kubic-project/salt/blob/d08a65251009262459fe566a45357d0c54094674/pillar/args.sls#L2-L17).

This commit allows to provide values for these settings. This is something we should **not** expose into the UI. It can be handy with certain customers who are looking for a way to pass not-yet-handled
flags to kubernetes.

BTW, don't be confused by the diff, I just added 4 lines and indented a bunch of others.


flavio:
I've added the backport label, but this should be done after 3.0 is released.

PR #589, opened 2018-06-19
ui: don't show failed alert for repeated orchestration

vitoravelino:
A failed alert is shown whenever one of the nodes fails to bootstrap.
This is an optimization to only show this alert once for a specific
orchestration run. If the user closes the alert, it won't come back
unless a new orchestration happens.

bsc#1097752

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -149,6 +149,7 @@ MinionPoller = {
         }
 
         State.minions = minions;
+        State.lastOrchestrationAt = data.last_orchestration_at;
 
         var pendingStateMinion = minions.find(function (minion) {
           return minion.highstate == "pending";
@@ -328,9 +329,20 @@ MinionPoller = {
   },
 
   alertFailedBootstrap: function() {
-    if (!$('.failed-bootstrap-alert').length) {
-      showAlert('At least one of the nodes is in a failed state. Please run "supportconfig" on the failed node(s) to gather the logs.', 'alert', 'failed-bootstrap-alert');
+    var cachedFailedLastOrchestration = window.localStorage.getItem('failedLastOrchestrationAt');
+
+    if ($('.failed-bootstrap-alert').length ||
+        cachedFailedLastOrchestration === State.lastOrchestrationAt) {
+      return;
     }
+
+    var $alert = showAlert('At least one of the nodes is in a failed state. Please run "supportconfig" on the failed node(s) to gather the logs.', 'alert', 'failed-bootstrap-alert');
+
+    window.localStorage.removeItem('failedLastOrchestrationAt');
+    $alert.on('closed.bs.alert', function () {
+      window.localStorage.setItem('failedLastOrchestrationAt', State.lastOrchestrationAt);
+      $alert.off('closed.bs.alert');
+    })
   },
 
   renderDashboard: function(minion) {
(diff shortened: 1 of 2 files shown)


vitoravelino:
Should we backport this? /cc @jordimassaguerpla

app/controllers/concerns/discovery.rb
```diff
@@ -18,7 +19,8 @@ def discovery
           cloud_jobs_failed:                 SaltJob.failed.count,
           admin:                             Minion.find_by(minion_id: "admin"),
           retryable_bootstrap_orchestration: Orchestration.retryable?(kind: :bootstrap),
-          retryable_upgrade_orchestration:   Orchestration.retryable?(kind: :upgrade)
+          retryable_upgrade_orchestration:   Orchestration.retryable?(kind: :upgrade),
+          last_orchestration_at:             last_orchestration_at
```
MaximilianMeister: this could be a one liner, and no need to cache it in a variable

```ruby
Orchestration.try(:last).try(:created_at)
```

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -328,8 +329,17 @@ MinionPoller = {
   },
 
   alertFailedBootstrap: function() {
-    if (!$('.failed-bootstrap-alert').length) {
-      showAlert('At least one of the nodes is in a failed state. Please run "supportconfig" on the failed node(s) to gather the logs.', 'alert', 'failed-bootstrap-alert');
+    var cachedFailedLastOrchestration = window.localStorage.getItem('failedLastOrchestrationAt');
+
+    if (!$('.failed-bootstrap-alert').length &&
+        cachedFailedLastOrchestration !== State.lastOrchestrationAt) {
```
MaximilianMeister: would it make sense to do an early return here instead of wrapping the code in a conditional block? not sure about js guidelines

vitoravelino: I personally don't have a strong opinion about this. I'd say that if the condition were simple, it's fine to do the early return. In this case we have a double condition and we can't write it as in ruby (return if ...). So I guess it's fine to leave it as it is.

app/controllers/concerns/discovery.rb
vitoravelino: Forgot the `try` method. If we were using ruby 2.5 we could use `.&`. I'll fix that. Thanks!

app/assets/javascripts/dashboard/dashboard.js
bear454: I agree with @MaximilianMeister; I'd prefer to see a guard function here. If Rubocop were able, he'd complain :P

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -328,8 +329,17 @@ MinionPoller = {
...
+      var $alert = showAlert('At least one of the nodes is in a failed state. Please run "supportconfig" on the failed node(s) to gather the logs.', 'alert', 'failed-bootstrap-alert');
+
+      window.localStorage.removeItem('failedLastOrchestrationAt');
+      $alert.on('closed.bs.alert', function () {
```
bear454: This isn't a request for change necessarily, but please be aware that "bs" has a very specific meaning in English vernacular, and it's neither _bootstrap_ nor _build service_. A "bs alert" just makes me snicker ;-)

vitoravelino:
@bear454 @MaximilianMeister Could one of you approve this again so we can merge it? Thanks!

PR #590, opened 2018-06-19
Introduce admission webhook settings page

flavio:
Allow user to enable AdmissionWebhook by using a new dedicated page.

To work this requires the following PR merged into the salt repo: https://github.com/kubic-project/salt/pull/555

The new UI has to be improved by @vitoravelino, right now it looks something like that:

image

The workflow is the following one:
  * go this new page
  * upload key and certificate
  * enable the feature
  * deploy the cluster

Notable things missing:
- [X] Validate input (required)
- [X] Fix UI
- [X] Add tests
- [x] Fix coverage


vitoravelino:
Current state with last commit:

screenshot-20180619174743-431x277
screenshot-20180619174814-670x444
screenshot-20180619174851-673x755

app/assets/javascripts/settings/index.js
```diff
@@ -11,7 +11,7 @@ $(function() {
     new RegistryForm($registryForm);
   }
 
-  if ($systemCertificateForm.length) {
-    new SystemCertificateForm($systemCertificateForm);
```
vitoravelino: There's no `SystemCertificateForm`. I let this pass on a previous PR. Removed it since it was causing a js error.

app/forms/admission_webhook_form.rb
```diff
@@ -0,0 +1,21 @@
+# AdmissionWebhookForm represents a form for AdmissionWebhookController
+class AdmissionWebhookForm
```
vitoravelino: The idea is to simulate a virtual model and take advantage of some nice things of a model like validation.

vitoravelino: This will be useful when we introduce the validation of the certificates itself.

vitoravelino:
@flavio This is good to be reviewed.

flavio:
It looks to me, I like the workflow. However there's a bug:

  * upload key and cert
  * save
  * disable the feature
  * save -> accept the warning message from the popup window
  * reload the page
  * enable the feature -> **the old key and cert are still shown**

Basically we don't actually delete them from the database.

flavio:
Also another bug: it's possible to upload empty files. The validation in place doesn't look for that

vitoravelino:
Fixed both cases, @flavio. Thanks!

vitoravelino:
screenshot-20180622141838-295x229

app/views/settings/admission_webhook/index.html.slim
```diff
@@ -0,0 +1,65 @@
+= render 'settings/apply'
+
+.alert.alert-warning role='alert'
+  p Warning: Entering invalid values for any of the following settings will cause
+    the nodes to enter into a broken state.
+
+h2 Admission Webhook
+
+p
+  | Enable Kubernetes ValidatingAdmissionWebhook admission controller. Once enabled
```
MaximilianMeister: you need to add a space at $ or at the beginning of the next line, otherwise it reads `Once enabledit is possible to...`

app/controllers/settings/admission_webhook_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  private
+
+  def set_instance_variables
+    @admission_webhook_enabled = Pillar.value(pillar: :api_admission_webhook_enabled) == "true"
+    @certificate = Pillar.value(pillar: :api_admission_webhook_cert)
+    @key = Pillar.value(pillar: :api_admission_webhook_key)
+  end
+
+  def admission_webhook_params
```
MaximilianMeister: can you still make sure to filter the cert/key from the logs in `config/initializers/filter_parameter_logging.rb` ?

vitoravelino: I thought about that but for file inputs the content doesn't appear as for text inputs. I can paste the content here if you want. 

If we still think that it's necessary to filter it, I can add it.

app/views/settings/admission_webhook/index.html.slim
vitoravelino: Thanks!

app/controllers/settings/admission_webhook_controller.rb
MaximilianMeister: hmm in my case it appears, but it's the devenv. if it doesnt appear in production it's totally fine ;)

vitoravelino: Oh. You mean the `current_cert` and `current_key`. Yeah, you are right. I thought that it was the `key` and `certificate` inputs. :sweat_smile:

vitoravelino: Fixed!

app/controllers/setup_controller.rb
```diff
@@ -246,7 +246,7 @@ def proxy_enabled
   end
 
   def redirect_to_dashboard
-    redirect_to root_path if setup_done?
+    # redirect_to root_path if setup_done?
```
MaximilianMeister: debug?

vitoravelino: Yeah... Reverted. Thanks again! :sweat_smile:

app/controllers/settings/admission_webhook_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @errors = Pillar.apply(
+      admission_webhook_pillars,
+      unprotected_pillars: admission_webhook_pillars.keys
+    )
+    if @form.valid? && @errors.empty?
+      redirect_to settings_admission_webhook_index_path,
+        notice: "Admission webhook settings successfully saved."
+    else
+      set_instance_variables
+      render action: :index, status: :unprocessable_entity
```
MaximilianMeister: what about possible `@errors`? we should show them in an alert, would be a waste to throw them away

app/forms/admission_webhook_form.rb
```diff
@@ -0,0 +1,42 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :enabled, :certificate, :key, :current_cert, :current_key
+
+  validates :key, presence: true, if: :current_key_blank?
+  validates :certificate, presence: true, if: :current_certificate_blank?
+
+  validate :cant_be_empty_file
```
MaximilianMeister: should we at least validate the filetype as well?

```ruby
def must_be_key
  `file #{path}`.include? "private key"
end

def must_be_cert
  `file #{path}`.include? "certificate"
end
```

otherwise i'm afraid we'll have an uncaught exception

MaximilianMeister: as discussed on irc we can address this in a follow up, as we have other cert handling as well, and it should be done for all parts where we upload certs

app/controllers/settings/admission_webhook_controller.rb
vitoravelino: Fixed.

app/forms/admission_webhook_form.rb
bergmannf: I am wondering if parts of that this `virtual` model is doing should maybe be part of the `certificate` class, as that handles storage of uploaded certificates?

Then we might have another class to handle the `keys` - as not all views need both.

vitoravelino: We could `include Certificate`. Would that be enough?

vitoravelino: As far as I can see the only thing that would be reused, and it would in the controller, is the `Certificate.get_certificate_text` method. Validation for this page has a condition and that differs from the `Certificate` model. 

The reason for that is because the user is not forced to update both certificate and key at once. They can update one or the other or both. So I guess it's fine leaving this way, right? 

We can reuse `get_certificate_text` after we merge #593.

bergmannf: Right now you are right - so let's leave it like this.

I assume later on the verification that the certificate is actually a certificate (X509 probably) will be part of certificate.rb, then it might be useful - but I guess we can do that once we work on that feature.

vitoravelino:
I guess we can proceed with the merge, right? I'll give my approval in behalf of @flavio since he created the PR and can't do that.

vitoravelino (dismissed):

vitoravelino:
I updated the code with with the x509 validation and added pem/der key format validation. Suggestion for wording are welcome. I also made a minor refactoring to make things more readable. 

Please review it again. Thanks!

@flavio, could you play with this again?

bergmannf:
Just a short update to fix the merge conflicts. I still intend to setup an admission webhook and actually try out the code.

vitoravelino:
> Just a short update to fix the merge conflicts. I still intend to setup an admission webhook and actually try out the code.

I don't see any conflicts. Is something going to change?

bergmannf:
Sorry - I just meant I repushed the changes after a rebase due to conflicts in ` app/controllers/internal_api/v1/pillars_controller.rb` - the `dex` connector apparently added some more pillars.

app/validators/pem_der_key_validator.rb
```diff
@@ -0,0 +1,10 @@
+require "openssl"
+
+# Verifies that an attribute is a valid PEM/DER encoded key
+class PemDerKeyValidator < ActiveModel::EachValidator
```
bergmannf: This validator seems to fail, when trying to upload a password protected key.

I think it is not supposed to be password-protected as a mutual TLS key, but maybe we should put this in the UI.

vitoravelino:
I've rebased this and should be good to be reviewed again.

spec/forms/admission_webhook_form_spec.rb
```diff
@@ -0,0 +1,105 @@
+require "rails_helper"
+
+describe AdmissionWebhookForm do
+  describe "validations" do
+    let(:cert_file) do
+      Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "admin.crt"))
```
bergmannf: I think this code can just use `fixture_file_upload` instead of `Rack::Test::UploadedFile`.

bergmannf:
LGTM with one small change (use `fixture_file_upload`).

However given that the salt PR is `on hold` I don't think we can merge this right now.

MalloZup:

PR #593, opened 2018-06-26
Allow uploading certificates via file

bergmannf:
Fixes bsc#1097753.

### DIFF
--- a/app/controllers/settings/base_certificate_controller.rb
+++ b/app/controllers/settings/base_certificate_controller.rb
@@ -25,17 +25,18 @@ class Settings::BaseCertificateController < SettingsController
   def new
     @certificate_holder = certificate_holder_type.new
     @cert = Certificate.new
+    @passed_cert = nil
   end
 
   def create
     @certificate_holder = certificate_holder_type.new(
-      certificate_holder_params.except(:certificate)
+      certificate_holder_params.except(:certificate, :current_cert)
     )
-    @cert = Certificate.find_or_initialize_by(certificate: certificate_param)
+    @cert = passed_certificate
 
     ActiveRecord::Base.transaction do
       @certificate_holder.save!
-      create_or_update_certificate! if certificate_param.present?
+      create_or_update_certificate! if passed_certificate.present?
     end
 
     redirect_to [:settings, @certificate_holder],
@@ -49,12 +50,12 @@ class Settings::BaseCertificateController < SettingsController
   end
 
   def update
-    @cert = @certificate_holder.certificate || Certificate.new(certificate: certificate_param)
+    @cert = @certificate_holder.certificate || passed_certificate
 
     ActiveRecord::Base.transaction do
       @certificate_holder.update_attributes!(certificate_holder_update_params)
 
-      if certificate_param.present?
+      if passed_certificate.present?
         create_or_update_certificate!
       elsif @certificate_holder.certificate.present?
         @certificate_holder.certificate.destroy!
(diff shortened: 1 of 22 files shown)


bergmannf:
[(3) M - Bug 1097753 - Certificate should be uploaded as a file](https://trello.com/c/IKaKPmSg/1033-3-m-bug-1097753-certificate-should-be-uploaded-as-a-file)

vitoravelino:
Previously we started allowing the user to copy and paste because it was more convenient but now we are letting the user to select a file. In #590, we are only allowing the file selection and no copy and paste.

To maintain consistency regarding certificates/keys, should we change the workflow here to only file selection or change the #590 to also allow copy and paste from the user?

// cc @flavio @eotchi

flavio:
> To maintain consistency regarding certificates/keys, should we change the workflow here to only file selection or change the #590 to also allow copy and paste from the user?

I agree with that. Let's keep everything consistent, we got feedback from early users, they didn't like the copy&paste thing -> let's adapt and move to file upload only.

vitoravelino:
> I agree with that. Let's keep everything consistent, we got feedback from early users, they didn't like the copy&paste thing -> let's adapt and move to file upload only.

Based on this, could you adapt the PR to be similar to what we did on #590, @bergmannf? There are some screenshots on that PR but lemme know if you still need help to understand them.

vitoravelino:
> agree on consistency. but we should make sure when we have upload only, that we also offer a download, otherwise it might be difficult for the user to see if the cert is the correct one

We have an input text that is disabled/readonly that shows what they've uploaded.

MaximilianMeister:
> We have an input text that is disabled/readonly that shows what they've uploaded.

oh right, that should be enough :+1:

bergmannf:
I reworked this PR: instead of adding more and more fixture files, I now use the generated certificates for the new tests, the way they are used in the `verification` tests.

spec/controllers/settings/registries_controller_spec.rb
```diff
@@ -4,6 +4,13 @@
   let(:user) { create(:user) }
   let(:admin_cert_text) { file_fixture("admin.crt").read.strip }
   let(:pem_cert) { create(:certificate) }
+  let(:pem_cert_text) { pem_cert.certificate.strip }
+  let(:pem_cert_file) do
+    Rack::Test::UploadedFile.new(to_file(pem_cert.certificate), "application/x-x509-user-cert")
```
vitoravelino: You can use `fixture_file_upload` in controller tests. :wink:

spec/rails_helper.rb
```diff
@@ -15,6 +15,7 @@
 Dir[Rails.root.join("spec", "support", "**", "*.rb")].each { |f| require f }
 
 RSpec.configure do |config|
+  config.fixture_path = File.expand_path("../fixtures/", __FILE__)
```
vitoravelino: Duplicated from rebase.

spec/rails_helper.rb
```diff
@@ -15,6 +15,7 @@
...
   # If we want Capybara + DatabaseCleaner + Poltergeist to work correctly, we
   # have to just set this to false.
   config.fixture_path = File.expand_path("../fixtures/", __FILE__)
```
vitoravelino: I guess you should remove this one and keep the other line since the comment doesn't make sense for this.

spec/controllers/settings/registries_controller_spec.rb
bergmannf: Yes, but I create the files on the fly, based on the certificate that is generated by the factory.

So either I have to put those files into the fixture directory (then I can use `fixture_file_upload`) or use this version which just uploads from anywhere.

I could change the `to_file` method to `to_fixture_file` and place the temporary files in the `fixtures` folder instead, which would kind of pollute the folder with a lot of runs though.

app/controllers/settings/base_certificate_controller.rb
```diff
@@ -95,12 +96,30 @@ def certificate_holder_update_params
...
 
+  def passed_certificate
+    # Storing the cert in the variable is required, as uploaded files (in RSpec
+    # tests at least) will not be available for multiple calls:
+    #
+    # > passed_certificate.present? => true
+    #
+    # > passed_certificate.present?  => false
+    if @passed_cert.present?
+      @passed_cert
```
MaximilianMeister: this statement is a bit useless here, as it's already done outside the condition, maybe early return here?

or simply:

```ruby
def passed_certificate
  @passed_cert ||= if certificate_holder_params[:certificate].present? ||
    certificate_holder_params[:current_cert].present?
    return Certificate.find_or_initialize_by(
      certificate: Certificate.get_certificate_text(certificate_holder_params)
    )
  end
end
```

bergmannf: Nice, I didn't know about the `||=` operator in ruby - looks quite handy.

I think I only have to remove the return, otherwise it won't assign anything to `@passed_cert`:

```ruby
[3] pry(main)> a ||= if 0 == 0
[3] pry(main)*   return 1
[3] pry(main)* end
SyntaxError: (eval):4: void value expression
```

MaximilianMeister: oops, yes, the `return` was wrong ;) yeah, the `||=` is often used in ruby to memoize variables

flavio:
What is blocking us from merging this PR on master?

PR #594, opened 2018-06-26
save online status through reconciler (bsc#1098436)

MaximilianMeister:
when a minion is down, it should be shown as down in the UI

miniondownred

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -346,7 +346,7 @@ MinionPoller = {
   },
 
   renderDashboard: function(minion) {
-    var statusHtml;
+    var appliedHtml;
     var checked;
     var masterHtml;
     var actionsHtml;
@@ -354,26 +354,26 @@ MinionPoller = {
 
     switch(minion.highstate) {
       case "not_applied":
-        statusHtml = '<i class="fa fa-circle-o text-success fa-2x" aria-hidden="true"></i>';
+        appliedHtml = '<i class="fa fa-circle-o text-success fa-2x" aria-hidden="true"></i>';
         break;
       case "pending_removal":
       case "pending":
-        statusHtml = '\
+        appliedHtml = '\
           <span class="fa-stack" aria-hidden="true">\
             <i class="fa fa-circle fa-stack-2x text-muted" aria-hidden="true"></i>\
             <i class="fa fa-refresh fa-stack-1x fa-spin fa-inverse" aria-hidden="true"></i>\
           </span>';
         break;
       case "failed":
-        statusHtml = '<i class="fa fa-times-circle text-danger fa-2x" aria-hidden="true"></i>';
+        appliedHtml = '<i class="fa fa-times-circle text-danger fa-2x" aria-hidden="true"></i>';
         MinionPoller.alertFailedBootstrap();
         break;
       case "removal_failed":
         removalFailedClass = 'removal-failed';
-        statusHtml = '<i class="fa fa-minus-circle text-danger fa-2x" aria-hidden="true"></i> Removal Failed';
+        appliedHtml = '<i class="fa fa-minus-circle text-danger fa-2x" aria-hidden="true"></i> Removal Failed';
         break;
       case "applied":
-        statusHtml = '<i class="fa fa-check-circle-o text-success fa-2x" aria-hidden="true"></i>';
+        appliedHtml = '<i class="fa fa-check-circle-o text-success fa-2x" aria-hidden="true"></i>';
         break;
     }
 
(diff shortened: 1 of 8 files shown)


vitoravelino:
I don't think a question mark is the best icon to indicate that a minion is down.

ereslibre:
I was thinking in other ways of doing this, but your solution looks really nice and easy. It's a +1 from me.

ereslibre:
As per @vitoravelino, maybe we can add another (on|off)line indicator in the UI, showing the "status" of the minion as we also show the "applied/applying/not applied" status?

vitoravelino:
> As per @vitoravelino's comment, maybe we can add another (on|off)line indicator in the UI, showing the "status" of the minion as we also show the "applied/applying/not applied" status?

Does it make sense to show "applied and offline" or "not applied and offline"?

I thought about the opaque circle in gray to represent offline. What do you think?

MaximilianMeister:
miniondownextrastatus

I've changed the layout a bit now to have 2 extra columns, probably some tests need to be adapted now ...

> I thought about the opaque circle in gray to represent offline. What do you think?

can you provide the fontawesome identifier for me?

MaximilianMeister:
@ereslibre other question, should we also disable the `remove` link when a node is down, or is that independent?

Martin-Weiss:
It would be cool if the icon / text for for an offline minion that does not respond for a while could indicate something like "last seen ... minutes / hours / days ago" or at least provide the information as "mouse over" after which time a minion is marked as "offline" would be great.

MaximilianMeister:
@Martin-Weiss thanks for the feedback, i've added a tooltip now with `last updated at: timestamp`

MaximilianMeister:
> Cool! Just to be sure to be sure - the timesstamp displayed is the "admin server UTC time" or is it the "browser local time"?

it's data from the back-end (rails database), so should be admin server time

Martin-Weiss:
Would it be possible to add the info („time displayed is admin server time“) to the UI?

MaximilianMeister:
> Would it be possible to add the info („time displayed is admin server time“) to the UI?

added

vitoravelino:
> I've changed the layout a bit now to have 2 extra columns, probably some tests need to be adapted now ...

I'd move the `online` column to the left since there's text regarding the applied column sometimes.

> can you provide the fontawesome identifier for me?

Since you've created another column, I'd go with something like

Online: fa-circle (green)
Offline: fa-circle (gray/red)

Not sure if it's ideal but let's see how it looks.

vitoravelino:
Nodes are marked as offline while they are being bootstrapped: 

screenshot-20180627114426-200x299

I guess they should be online, right?

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -420,7 +420,8 @@ MinionPoller = {
 
     return '\
       <tr> \
-        <td class="status">' + statusHtml +  '</td>\
+        <td class="online">' + onlineHtml(minion) +  '</td>\
```
vitoravelino: Change class to `status`. This will align the icon more to the right.

MaximilianMeister:
> I guess they should be online, right?

right

hmm, i **think** this is due to the default `online: false` - we could either set it to `true` or we need to run `Minion.update_grains` before bootstrap

setting it by default to `true` would make sense i guess, because the moment when the minion is created (in the db), we can assume at this point it is online

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -1062,3 +1063,11 @@ function requestNodeForceRemoval(minionId) {
     .fail(enableOrchTriggers)
     .fail(notifyRemovalError);
 }
+
+function onlineHtml(minion) {
+  if(minion.online) {
+    return '<i class="fa fa-circle text-success fa-2x" aria-hidden="true"></i>';
```
vitoravelino: Title attribute with `Online` would be nice.

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -1062,3 +1063,11 @@ function requestNodeForceRemoval(minionId) {
...
+  } else {
+    return '<i class="fa fa-circle text-danger fa-2x" aria-hidden="true" title="last updated at: ' + minion.updated_at + ' (admin server time)"></i>';
```
vitoravelino: Minor improvement adding `Offline` to the title like `Offline (last update at: ...)`.

MaximilianMeister: added

app/assets/javascripts/dashboard/dashboard.js
MaximilianMeister: good idea! added

MaximilianMeister:
we have the same issue here as explained in https://github.com/kubic-project/velum/pull/585#discussion_r198378403 so jenkins will likely not be seen green before merging this

flavio:
I know I'm late, but what about having a single column "status" that shows:
* green checkmark (online + updated), 
* something spinning (online + upgrading)
* red: online, update failed
* another color (orange?), whole row "disabled": offline, unknown state

PR #599, opened 2018-06-28
ui: removed fixed width for main header logo

vitoravelino:
Logo can be both CaaSP or Kubic now and the fixed width was causing a
small misalignment with Kubic logo. Turns out that this fixed width is
not necessary at all.

fix#kubic-logo

screenshot-20180628065256-289x144
screenshot-20180628065323-297x320


vitoravelino:
[kubic logo too big - only on login screen](https://trello.com/c/RXAEYDuf/1020-kubic-logo-too-big-only-on-login-screen)

PR #600, opened 2018-06-28
Certificate validation

bergmannf:
Fixes bsc#1097754: a certificate's text will be verified to be a `X509` certificate.

### DIFF
--- a/app/models/certificate.rb
+++ b/app/models/certificate.rb
@@ -2,5 +2,5 @@
 class Certificate < ActiveRecord::Base
   has_many :certificate_services, dependent: :destroy
 
-  validates :certificate, presence: true
+  validates :certificate, presence: true, x509_certificate: true
 end
--- a/spec/models/certificate_spec.rb
+++ b/spec/models/certificate_spec.rb
@@ -1,6 +1,25 @@
 require "rails_helper"
 
-describe Certificate do
+RSpec.describe Certificate do
   it { is_expected.to have_many(:certificate_services) }
   it { is_expected.to validate_presence_of(:certificate) }
+
+  context "when a certificate was passed" do
+    it "accepts a PEM formatted certificate" do
+      x509_cert = OpenSSL::X509::Certificate.new(create(:certificate).certificate)
+      cert = described_class.new(certificate: x509_cert.to_pem)
+      expect(cert.valid?).to eq(true)
+    end
+
+    it "accepts a PER formatted certificate" do
+      x509_cert = OpenSSL::X509::Certificate.new(create(:certificate).certificate)
+      cert = described_class.new(certificate: x509_cert.to_der)
+      expect(cert.valid?).to eq(true)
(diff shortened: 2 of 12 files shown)


bergmannf:
[(3) M - Bug 1097754 - Certificates are not validated](https://trello.com/c/YnZg22NC/1034-3-m-bug-1097754-certificates-are-not-validated)

app/models/certificate.rb
```diff
@@ -1,6 +1,15 @@
+require "openssl"
+
 # Certificate store
 class Certificate < ActiveRecord::Base
   has_many :certificate_services, dependent: :destroy
 
   validates :certificate, presence: true
+  validate :valid_certificate_format
```
vitoravelino: Not sure if it's overengineering but could we extract this to something similar to the `EmailValidator` example at http://guides.rubyonrails.org/active_record_validations.html#custom-validators?

This way I could reuse elsewhere by just `validates :field, certificate: true`. Otherwise, how would I reuse this kind of validation?

bergmannf: @vitoravelino I assumed that every certificate would go through the `certificate.rb` class in the end.

But I think in general having the `validator` separate is better design - as we might need different ones for `PKCS` certificates for some views.

vitoravelino: > @vitoravelino I assumed that every certificate would go through the certificate.rb class in the end.

I thought you had assumed that, so it's fine. The problem is that for simpler features we are storing directly though Pillar instead of its own model. 

> But I think in general having the validator separate is better design - as we might need different ones for PKCS certificates for some views.

That's also true.

vitoravelino:
> rest LGTM, however we should also check if we still need the old cert fixtures in https://github.com/kubic-project/velum/tree/master/spec/fixtures and if not, drop them 

I've used the the `admin.key` and `admin.crt` in #590 since they were already there.

spec/models/certificate_spec.rb
```diff
@@ -1,6 +1,25 @@
...
+
+  context "when a certificate was passed" do
+    it "accepts a PEM formatted certificate" do
+      pem_cert = file_fixture("admin.crt").read
+      cert = described_class.new(certificate: pem_cert)
+      expect(cert.valid?).to eq(true)
+    end
+
+    it "accepts a PER formatted certificate" do
+      per_cert = file_fixture("cert.der").binread
```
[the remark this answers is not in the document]
bergmannf: It seems we can generate certificates using the `openssl` library:

http://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL.html#module-OpenSSL-label-X509+Certificates

I will switch to that for the `binary` format.

bergmannf:
In some of the tests we need two different certificates - that's why I added another set.

Maybe it would be better to make the factory classes generate more than 1 certificate instead and then just use those instead.

PR #602, opened 2018-07-03
[CPI] Add option to ignore OpenStack Cinder availability zone, bsc#1095572

mjura:
Ignore OpenStack Cinder avability zone when attaching volumes.
When Nova and Cinder have different availability zones,
this should be set to true. Default is false.

### DIFF
--- a/app/controllers/internal_api/v1/pillars_controller.rb
+++ b/app/controllers/internal_api/v1/pillars_controller.rb
@@ -141,7 +141,8 @@ class InternalApi::V1::PillarsController < InternalApiController
           floating:       Pillar.value(pillar: :cloud_openstack_floating),
           subnet:         Pillar.value(pillar: :cloud_openstack_subnet),
           bs_version:     Pillar.value(pillar: :cloud_openstack_bs_version),
-          lb_mon_retries: Pillar.value(pillar: :cloud_openstack_lb_mon_retries)
+          lb_mon_retries: Pillar.value(pillar: :cloud_openstack_lb_mon_retries),
+          ignore_vol_az:  Pillar.value(pillar: :cloud_openstack_ignore_vol_az)
         }
       }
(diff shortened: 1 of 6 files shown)


mjura:
[Bug 1095572 - [ CPI ] Scheduling pods with automatic PersistentVolumenClaim doesn't work](https://trello.com/c/GjzL1LwJ/1061-bug-1095572-cpi-scheduling-pods-with-automatic-persistentvolumenclaim-doesnt-work)

app/controllers/internal_api/v1/pillars_controller.rb
```diff
@@ -142,6 +142,7 @@ def openstack_cloud_contents
           subnet:         Pillar.value(pillar: :cloud_openstack_subnet),
           bs_version:     Pillar.value(pillar: :cloud_openstack_bs_version),
           lb_mon_retries: Pillar.value(pillar: :cloud_openstack_lb_mon_retries)
```
vitoravelino: Missing a comma and the end of the line.

mjura: Fixed

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -309,7 +309,8 @@
             floating:       "9bc3e819-a6ca-648b-b5e3-c26c9e6c5e57",
             subnet:         "4b64b38d-0b38-40d0-a69f-ade7299ef4ab",
             bs_version:     "v2",
-            lb_mon_retries: "3"
+            lb_mon_retries: "3",
+            ignore_vol_az:  "false"
```
kiall: Is there a reason for this to be a string instead of a boolean? The JSON is going to be a little unnatural as a string

PR #604, opened 2018-07-04
ui: renamed new nodes -> unassigned nodes

vitoravelino:
On the cluster status summary we decided to rename "new" by "unassigned"
to avoid mixing terms that may confuse the user.

bsc#1100113

screenshot-20180704134245-229x191


vitoravelino:
As soon as I get an approval I'll backport it to 3.0.

PR #605, opened 2018-07-05
[3.0] ui: renamed new nodes -> unassigned nodes

vitoravelino:
On the cluster status summary we decided to rename "new" by "unassigned"
to avoid mixing terms that may confuse the user.

bsc#1100113

(cherry picked from commit b18c46e19129e8c43d5e3abc87e355beae32d562)

Backport of #604


r0ckarong:
Could you uppercase the text in the brackets after the count to

so

(New)
instead of
(new)

Looks more polished

PR #614, opened 2018-07-31
Update the minion hostname in a regular basis, based on the grain information.

ereslibre:
Fixes: bsc#1103307

### DIFF
--- a/app/models/minion.rb
+++ b/app/models/minion.rb
@@ -25,7 +25,8 @@ class Minion < ApplicationRecord
 
         tx_update_reboot_needed = minion_grains["tx_update_reboot_needed"] || false
         tx_update_failed = minion_grains["tx_update_failed"] || false
-        minion.update_columns online:                  online,
+        minion.update_columns fqdn:                    minion_grains["fqdn"],
+                              online:                  online,
                               tx_update_reboot_needed: tx_update_reboot_needed,
                               tx_update_failed:        tx_update_failed
       rescue StandardError


spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -310,6 +310,9 @@
       {
         registries:          [],
         system_certificates: [],
+        dex:                 {
+          connectors: []
+        },
```
flavio: I think this is unrelated with this PR, isn't it?

app/models/minion.rb
```diff
@@ -25,7 +25,8 @@ def self.update_grains
 
         tx_update_reboot_needed = minion_grains["tx_update_reboot_needed"] || false
         tx_update_failed = minion_grains["tx_update_failed"] || false
-        minion.update_columns online:                  online,
+        minion.update_columns fqdn:                    minion_grains["fqdn"],
+                              online:                  online,
```
flavio: Is the reconciliation loop going to keep this grain in sync with the database?

spec/controllers/internal_api/v1/pillars_controller_spec.rb
vitoravelino: It will disappear after a rebase. An old PR was merged without rebasing and introduced this small issue that was already fixed.

app/models/minion.rb
MaximilianMeister: @flavio yes it does. it [calls exactly this method here](https://github.com/kubic-project/velum/blob/master/lib/tasks/salt.rake#L18)

spec/controllers/internal_api/v1/pillars_controller_spec.rb
mjura: lol

flavio:
It will LGTM to me after the rebase is done

spec/controllers/internal_api/v1/pillars_controller_spec.rb
ereslibre: Yes, this was in purpose because travis was red because of this. Rebasing now.

ereslibre:
Rebased.

PR #616, opened 2018-07-31
Added LDAP Connector validation button and logic

danielorf:
Added new controller "ldap_test_controller".
Updated _form.html.slim to add test button and message.
Updated application.js to add jquery logic to test button in place and
modify message.
Updated routes.rb for new controller.
Added RSpec tests.

Other notes:
RSpec tests require a working LDAP server.  Additionally, the LDAP server must allow for anonymous binds, support TLS and be populated with at least one user having "objectClass=person".  The certificate in the tests is pulled from the dockerized LDAP server used during development/testing and can be found here:  https://github.com/osixia/docker-openldap

There are coverage guards/exceptions around two "rescue StandardError" code blocks in ldap_test_controller.rb.  These blocks are fail-safes in case the error-handling blocks above them fail to catch an error.  Fail-safes are deemed necessary in both cases because the "begin" blocks run "net/ldap" code which is light on documentation regarding errors.  We are not currently aware of a state that will enter these "rescue StandardError" code blocks, thus they are skipped in testing.


app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+  $("#ldap_conn_test").click(function (event) {
+
+    // Get values from form 
+    var host = $('#dex_connector_ldap_host').val();
+    var port = $('#dex_connector_ldap_port').val();
+    var start_tls = $("#dex_connector_ldap_start_tls_true").parent().hasClass('active');
+    var cert = $('#dex_connector_ldap_certificate').val();
+    // Flip logic of #anonymous_bind_toggle since expanded div means anonymous binding is 'false'
+    var anon_bind = ($('#anonymous_bind_toggle').attr('aria-expanded') == "false");
+    var dN = $('#dex_connector_ldap_bind_dn').val();
```
nanoscopic: Is there a reason the N here in dN is capital?

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+    var pass = $('#dex_connector_ldap_bind_pw').val();
+    var baseDN = $('#dex_connector_ldap_user_base_dn').val();
```
nanoscopic: In order to match the style of the other variables, consider using 'base_dn' as the variable name here.

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+    var filter = $('#dex_connector_ldap_user_filter').val();
+
+    base_url = "/settings/ldap_test";
```
nanoscopic: Where is base_url defined? I'm not seeing it here in this file.

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+
+    // Check if form inputs are filled in
+    if (host.length && port.length && cert.length && dN.length && pass.length &&
+      baseDN.length && filter.length) {
+
+      $.post(base_url, {
+        host: host, port: port, start_tls: start_tls,
```
nanoscopic: I would recommend putting each item on a line of its own to match preferred styling and make it easier to scan all the variables being posted.

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+        cert: cert, anon_bind: anon_bind, dN: dN, pass: pass,
+        baseDN: baseDN, filter: filter
+      }).done(function (data) {
+
+        if (data.result.status == "pass") {
```
nanoscopic: For true false pass/false it would make sense to use a boolean true/false here. Instead of status, you could use "passed" or something similar.

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+          val_message = "Sucessfully validated connection to LDAP server";
```
nanoscopic: val_message should be defined somewhere. I am assuming it is? Could you comment where?

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+          $("#ldap_conn_save").prop("disabled", false);
+          $("#ldap_conn_message")[0].style.color = "#04a36f";
+          console.log(data);
```
nanoscopic: console.log should not generally be present in production code

app/assets/javascripts/application.js
```diff
@@ -22,6 +22,65 @@
...
+        }
+        else {
+          val_message = "Validation failure. " + data.result.message;
+          $("#ldap_conn_save").prop("disabled", true);
```
nanoscopic: I've generally seen "disabled" set to "disabled" not true. I believe this is necessary for some specific browsers.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+  def ldap_filter_search(ldap, baseDN, filter_string)
+    filter = Net::LDAP::Filter.construct(filter_string)
+    attrs = ["mail", "cn", "sn", "objectclass"]
+    search_object = ldap.search(
+      base:          baseDN,
+      filter:        filter,
+      attributes:    attrs,
+      return_result: true
+    )
+    search_object
```
nanoscopic: I believe this function could just end in ldap.search, and the return of it would automatically be returned by the function. As is it does make it more clear that it is intended to return search_object though, so it may be better to stay this way.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def render_result(result_hash, status, message)
+    result_hash[:result] = { status: status, message: message }
+    render json: result_hash
```
nanoscopic: This needlessly returns the value of whatever the render function returns. Intended? If not the function should end in "return"

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  # Returns an ldap object
+  def get_ldap(host, port, start_tls, pem_string, anon_bind, dN, pass)
+    tls = if start_tls == "true"
```
nanoscopic: Recommend using the ternary operator here. It is more confusing the way it is written here.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+    else
+      :simple_tls
+    end
+
+    decoded_pem = Base64.decode64(pem_string)
+    cert = OpenSSL::X509::Certificate.new(decoded_pem)
+    cert_store = OpenSSL::X509::Store.new
+    cert_store.add_cert(cert)
+    tls_options = {
+      verify_mode: OpenSSL::SSL::VERIFY_PEER,
```
nanoscopic: A comment explaining what this flag does would be helpful. "This flag causes OpenSSL to check the validity of the server certificate."

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+      cert_store:  cert_store
+    }
+
+    auth_hash = if anon_bind == "true"
```
nanoscopic: Not a fan of this method of using if "a = b == c   if   x  else y  end" just looks very odd. Perhaps parentheses around the whole if block would help make it more clear what is happening? Would that work?

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+      {
+        method: :anonymous
```
nanoscopic: Not a fan of using colon strings for values. For keys yes, but for values quoted strings look better.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+        method:   :simple,
+        username: dN,
+        password: pass
+      }
+    end
+    Net::LDAP.new host:       host,
+                  port:       port,
+                  auth:       auth_hash,
+                  encryption: {
+                    method:      tls, # could also be :start_tls
```
nanoscopic: Comment here no longer makes sense since it has been changed to be a variable. Also a better variable name would be 'encryption_method'

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+                    tls_options: tls_options
+                  }
+  end
+
+  def create
+    result_hash = { initial_bind: {}, user_search: {}, result: {} }
+    result_hash.store("params", params)
```
nanoscopic: You can just set the params within the result_hash definition on the previous line. It's odd to split it up.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+    # Incoming params
+    host = params[:host]
+    port = params[:port]
+    start_tls = params[:start_tls]
+    pem_string = params[:cert]
+    anon_bind = params[:anon_bind]
+    dN = params[:dN]
+    pass = params[:pass]
+    baseDN = params[:baseDN]
+    filter_string = params[:filter]
```
nanoscopic: Lining up the equals lines on all of these would be nice. I thought this is required by rubocop?

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+      ldap = get_ldap(host, port, start_tls, pem_string, anon_bind, dN, pass)
+    rescue OpenSSL::X509::CertificateError
+      result_hash.store("error", "Invalid Cerificate Format")
+      render_result(result_hash, "fail", "Invalid certificate, check format and try again")
+      return
+    # :nocov:
+    rescue StandardError => error
+      # Using StandardError because each OpenSSL::X509 errors must be caught individually
+      # and only CertificateError is documented: https://ruby-doc.org/stdlib-2.5.0/libdoc/openssl/rdoc/OpenSSL/X509/CertificateError.html
+      result_hash.store("error", error.inspect)
```
nanoscopic: Adding error.inspect into output may leak server information. This specific addition looks to be just for debugging and shouldn't go into production.

app/controllers/settings/ldap_test_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+    rescue Net::LDAP::Error => error
+      result_hash.store("error", error.inspect)
+      render_result(result_hash, "fail", "Check that host/port are correct and reachable.")
+      return
+    end
+
+    # Search for user
+    begin
+      user_search_records = ldap_filter_search(ldap, baseDN, filter_string)
+      if !user_search_records.empty?
```
nanoscopic: Consider reversing the if so that the '!' can be dropped

spec/controllers/settings/ldap_test_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    before do
+      setup_done
+      sign_in user
+    end
+
+    describe "POST #create" do
+        let(:host) { "ldaptest.com" }
+        let(:port) { 636 }
+        let(:start_tls) { "false" }
+        let(:cert) { "MIIC0zCCAlmgAwIBAgIUCfQ+m0pgZ/BjYAJvxrn/bdGNZokwCgYIKoZIzj0EAwMw
```
nanoscopic: It may make sense to use TLS library directly to fetch the cert from the host at the very start, so that whatever cert ldaptest.com provides it still functions.

spec/controllers/settings/ldap_test_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+            post :create, host: host, port: port, start_tls: start_tls, cert: cert, 
+            anon_bind: anon_bind, dN: dN, pass: pass, baseDN: baseDN, filter: filter
+            expect(response).to have_http_status(:ok)
+        end
+
+        it "has the expected result status" do
+            post :create, host: host, port: port, start_tls: start_tls, cert: cert, 
+            anon_bind: anon_bind, dN: dN, pass: pass, baseDN: baseDN, filter: filter
+            parse_json = JSON(response.body)
+            expect(parse_json["result"]["status"]).to eq("pass")
```
nanoscopic: As stated before, a boolean 'passed' would be sensible for JSON return.
Also it may be reasonable to use http status code to indicate a pass or fail.

app/assets/javascripts/application.js
danielorf: I was attempting to keep it as the capitalized LDAP convention, 'DN'.  I settled with 'dN' because Ruby automatically sets capitalized variables as constants.  There is no reason to not follow ruby convention and change to 'dn' though.  So I am changing it to 'dn'.

app/assets/javascripts/application.js
danielorf: Incorporated now.

app/assets/javascripts/application.js
danielorf: Forgot 'var'.  Incorporated now.

app/assets/javascripts/application.js
danielorf: Incorporated now.

app/controllers/settings/ldap_test_controller.rb
danielorf: Adding a return causes a rubocop "Style/RedundantReturn" error.

app/controllers/settings/ldap_test_controller.rb
danielorf: colon strings appear to be a common convention in ruby.

app/assets/javascripts/application.js
danielorf: Fixed now.

app/assets/javascripts/application.js
danielorf: Fixed now.

app/assets/javascripts/application.js
danielorf: Fixed now.

app/assets/javascripts/application.js
danielorf: Fixed now.

app/controllers/settings/ldap_test_controller.rb
danielorf: Fixed now.

app/controllers/settings/ldap_test_controller.rb
danielorf: This was a rubocop recommendation.

app/controllers/settings/ldap_test_controller.rb
danielorf: Incorporated now.

app/controllers/settings/ldap_test_controller.rb
danielorf: Fixed now

app/controllers/settings/ldap_test_controller.rb
danielorf: Line 64 ("result_hash.store("params", params)") was used for debug and is vestigial.  Removed it.

app/controllers/settings/ldap_test_controller.rb
danielorf: Not required by rubocop, but incorporated now.

app/controllers/settings/ldap_test_controller.rb
danielorf: Removed it.

app/controllers/settings/ldap_test_controller.rb
danielorf: This format was chosen as personal convention to to keep the passing condition at top and failing condition at bottom.

spec/controllers/settings/ldap_test_controller_spec.rb
danielorf: Added code to the RSpec test (ldap_test_controller_spec.rb) which gets the necessary cert from the LDAP server.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,68 @@
...
+          cert: cert,
+          anon_bind: anonBind,
+          dn: dn,
+          pass: pass,
+          base_dn: baseDN,
+          filter: filter
+        }).done(function (data) {
+          if (data.result.test_pass) {
+            valMessage = 'Sucessfully validated connection to LDAP server'
+            $('#ldap_conn_save').prop('disabled', null)
```
vitoravelino: To remove a prop you can use `removeProp` method.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,68 @@
...
+            $('#ldap_conn_message')[0].style.color = '#04a36f'
```
vitoravelino: Instead of manipulating the color directly in the element, it would be nicer to use classes to map the state of the element. You could add a class like `valid` or `#ldap_conn_message--valid` and add it here. Same applies for the failure/invalid message below.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,68 @@
...
+            $('#ldap_conn_save').prop('disabled', 'disabled')
+            $('#ldap_conn_message')[0].style.color = 'red'
+          }
+          $('#ldap_conn_message').html(valMessage) // Redundant because of async calls
+        })
+      } else {
+        valMessage = 'Missing form data'
+        $('#ldap_conn_save').prop('disabled', 'disabled')
+        document.getElementById('ldap_conn_message').style.color = 'red'
+        $('#ldap_conn_message').html(valMessage) // Redundant because of async calls
```
vitoravelino: You can extract this to a method and call it whenever necessary.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,68 @@
...
+            valMessage = 'Sucessfully validated connection to LDAP server'
+            $('#ldap_conn_save').prop('disabled', null)
+            $('#ldap_conn_message')[0].style.color = '#04a36f'
+          } else {
+            valMessage = 'Validation failure. ' + data.result.message
+            $('#ldap_conn_save').prop('disabled', 'disabled')
+            $('#ldap_conn_message')[0].style.color = 'red'
+          }
+          $('#ldap_conn_message').html(valMessage) // Redundant because of async calls
+        })
```
vitoravelino: What if the request fails? I think we should handle the `fail` callback just to avoid the "nothing happens when I click on..." issue.

danielorf: Fixed now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Fixed now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Fixed now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Fixed now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Added classes:  ldap-conn-message-default, ldap-conn-message-invalid, ldap-conn-message-valid

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Added method RenderValidity to consolidate this code.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,78 @@
+$(function () {
+  function NoneEmpty (arr) {
+    for (var i = 0; i < arr.length; i++) {
+      if (arr[i] === '') return false
+    }
+    return true
+  };
+
+  function RenderValidity (msg, isValid) {
```
vitoravelino: In javascript we usually write names in lower camel case.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,78 @@
...
+    if (isValid) {
```
vitoravelino: Before this, it's a good idea to remove both classes to make sure we don't mix them (first it was invalid and then valid) to avoid an unwanted behavior. There's a jQuery method called `removeClass` that does that separating classes by space.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,78 @@
...
+      $('#ldap_conn_save').removeProp('disabled')
+      $('#ldap_conn_message').attr('class', 'ldap-conn-message-valid')
```
vitoravelino: Instead of `attr`, prefer `addClass`.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,78 @@
...
+          cert: cert,
+          anon_bind: anonBind,
+          dn: dn,
+          pass: pass,
+          base_dn: baseDN,
+          filter: filter
+        }).done(function (data) {
+          if (data.result.test_pass) {
+            valMessage = 'Sucessfully validated connection to LDAP server'
+            RenderValidity(valMessage, true)
```
vitoravelino: Besides the `fail` case, you can pass directly the string as an argument instead of assigning it to a variable first.

danielorf: Changed on all but the 'fail' case.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: 'attr' was initially chosen due to its conciseness - no need to remove and add classes as it overwrites the current class.  In hindsight, it's obvious that it it would prevent the use of additional classes as they would get replaced.  Fixed using your suggestion now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Fixed now.  See comment below regarding 'attr'.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Yep, my mistake.  Fixed now.

app/assets/stylesheets/pages/settings.scss
```diff
@@ -92,4 +92,16 @@ $sidebar-item-highlight: #00c081;
...
+.ldap-conn-message-default {
+  color: black;
+}
+
+.ldap-conn-message-invalid {
+  color: red;
+}
+
+.ldap-conn-message-valid {
+  color: #04a36f;
```
vitoravelino: You can replace the hardcoded color with predefined vars. In this case `$brand-success`.

app/assets/stylesheets/pages/settings.scss
```diff
@@ -92,4 +92,16 @@ $sidebar-item-highlight: #00c081;
...
-}
\ No newline at end of file
+}
+
+.ldap-conn-message-default {
+  color: black;
+}
+
+.ldap-conn-message-invalid {
+  color: red;
```
vitoravelino: For this `$brand-danger`

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -0,0 +1,76 @@
...
+      $('#ldap_conn_save').removeProp('disabled')
+      $('#ldap_conn_message').addClass('ldap-conn-message-valid')
+    } else {
+      $('#ldap_conn_save').prop('disabled', 'disabled')
+      $('#ldap_conn_message').addClass('ldap-conn-message-invalid')
+    }
+    $('#ldap_conn_message').html(msg)
+  }
+
+  $(document).ready(function () {
```
vitoravelino: This is already inside of a `ready` callback, `$(function() { ...` it's a shortcut for this.

app/assets/stylesheets/pages/settings.scss
```diff
@@ -92,4 +92,16 @@ $sidebar-item-highlight: #00c081;
     margin-top: 0;
     margin-bottom: 15px;
   }
-}
\ No newline at end of file
+}
+
+.ldap-conn-message-default {
```
vitoravelino: I guess we don't need this since it will inherit from body and no other class will override it. Remember to remove the js references for this.

vitoravelino:
Besides my latest comments, LGTM. 

Another thing, travis is failing because of coverage.

app/assets/stylesheets/pages/settings.scss
danielorf: Fixed now.

app/assets/stylesheets/pages/settings.scss
danielorf: Fixed now.

app/assets/javascripts/settings/ldap-conn-test.js
danielorf: Fixed now.

app/assets/stylesheets/pages/settings.scss
danielorf: #ldap_conn_message inherits from h4 which is set to $brand-success color.  I went with h4 on this message for the sake of consistency with the rest of the page.  Thus I kept ldap-conn-message-default style in place so that the message starts out black instead of green ($brand-success).

vitoravelino:
Could you rebase it once again?

vitoravelino:
It would be helpful for docs if you could add some screenshots.

PR #618, opened 2018-08-02
[3.0] [CPI] Add option to ignore OpenStack Cinder availability zone, bsc#1095572

mjura:


dannysauer:
This was merged into master (with review from @vitoravelino & @MaximilianMeister) back in July, and was previously approved here before Kiall left.  What happened to prevent merge into the release?

PR #620, opened 2018-08-06
[CPI] Split logic for Project and Domain ID, bsc#1103873

mjura:


mjura:
I've tested it and works fine as we expected

PR #629, opened 2018-08-18
PublicCloud#remove: Not yet implemented

bear454:
Removing nodes is not implemented in public clouds (Azure, EC2, GCE) at this time, so the interfaces for doing so should be disabled until the feature is implemented.


bear454:
When removal of nodes is implemented for public cloud, this can be reverted.

PR #631, opened 2018-08-20
[3.0] PublicCloud#remove: Not yet implemented

bear454:
Removing nodes is not implemented in public clouds (Azure, EC2, GCE) at this
time, so the interfaces for doing so should be disabled until the feature
is implemented.

(cherry picked from commit 377e4fe09e3ec239cfcac284e4dfef17018f95a7)

Backport of #629

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -391,10 +391,15 @@ MinionPoller = {
       appliedHtml = '<i class="fa fa-arrow-circle-up text-danger fa-2x" aria-hidden="true"></i> Update Failed';
     }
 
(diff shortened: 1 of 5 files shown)


bear454:
vitoravelino reviewed 10 minutes ago

> Are we backporting without bugzilla ref?

I don't know how this is being handled at the PM level, but please be aware that for public cloud, it's not a backport; 3.0 has not been released there yet.

vitoravelino:
> I don't know how this is being handled at the PM level, but please be aware that for public cloud, it's not a backport; 3.0 has not been released there yet.

In this case I guess you could use something like `feature#cloud-node-removal` just to help tracking stuff.

PR #632, opened 2018-08-21
[followup #630] require uglifier

MaximilianMeister:
the constant `Uglifier` is not accessible when opening a
rails console or running a rake command

fix#uglifier

## Follow up of

https://github.com/kubic-project/velum/pull/630

@jordimassaguerpla a more elegant **nitpicky** solution would probably be to only configure `Uglifier` when we precompile the assets, as during runtime we don't really need it. But since we dont have the `PACKAGING` env anymore, we don't have an easy way to determine if we're about to precompile or not. it would look like this:

```ruby
...
  if ENV["PACKAGING"]
    config.assets.js_compressor = Uglifier.new(harmony: true)
  end
...
```

On the other hand it might not be an issue to load `Uglifier` during the rails console session, as during runtime it seems to be loaded and accessible already

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -19,7 +19,9 @@ Rails.application.configure do
   config.serve_static_files = ENV["RAILS_SERVE_STATIC_FILES"].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
+  if ENV["RPM_BUILD_ROOT"]
+    config.assets.js_compressor = Uglifier.new(harmony: true)
+  end
   # config.assets.css_compressor = :sass
 
   # include svg in the precompile


MaximilianMeister:
> View changes 
> Do we require uglifier in production? I thought this was needed when compiling the assets.

It's only used to compile/compress the assets. but nevertheless this file is parsed, and (at least) the constant `Uglifier` needs to be loaded

jordimassaguerpla:
uglifier is in the assets group. IIRC this group won't be installed on a production environment, thus requiring uglifier will break the RPM. Am I missing something?

MaximilianMeister:
> uglifier is in the assets group. IIRC this group won't be installed on a production environment, thus requiring uglifier will break the RPM. Am I missing something?

@jordimassaguerpla that makes total sense, maybe it's a devsetup issue. when i run `docker exec -it $(docker ps | grep "velum-dashboard" | awk '{print $1}') entrypoint.sh bash -c "export RAILS_ENV=production; bundle exec rails c"` I get this error. and look at the RAILS_ENV here ... :thinking:

config/environments/production.rb
```diff
@@ -19,7 +19,9 @@
   config.serve_static_files = ENV["RAILS_SERVE_STATIC_FILES"].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
+  if ENV["RPM_BUILD_ROOT"]
+    config.assets.js_compressor = Uglifier.new(harmony: true)
+  end
```
MaximilianMeister: @jordimassaguerpla what about this workaround ? this should ensure that it's only called during build time

jordimassaguerpla: If that works, lgtm

PR #627, opened 2018-08-13
bsc#1098085 Add nodes in public clouds

bear454:
* Requires #595
* Requires #610

Adapt public cloud bootstrap views & controller actions to enable adding nodes after boostrap.

PR #633, opened 2018-08-21
prevent accessing discovery after bootstrap

MaximilianMeister:
PR #627 introduced a weird behaviour of skip_before_action
when 'only' and 'if' is used at the same time

see:
https://github.com/rails/rails/issues/9703

we can just omit the 'if setup_done?' check because that's
already implicitly happening through the application_controller


MaximilianMeister:
to reproduce the bug, bootstrap a cluster and access `/setup/discovery`

PR #634, opened 2018-08-21
Updates to prevent auto complete from filling out LDAP fields

Klaven:
A extra text field was added to break up the form flow to keep firefox from auto filling in the user name and password. this extra field was then hidden by css. if hidden in the HTML firefox would know and still auto fill.

### DIFF
--- a/app/views/settings/dex_connector_ldaps/_form.html.slim
+++ b/app/views/settings/dex_connector_ldaps/_form.html.slim
@@ -53,14 +53,17 @@
     = error_messages_for(@certificate_holder, :bind_anon)
 
   .ldap-dn-authentication.collapse class="#{'in' if !@certificate_holder.bind_anon}" id="anonymous_bind_toggle" aria-expanded="true"
-    .form-group class="#{error_class_for(@certificate_holder, :bind_dn)}"
+    .form-group class="#{error_class_for(@certificate_holder, :bind_dn)}" autocomplete="nope"
       = f.label :bind_dn, "DN"
       = f.text_field :bind_dn, class: "form-control", value: @certificate_holder.bind_dn, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
       = error_messages_for(@certificate_holder, :bind_dn)
 
-    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}"
+
+    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}" autocomplete="nope"
       = f.label :bind_pw, "Password"
-      = f.password_field :bind_pw, class: "form-control", value: @certificate_holder.bind_pw, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
+        / This is done to keep firefox from auto filling saved creditials.
+      = f.text_field :bind_pw, class: "form_control settings-hidden", value:"none" 
+      = f.password_field :bind_pw, class: "form-control", value: @certificate_holder.bind_pw, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon, autocomplete: "new-password"
       = error_messages_for(@certificate_holder, :bind_pw)
 
   hr
--- a/app/assets/stylesheets/pages/settings.scss
+++ b/app/assets/stylesheets/pages/settings.scss
@@ -13,6 +13,11 @@ $sidebar-item-highlight: #00c081;
   z-index: 10;
 }
 
+//used to make elements hidden
+.settings-hidden {
+  display: none
+}
+
 .settings-sidebar-section {
   .title {
     color: $sidebar-section-title;


app/views/settings/dex_connector_ldaps/_form.html.slim
```diff
@@ -53,14 +53,17 @@
     = error_messages_for(@certificate_holder, :bind_anon)
 
   .ldap-dn-authentication.collapse class="#{'in' if !@certificate_holder.bind_anon}" id="anonymous_bind_toggle" aria-expanded="true"
-    .form-group class="#{error_class_for(@certificate_holder, :bind_dn)}"
+    .form-group class="#{error_class_for(@certificate_holder, :bind_dn)}" autocomplete="nope"
```
vitoravelino: I'd move this to the text field below instead of the form group.

vitoravelino: There's no advantage on putting this in the `.form-group` div element. I suggest you to move it directly to the input text field element.

app/views/settings/dex_connector_ldaps/_form.html.slim
```diff
@@ -53,14 +53,17 @@
...
       = f.text_field :bind_dn, class: "form-control", value: @certificate_holder.bind_dn, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
       = error_messages_for(@certificate_holder, :bind_dn)
 
-    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}"
+
+    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}" autocomplete="nope"
       = f.label :bind_pw, "Password"
-      = f.password_field :bind_pw, class: "form-control", value: @certificate_holder.bind_pw, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
+        / This is done to keep firefox from auto filling saved creditials.
+      = f.text_field :bind_pw, class: "form_control settings-hidden", value:"none"
```
vitoravelino: Is this really necessary?

app/views/settings/dex_connector_ldaps/_form.html.slim
vitoravelino: I also read about adding an invalid option to enforce non-autocompletion but a simply `off` would do the job, no? I'm not aware of the unexpected behavior scenarios. Did you face any?

app/views/settings/dex_connector_ldaps/_form.html.slim
```diff
@@ -53,14 +53,17 @@
...
       = error_messages_for(@certificate_holder, :bind_dn)
 
-    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}"
+
+    .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}" autocomplete="nope"
       = f.label :bind_pw, "Password"
-      = f.password_field :bind_pw, class: "form-control", value: @certificate_holder.bind_pw, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
+        / This is done to keep firefox from auto filling saved creditials.
+      = f.text_field :bind_pw, class: "form_control settings-hidden", value:"none" 
+      = f.password_field :bind_pw, class: "form-control", value: @certificate_holder.bind_pw, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon, autocomplete: "new-password"
```
vitoravelino: As the one above, I'd suggest to use the autocomplete attribute directly into the input field instead of a wrapper div element. So you can keep this one and remove the other from the wrapper.

app/views/settings/dex_connector_ldaps/_form.html.slim
nanoscopic: Using "off" is not sufficient. It works in Chrome to fix the problem but not Firefox. The extra field seems unnecessary but we determined it is needed to correct the bug.

app/views/settings/dex_connector_ldaps/_form.html.slim
nanoscopic: Yes it is necessary.

vitoravelino: Ok. I think I gotcha. If that's necessary, there's a better trick to do that instead of adding a text field and hiding it via css. Try to use a `hidden_field` instead of `text_field`. With that change you can remove the `settings-hidden` class from css too.

nanoscopic: A field that is hidden does not work. We tried that. This is the solution.

app/views/settings/dex_connector_ldaps/_form.html.slim
nanoscopic: It should be on the form as well for completeness for some browsers in some scenarios. This is a micro-optimization. There is no need to remove this as it doesn't hurt anything and potentially helps.

PR #635, opened 2018-08-21
Updated tool tips and removed default data

Klaven:
- Added a database migration that removed default data from the database
 - Added smart documentation on each LDAP field in place of the default data.

### DIFF
--- a/app/views/settings/dex_connector_ldaps/_form.html.slim
+++ b/app/views/settings/dex_connector_ldaps/_form.html.slim
@@ -3,6 +3,8 @@
     = f.label :name
     = f.text_field :name, class: "form-control", value: @certificate_holder.name
     = error_messages_for(@certificate_holder, :name)
+    small.form-text.text-muted
+      | Name shown to user when selecting a connector
 
   hr
   h4 Server
@@ -11,11 +13,19 @@
     = f.label :host
     = f.text_field :host, class: "form-control", value: @certificate_holder.host
     = error_messages_for(@certificate_holder, :host)
+    small.form-text.text-muted
+      | Host name of LDAP server reachable from the cluster
 
   .form-group class="#{error_class_for(@certificate_holder, :port)}"
     = f.label :port
     = f.text_field :port, class: "form-control", value: @certificate_holder.port
     = error_messages_for(@certificate_holder, :port)
+    small.form-text.text-muted
+      | The port on which to connect to the host (e.g. StartTLS On:
+      code 389
+      | , StartTLS Off:
+      code 636
+      | )
 
   .form-group class="#{error_class_for(@certificate_holder, :start_tls)}"
     = f.label :start_tls, "StartTLS"
@@ -28,10 +38,13 @@
         = f.radio_button :start_tls, "false", checked: !@certificate_holder.start_tls
         | Off
     = error_messages_for(@certificate_holder, :start_tls)
+    br
+    small.form-text.text-muted 
+      | When enabled use StartTLS otherwise TLS will be used
 
   .form-group.form-group-certificate class="#{error_class_for(@cert, :certificate) if @cert.present?}"
     = f.label :certificate
-    p Upload the certificate for the LDAP server here
+    p Upload the certificate of the root CA that issued the LDAP server certificate
     = f.file_field :certificate, class: "form-control", required: @cert.present?
     - if @cert.present?
       = error_messages_for(@cert, :certificate)
@@ -51,12 +64,17 @@
         = f.radio_button :bind_anon, "false", checked: !@certificate_holder.bind_anon
         | False
     = error_messages_for(@certificate_holder, :bind_anon)
+    br
+    small.form-text.text-muted 
+      | Use anonymous authentication to do initial user search
 
   .ldap-dn-authentication.collapse class="#{'in' if !@certificate_holder.bind_anon}" id="anonymous_bind_toggle" aria-expanded="true"
     .form-group class="#{error_class_for(@certificate_holder, :bind_dn)}" autocomplete="nope"
       = f.label :bind_dn, "DN"
       = f.text_field :bind_dn, class: "form-control", value: @certificate_holder.bind_dn, required: !@certificate_holder.bind_anon, disable: @certificate_holder.bind_anon
-      = error_messages_for(@certificate_holder, :bind_dn)
+      = error_messages_for(@certificate_holder, :bind_dn)    
+      small.form-text.text-muted
+        | Bind DN of user that can do user searches
 
 
     .form-group class="#{error_class_for(@certificate_holder, :bind_pw)}" autocomplete="nope"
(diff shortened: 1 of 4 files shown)


app/views/settings/dex_connector_ldaps/_form.html.slim
```diff
@@ -11,11 +13,18 @@
     = f.label :host
     = f.text_field :host, class: "form-control", value: @certificate_holder.host
     = error_messages_for(@certificate_holder, :host)
+    small.form-text.text-muted id="#{@certificate_holder.host}_ldap_host_help"
+      | Host name of LDAP server reachable from the cluster
 
   .form-group class="#{error_class_for(@certificate_holder, :port)}"
     = f.label :port
-    = f.text_field :port, class: "form-control", value: @certificate_holder.port
```
vitoravelino: You can revert this since the user can go back and forth through the pages and they wouldn't be seeing what they submitted previously.

app/views/settings/dex_connector_ldaps/_form.html.slim
```diff
@@ -11,11 +13,18 @@
...
+    = f.text_field :port, class: "form-control"
     = error_messages_for(@certificate_holder, :port)
+    small.form-text.text-muted id="#{@certificate_holder.host}_ldap_port_help"
```
vitoravelino: Any reason to add the `id` attribute?

app/views/settings/dex_connector_ldaps/_form.html.slim
Klaven: reverted.

r0ckarong:
@Klaven Please review the comments in the trello card regarding the descriptions: 
https://trello.com/c/Ipsm0Da9/940-3-external-auth-ldap-documentation?menu=filter&filter=external

spec/features/settings/dex_connector_ldap_feature_spec.rb
```diff
@@ -63,14 +63,29 @@
...
+      fill_in id: "dex_connector_ldap_name", with: "name"
+      fill_in id: "dex_connector_ldap_group_attr_user", with: "user"
+      fill_in id: "dex_connector_ldap_group_attr_group", with: "group"
+      fill_in id: "dex_connector_ldap_user_attr_id", with: "uid"
+
       page.execute_script("$('#ldap_conn_save').removeProp('disabled')")
       click_button("Save")
-
       last_ldap_connector = DexConnectorLdap.last
-      expect(page).to have_content(admin_cert_text)
```
vitoravelino: No need to change this, you can revert it.

nanoscopic: The change to the expect for the cert is required for the test to pass. It is unclear why but this is in fact the case. "you can revert it" is not true from what I've seen.

vitoravelino: The previous and the current change does the same thing in the end. The reason I suggested to revert was to maintain consistency on the written code. I just reverted and ran the test locally and it passed.

nanoscopic: The following three things all showed the old version without the to_text breaking:
1. My machine running the tests
2. Marek's machine running the tests
3. The post-commit tests run automatically

The fixes should remain. By your own admission they should be the same so it is a waste of effort and time to discuss or revert this.

We did try more specific tests that just test the textarea as that would be the most proper method rather than testing the whole page but they all failed. This fix is hacky and suboptimal.

Both the original and the new form are not good, so "maintaining consistency" of hacky code ceases to become relevant. The test works.

ereslibre: I think something is flaky in this test (didn't have the time to properly check what though). I observed an issue with seed `19873`, and in this case, it's reproducible in both cases (with the `have_content` and `have_text` variant); at least on my computer. This is what I did to confirm it:

- With ` expect(page).to have_content(admin_cert_text)`:

```
admin:~ # docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bash -c "RAILS_ENV=test bundle exec rspec --seed 19873 spec/features/settings/dex_connector_ldap_feature_spec.rb"

Randomized with seed 19873
...........

Finished in 22.53 seconds (files took 3.15 seconds to load)
11 examples, 0 failures

Randomized with seed 19873

Coverage report generated for RSpec to /srv/velum/public/coverage. 440 / 2712 LOC (16.22%) covered.
Coverage (16.22%) is below the expected minimum coverage (100.00%).
admin:~ # docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bash -c "RAILS_ENV=test bundle exec rspec --seed 19873 spec/features/settings/dex_connector_ldap_feature_spec.rb"

Randomized with seed 19873
......^C
RSpec is shutting down and will print the summary report... Interrupt again to force quit.
^C
```
Sometimes works, but others it's stuck, I presume it's capybara waiting.

- With `expect(page.text).to have_text(admin_cert_text)` I can get the example failing in a stable manner

```
admin:~ # docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bash -c "RAILS_ENV=test bundle exec rspec --seed 19873 spec/features/settings/dex_connector_ldap_feature_spec.rb"

Randomized with seed 19873
......F....

Failures:

  1) Feature: LDAP connector settings #new allows a user to create an ldap connector (w/ certificate)
     Failure/Error: expect(page.text).to have_text(admin_cert_text)
       expected to find text "-----BEGIN CERTIFICATE----- MIICnzCCAYegAwIBAgIBADANBgkqhkiG9w0BAQUFADATMREwDwYDVQQDDAhob3N0 bmFtZTAeFw0xODA4MjcyMDA3MDZaFw0yMDA4MjYyMDA3MDZaMBMxETAPBgNVBAMM CGhvc3RuYW1lMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsM24K5xd NkT3ZA41DI0Gt0umpLLQXnsykmsnnoLsDzTpBpW77oE8d2zerVoVRuVGRaKGmPj/ akBzrDIsuak3flgr63Rb5/hpX3nZ7lMF/L8DkZ5L0vfC/Tfu2PCKNiBVRmqG0p7J nN0EzCJ/+OuNuXjbW+OzJWapzszE+38zs01gfr8auYTm/zzgC5gIrVHyycRoWTkO mLWSsbBdPYVbxJYti1a0svlZapECRqbcnq4ighn0f+Xl8gCqeQIff9f3OKf0tRVR JqtTG6EhDLVcfDUKFCvRSKN3mMb8L3cnbwQJO4/FPnTYa03os0XtOiG496ClIO7o MNKikuYjFTFdbQIDAQABMA0GCSqGSIb3DQEBBQUAA4IBAQB3tilTGNQop/A1OU+M C34Az9de0L7aIP8JDyWYrANt85TQibitpF/3k9Da0K2ReRbW6QvKKJr7C9JO88qu DZuzi0wzaM6LPXSAeIwiVzYbDpQ/Nmdtz4yRvO8XZ19I+jdepPDUMbE9tn3ylafS yiBGlcZ+/mIQm0Lw1IEMfR/hZeQ10kgk2pRgy0FaVWKL9dw1gouyKe7bXGVDfAYI pW45CRDxQYOOQ/VBsQE7YFuGSX0nLcDnnUvkwRJaIwtEx9jYcDb3XKiWdx5oVy+j s4c/gklQnBJVYo0g8pcSPlD+IDVWQNgrYslgUKcYr1R05xpgRST05bu+cgGNbZCQ IPht -----END CERTIFICATE-----" in "Home Settings Logout REGISTRIES Remote Registries Mirrors System wide certificates EXTERNAL AUTHENTICATION LDAP Connectors KUBERNETES Compute Resources Reservations Auditing New LDAP Connector NameName shown to user when selecting a connector Server HostHost name of LDAP server reachable from the cluster PortThe port on which to connect to the host (e.g. StartTLS On:389, StartTLS Off:636) StartTLS OnOff When enabled use StartTLS otherwise TLS will be used Certificate Upload the certificate of the root CA that issued the LDAP server certificate Certificate can't be blank Authentication Anonymous TrueFalse Use anonymous authentication to do initial user search DNBind DN of user that can do user searches PasswordPassword of the user User Search Identifying User AttributeLabel of LDAP attribute users will enter to identify themselves (e.g.username) Base DNBaseDN where users are located (e.g.cn=users,dc=example,dc=com) FilterFilter to specify type of user objects (e.g.\"(objectClass=person)\") User Attribute Map UsernameAttribute users will enter to identify themselves IDAttribute used to identify user within the system (e.g.uid) EmailAttribute containing email of users NameAttribute used as username used within OIDC tokens Group Search Base DNBaseDN where groups are located (e.g.cn=groups,dc=example,dc=com) FilterFilter to specify type of group objects. (e.g.\"(objectClass=group)\") Group Attribute Map UserAttribute to map as user (e.g.uid) GroupAttribute identifying membership (e.g.member) NameAttribute to map as name (e.g.name) CancelTest Connection Test Connection first before saving SUSE® CaaS Platform 4.0.0+dev | © SUSE Linux 2018"
     # ./spec/features/settings/dex_connector_ldap_feature_spec.rb:88:in `block (3 levels) in <top (required)>'

Finished in 22.27 seconds (files took 3.16 seconds to load)
11 examples, 1 failure

Failed examples:

rspec ./spec/features/settings/dex_connector_ldap_feature_spec.rb:63 # Feature: LDAP connector settings #new allows a user to create an ldap connector (w/ certificate)

Randomized with seed 19873

Coverage report generated for RSpec to /srv/velum/public/coverage. 439 / 2712 LOC (16.19%) covered.
```

It looks to me like some test is being flaky, or maybe depends on previous state to pass, this can lead to future runs on CI randomly failing. Can you confirm if this happens to you too? (please make sure to try with `docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bash -c "RAILS_ENV=test bundle exec rspec --seed 19873 spec/features/settings/dex_connector_ldap_feature_spec.rb"` command).

nanoscopic: Agreed that stability still existed in both cases. This has now been addressed by enforcing order of the tests being run. It appears that either capybara itself or the underlying phantamjs is preserving some state that it shouldn't in between tests.

It is clear that the tests themselves are fine because it is possible to make it work as well by adding a dummy test in between tests. Doing that somehow causes the preserved state the breaks the functionality to be cleared. Specifying the order also resolves the problem, even when you place the tests in the 'bad order'. That is, specifying to run the tests in order appears to prevent capybara from having the bug that is being seen.

More investigation and consideration is warranted but should not continue as part of this PR. This PR is not even directly related to these breaking tests, we have only fixed them as a side effect of seeing it breaking.

Once this passes CI it should be merged and the discussion of these flaky tests, ordering, and instability of capybara and/or phantomjs be moved elsewhere.

Issue #399, opened 2017-12-11
Pillar endpoint does not support lists


jgleissner:
The introduction of the pillar endpoint https://github.com/kubic-project/velum/pull/315 removes support for lists in pillar data, which used to be available through the external SQL pillar in salt master. For public cloud adaption, we want to store salt cloud specific data in the pillar, which requires a structure like this:
```
  profiles:
    cluster_node:
      block_device_mappings:
      - DeviceName: /dev/sda1
        Ebs.VolumeSize: 30
      image: ami-12345678
      network_interfaces:
      - AssociatePublicIpAddress: false
        DeviceIndex: 0
        SecurityGroupId: sg-1234abcd
        SubnetId: subnet-1234abcd
      provider: ec2
```
In this example, network_interfaces is an array consisting of one multi-key dictionary. It would be extremely helpful if structures like those could be served through the velum pillar.

ereslibre:
The old pillar implementation didn't support lists as they were implemented as collisions. This means that something like: `my_list: ["something"]` was unfeasible (one value would mean scalar, a collision would mean a list with 2 items -- hence, there was not way to represent lists of arbitrary length, e.g. of size 1).

The new pillar allows *anything* to be rendered, including strings, lists, nested hashes or any kind of type json supports. Basically anything.

In order to render anything as you want you only need  to modify: https://github.com/kubic-project/velum/blob/master/app/controllers/internal_api/v1/pillars_controller.rb

If you extend `pillar_contents` having lists, hashes, strings, integers, floats... everything will work. So we have right now a much more rich pillar in reality, what we didn't with the old implementation.

ereslibre:
So the idea is that you can add new models that makes sense to you (no need to stick to the `Pillar` model), and then render that on the `PillarController` as you wish, this will be merged by salt correctly and any type and complexity is allowed.

jgleissner:
The old pillar supports lists of size 1, if you specify which key is a list key with the with_lists keyword (in theory, it's actually broken, but that's a different story). It's awkward though and I'm not arguing to get the old pillar back. However, we'll need lists and the current implementation in pillar controller really only supports a key=value model.

I haven't thought of adding a separate endpoint for cloud specific data. Not sure whether that gains us anything, since we need the pillars in the salt master so they are automatically available to salt cloud. I'm gonna look into it. Thanks for the info so far.

ereslibre:
[...]

PR #636, opened 2018-08-22
Fix pillar issue in public cloud

jgleissner:
I've added a call_runner function to salt interface and added a call to sync_pillar runner in cloud_cluster.rb. This is required to activate the external pillar, which is normally done at orchestration time, but for creating cluster nodes in the cloud, the pillar is needed before that. See also #399 for details.

I've also added a sleep(1) after the refresh_pillar, as we ran into the issue that the succeeding salt-cloud still works on the old pillar data, and this seems to work around it. I know that this is a bit ugly and perhaps can be fixed better in salt directly, but I haven't had any luck yet with narrowing the issue down.

### DIFF
--- a/app/models/cloud_cluster.rb
+++ b/app/models/cloud_cluster.rb
@@ -68,6 +68,10 @@ class CloudCluster
     persist_to_pillar!(:cloud_worker_security_group, @security_group_id)
     Velum::Salt.call_runner(action: "saltutil.sync_pillar")
     Velum::Salt.call(targets: "admin", action: "saltutil.refresh_pillar")
+    # During tests, the salt-cloud call that immediately follows the pillar
+    # refresh was found to sometimes operate on the old pillar data and
+    # thus fail. The sleep(1) works around this issue. Ideally this should
+    # be resolved in salt and the work-around removed.
     sleep(1)
   end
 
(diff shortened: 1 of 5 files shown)


jordimassaguerpla:
Hi @jgleissner ! Could you check the failing tests in travis?

jgleissner:
Hi, I'm looking into it. I believe it's because I didn't describe the call_runner function in spec/lib/velum/salt_spec.rb. I'm not really familiar with VCR cassettes and all that. It'll take me a bit to get up to speed on this. Sorry about that.

jgleissner:
Hi @jordimassaguerpla , is there an easy way to run that travis test locally?

jgleissner:
Don't worry about it, I think I got the test suite running ok.

ereslibre:
However, travis is red...

app/models/cloud_cluster.rb
```diff
@@ -66,7 +66,9 @@ def save!
     persist_to_pillar!(:cloud_worker_net, @network_id)
     persist_to_pillar!(:cloud_worker_subnet, @subnet_id)
     persist_to_pillar!(:cloud_worker_security_group, @security_group_id)
-    Velum::Salt.call(action: "saltutil.refresh_pillar")
+    Velum::Salt.call_runner(action: "saltutil.sync_pillar")
+    Velum::Salt.call(targets: "admin", action: "saltutil.refresh_pillar")
+    sleep(1)
```
MaximilianMeister: why do we need the sleep(1)? maybe add a comment

jgleissner: Makes sense, I'll add one.

jgleissner:
Regarding the failing test, I believe this is due to the new runner call that does a http request to 127.0.0.1/run, which wasn't the case before, so the recorded cassette doesn't match anymore. I've been trying to figure out how this stuff works and made some progress, but haven't managed to record a new one yet. Sorry about that. If you can point me to any useful documentation etc. I'd appreciate it.

PR #638, opened 2018-08-23
Update channels

MaximilianMeister:
### Depends on

https://github.com/kubic-project/salt/pull/654

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -56,7 +56,12 @@ MinionPoller = {
         var pendingRendered = "";
         var masterApplied = false;
         var updateAvailable = false;
+        var migrationAvailable = false;
         var updateAvailableNodeCount = 0;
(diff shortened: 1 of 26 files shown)


app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -287,25 +287,45 @@ MinionPoller = {
   },
 
   handleAdminUpdate: function(admin) {
```
vitoravelino: Instead of handling everything here, we could extract all the migration type into something like `handleAdminMigrationUpdate` and `handleAdminMaintenanceUpdate`. Trying to stay sane... :sweat_smile:

MaximilianMeister: yeah, good idea! this is still very raw ;) I'll try to get the back-end things working and then make the UI prettier

MaximilianMeister:
please review https://github.com/kubic-project/velum/pull/654 first

ereslibre:
This PR currently has conflicts with our schema patch

PR #641, opened 2018-08-29
use online status during update for update status

MaximilianMeister:
we can make use of the online status to have a more accurate status
during the update

if a node is offline during the update, it likely means that it's
rebooting

### Screenshot of current status

It could be misleading to see the node offline during the update. Without knowing why, it looks like sth went wrong

update-reboot

### Screenshot with the PR applied

update-rebooting


MaximilianMeister:
tested with a real update

PR #645, opened 2018-09-02
Fix conflicts with cert upload and ldap connection test

nanoscopic:
The change from a text box to enter certificate content to a file upload broke the functionality of the LDAP external auth feature. That feature requires the inforation entered to be tested to connect to the ldap server before saving. No one tested that after the file upload change and hence it is broken and had to be updated. These changes make it work once more.

The file upload could not be used for the ldap external auth as modern browsers block getting the contents of a file selected for upload in javascript when that file has the extension ".crt" or ".pem". As a result the only way to fetch the cert to use it for the check would be to upload it to the backend somewhere temporarily.

The cert for external auth, though, should never have been changed to a file upload. The planned addition is a button to fetch from the server automatically and confirm to the user the fingerprint and information of the cert. Connecting the external auth to the other certificate handling in the system is not appropriate at this point in time.

This change also includes another fix for ldap connection test to be able to properly fetch the password. That is another breaking change that was not detected, also because there is no automated UI capybara test against the ldap connection stuff. An automated test to prevent this from breaking repeatedly will need to be added in the future as well. It is not included in this commit intentionally. This commit is intended just to get the feature back into a working state having been left broken.

### DIFF
--- a/app/controllers/settings/dex_connector_ldaps_controller.rb
+++ b/app/controllers/settings/dex_connector_ldaps_controller.rb
@@ -1,6 +1,36 @@
 # Settings::DexConnectorLdapsController is responsible to manage requests
 # related to LDAP connectors.
 class Settings::DexConnectorLdapsController < Settings::BaseCertificateController
+  def new
+    @certificate_holder = certificate_holder_type.new
+  end
+
+  def create
+    @certificate_holder = certificate_holder_type.new(
+      certificate_holder_params
+    )
+
+    ActiveRecord::Base.transaction do
+      @certificate_holder.save!
+    end
+
+    redirect_to [:settings, @certificate_holder],
+                notice: "#{@certificate_holder.class} was successfully created."
+  rescue ActiveRecord::RecordInvalid
+    render action: :new, status: :unprocessable_entity
+  end
+
+  def update
+    ActiveRecord::Base.transaction do
+      @certificate_holder.update_attributes!(certificate_holder_update_params)
+    end
+
+    redirect_to [:settings, @certificate_holder],
+                notice: "#{@certificate_holder.class} was successfully updated."
+  rescue ActiveRecord::RecordInvalid
+    render action: :edit, status: :unprocessable_entity
+  end
+
   def index
     @ldap_connectors = DexConnectorLdap.all
   end
@@ -22,7 +52,7 @@ class Settings::DexConnectorLdapsController < Settings::BaseCertificateControlle
   end
 
   def certificate_holder_update_params
-    ldap_connector_params.except(:certificate, :current_cert)
+    ldap_connector_params
   end
 
   private
@@ -34,6 +64,6 @@ class Settings::DexConnectorLdapsController < Settings::BaseCertificateControlle
                                                :user_base_dn, :user_filter, :user_attr_username,
                                                :user_attr_id, :user_attr_email, :user_attr_name,
                                                :group_base_dn, :group_filter, :group_attr_user,
-                                               :group_attr_group, :group_attr_name)
+                                               :group_attr_group, :group_attr_name, :certificate)
   end
 end
--- a/app/models/dex_connector_ldap.rb
+++ b/app/models/dex_connector_ldap.rb
@@ -1,7 +1,5 @@
 # Model that represents a dex authentication connector for LDAP
 class DexConnectorLdap < ActiveRecord::Base
-  has_one :certificate_service, as: :service, dependent: :destroy
-  has_one :certificate, through: :certificate_service
   self.table_name = "dex_connectors_ldap"
 
   validates :name, presence: true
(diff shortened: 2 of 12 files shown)


app/controllers/settings/dex_connector_ldaps_controller.rb
```diff
@@ -1,6 +1,40 @@
...
+  end
+
+  def update
+    ActiveRecord::Base.transaction do
+      @certificate_holder.update_attributes!(certificate_holder_update_params)
+    end
+
+    ActiveRecord::Base.transaction do
+      @certificate_holder.update_attributes!(certificate_holder_update_params)
+    end
```
MaximilianMeister: this block looks redundant

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20181715075510) do
+ActiveRecord::Schema.define(version: 20182030152200) do
```
MaximilianMeister: we still need to rebase the [patch](https://github.com/kubic-project/velum/blob/master/packaging/suse/patches/0_set_default_salt_events_alter_time_column_value.rpm.patch) when changing the schema. example -> https://github.com/kubic-project/velum/pull/524/commits/4ec7dbab1666f77a1b236a65e1d2e83b5f80bcb3

app/models/dex_connector_ldap.rb
```diff
@@ -21,4 +19,5 @@ class DexConnectorLdap < ActiveRecord::Base
   validates :group_attr_user, presence: true
   validates :group_attr_group, presence: true
   validates :group_attr_name, presence: true
+  validates :certificate, presence: true
```
bergmannf: Maybe not needed, as the validity is already checked in `javascript`, but this could also use the `x509_validator` (`x509_certificate: true`) used for the other certificates as a second check.

bergmannf:
Just a FYI as I got curious about not being able to read `crt` and `pem` files in Javascript - I tested it in [`jsfiddle`](https://jsfiddle.net/kwtdjvuy/18/) and that worked without problems for me (Firefox 60 + Chromium 68).

The files were `certificates` in both cases and my browser are not the most recent version I think - so maybe that's why it worked for me?

So if we want to change to file upload in the future it might still be possible after all.

vitoravelino:
> Just a FYI as I got curious about not being able to read crt and pem files in Javascript - I tested it in jsfiddle and that worked without problems for me (Firefox 60 + Chromium 68).

@bergmannf I was also able to reproduce it as you've mentioned, also in this [codepen example](https://codepen.io/anon/pen/bxRbMG) and inside velum with a quick copy and paste. I've used Firefox 60 (leap15), 61 (win10) and Chrome 68 (leap15, win10).

> The file upload could not be used for the ldap external auth as modern browsers block getting the contents of a file selected for upload in javascript when that file has the extension ".crt" or ".pem".

@nanoscopic could you help us to reproduce it or point out an example of that happening?

app/models/dex_connector_ldap.rb
nanoscopic: As it is written currently it does not store the begin cert / end cert lines in the DB. I am not sure but I believe this will cause it not to pass x509 validity.

Agreed also that the check is done in JS, but I haven't checked all situations here. For example I believe if you enter valid values, hit the 'test connection' button, and then alter the values you can then hit save with bad values. That should be correct and is a bug I need to enter for the setup. This PR notably is to fix the fact that head is broken and unusable for external auth, not to correct all possible problems. That will be addressed in a subsequent change and PR.

nanoscopic:
If you have code that works letting JS read a selected file, please show it. I wouldn't just make up that it is disallowed. I don't have any additional extensions preventing that. I am using Firefox 60 and it is blocked.

vitoravelino:
> ... I was trying to read it when a file has already been selected. 

By that you mean when the user clicks on the "test connection" button? Something like [this](https://codepen.io/anon/pen/WgOjJN)? For me it worked the same way in both cases.

nanoscopic:
Using the demos I see the same thing being said, that pem files can be read through JS using the same mechanism I tried.

Despite this, when I tested in the system itself, within a modified velum, it definitely did not work unless the file did not end in '.pem'. I searched the internet to attempt to find mention of this "security feature" in Firefox but I see no evidence of such behavior. I grep'd through the source code of Firefox ( gecko ) to see if I could locate the code that causes the failure as well; I as not able to find code that would cause it.

I could change the system back to test this again and further look into that particular aspect, but as explained in more detail via email; it is undesirable to use a file upload for external ldap authentication certs. It will soon be changed to be a button to get the cert from the server directly, and that will be more straightforward to do using a textarea or hidden field instead of a file upload.

As a result, I see no reason to spend more effort getting reading of the file browse via JS working within the external ldap auth feature.

app/controllers/settings/dex_connector_ldaps_controller.rb
nanoscopic: Agreed; removing

db/schema.rb
nanoscopic: Have updated the patch.

PR #647, opened 2018-09-04
FIX bsc#1099985 removes unassigned nodes and add a reject node link

chentex:
This adds the possibility to remove an unassigned node, from the assign_nodes page.
I also added a reject node link to the unassigned nodes, seemed to me that that functionality was missing.
*The setup step discovery can also do this.

**CATCH/HELP:** There is a time between when the remove node it's clicked and accepted and when the salt event removes that node from the DB, that the user could assign that node to a role and apply the changes, that leaves the cluster in an unstable state. If someone could help me to figure this out I would appreciate it, maybe disabling it via JS or something, don't know the correct way to approach it.

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -268,20 +268,25 @@ MinionPoller = {
   },
 
   renderPendingNodes: function(pendingMinionId) {
-    var acceptHtml;
+    var acceptHtml = '';
+    var rejectHtml = '';
 
     if (State.hasPendingStateNode || State.pendingRemovalMinionId) {
       acceptHtml = '';
+      rejectHtml = '';
     } else if (hasPendingAcceptance(pendingMinionId)) {
       acceptHtml = 'Acceptance in progress';
+    } else if (hasPendingRejection(pendingMinionId)) {
+      rejectHtml = 'Rejection in progress';
     } else {
-      acceptHtml = '<a class="accept-minion" href="#" data-minion-id="' + pendingMinionId + '">Accept Node</a>';
+      acceptHtml = '<a class="accept-minion" href="#" data-minion-id="' + pendingMinionId + '">Accept</a>';
+      rejectHtml = '&nbsp;|&nbsp;<a class="reject-minion" href="#" data-minion-id="' + pendingMinionId + '">Reject</a>';
     }
 
     return '\
-      <tr> \
+      <tr class="minion_'+ pendingMinionId +'"> \
         <td>' + pendingMinionId + '</td>\
-        <td class="pending-accept-link">' + acceptHtml + '</td>\
+        <td class="pending-accept-link">' + acceptHtml + rejectHtml + '</td>\
       </tr> \
     ';
   },
@@ -443,43 +448,65 @@ MinionPoller = {
     var isWorker = MinionPoller.selectedNodes.indexOf(minion.id) !== -1;
     var isUnused = !isMaster && !isWorker;
 
-    var masterHtml;
     var masterChecked = '';
-    var minionHtml;
     var minionChecked = '';
+    var disabledHtml = '';
+    var disabledClass = '';
+
+    if (State.pendingRemovalMinionId != null){
+      disabledHtml = 'disabled="disabled"';
+      disabledClass = 'disabled';
+      $('.select-nodes-btn').prop('disabled', true);
+    }else{
+      $('.select-nodes-btn').prop('disabled', false);
+    }
 
     if (isMaster) {
       masterChecked = 'checked';
     }
 
-    masterHtml = '\
-      <input name="roles[master][]" id="roles_master_' + minion.id +
+    var masterHtml = '\
+      <input hidden name="roles[master][]" id="roles_master_' + minion.id +
       '" value="' + minion.id + '" type="checkbox" title="Select node as master" ' + masterChecked + '>';
 
     if (isWorker) {
       minionChecked = 'checked';
     }
 
-    minionHtml = '<input name="roles[worker][]" id="roles_worker_' + minion.id +
+    var minionHtml = '\
+      <input hidden name="roles[worker][]" id="roles_worker_' + minion.id +
       '" value="' + minion.id + '" type="checkbox" title="Select node for bootstrapping" ' + minionChecked + '>';
-
     var roleHtml = '\
-      <td class="role-column">' +
+      <td class="role-column" id="role-column-'+ minion.minion_id +'">' +
         masterHtml +
         minionHtml +
         '<div class="btn-group role-btn-group">\
-          <button type="button" class="btn btn-default master-btn '+ (isMaster ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="master">Master</button>\
-          <button type="button" class="btn btn-default worker-btn ' + (isWorker ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="worker">Worker</button>\
-          <button type="button" class="btn btn-default unused-btn ' + (isUnused ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '">Unused</button>\
+          <button type="button" '+ disabledHtml +' class="btn btn-default master-btn '+ (isMaster ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="master">Master</button>\
+          <button type="button" '+ disabledHtml +' class="btn btn-default worker-btn ' + (isWorker ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="worker">Worker</button>\
+          <button type="button" '+ disabledHtml +' class="btn btn-default unused-btn ' + (isUnused ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '">Unused</button>\
         </div>\
       </td>\
     ';
+
+    var actionHtml = '\
+      <td class="action-column">\
+        <a class="remove-minion '+ disabledClass +'" href="#" data-minion-id="' + minion.minion_id + '" data-minion-fqdn="' + minion.fqdn + '">Remove</a>\
+      </td>\
+    ';
+
+    switch (minion.highstate) {
+      case "pending_removal":
+        $('#set-roles').prop('disabled', true);
+        actionHtml = '<td class="action-column">Pending removal</td>';
+        break;
+    }
 
     return '\
       <tr class="minion_' + minion.id + '">' +
         '<td>' + minion.minion_id + '</td>\
         <td class="minion-hostname">' + minion.fqdn + '</td>' +
         roleHtml +
+        actionHtml +
       '</tr>';
   },
 
@@ -487,6 +514,44 @@ MinionPoller = {
     return MinionPoller.renderDiscovery(minion);
   }
 };
+
+function removeNode(minionId, hostname) {
+  var $alert = $('.failed-removal-alert');
+  var error = 'Failed to remove node '+ minionId +'. Please try again.';
+
+  if (confirm('Are you sure you want to remove ' + hostname + '?')) {
+    setPendingRemoval(minionId);
+    $.ajax({
+      url: '/minions/'+ minionId +'/remove-minion.json',
+      method: 'POST',
+      data: { minion_id: minionId },
+      dataType: 'text'
+    }).fail(function () {
+      $alert.remove();
+      removePendingRemoval();
+      showAlert(error, 'alert', 'failed-removal-alert');
+    });
+  }
+}
+
+function rejectNode(minionId) {
+  var $alert = $('.failed-rejection-alert');
+  var error = 'Failed to reject node '+ minionId +'. Please try again.';
+
+  if (confirm('Are you sure you want to reject ' + minionId + '?')) {
+    setPendingRejection(minionId);
+    $.ajax({
+      url: '/minions/'+ minionId +'/reject-minion.json',
+      method: 'POST',
+      data: { minion_id: minionId },
+      dataType: 'text'
+    }).fail(function () {
+      removePendingRejection(minionId);
+      $alert.remove();
+      showAlert(error, 'alert', 'failed-rejection-alert');
+    });
+  }
+}
 
 function isTheLast(minion, role) {
   var remainingValid = State.minions.filter(function (m) {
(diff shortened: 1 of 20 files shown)


app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -565,6 +606,22 @@ $('body').on('click', '.accept-minion', function(e) {
...
+  e.preventDefault();
+  deleteNode(minionId, fqdn);
+});
+
+$('body').on('click', '.reject-minion', function(e) {
+  var minionId = $(this).data('minion-id');
+  var fqdn = $(this).data('minion-fqdn')
+
+  e.preventDefault();
+  rejectNode(minionId, fqdn);
```
vitoravelino: Unfortunately we don't have access to the minion hostname/fqdn before accepting it, so in this case you can simply use the minion id.

lib/velum/salt.rb
```diff
@@ -70,6 +70,23 @@ def self.accept_minion(minion_id: "")
...
+                                    fun:    "key.delete",
+                                    match:  minion_id })
+    end
+
+    # Rejects a minion from the cluster
+    def self.reject_minion(minion_id: "")
+      res = perform_request(endpoint: "/", method: "post",
+                            data: { client: "wheel",
+                                    fun:    "key.reject",
+                                    arg:    "--incldue-accepted",
```
ereslibre: I don't think you can pass args like this (besides, there's a typo in `include`). I think this should be something like:

```
      res = perform_request(endpoint: "/", method: "post",
                            data: { client: "wheel",
                                    fun:    "key.reject",
                                    include_accepted: True,
                                    match:  minion_id })
```

config/routes.rb
```diff
@@ -25,6 +25,8 @@
   get "/_health", to: "health#index"
   post "/update", to: "salt#update"
   post "/accept-minion", to: "salt#accept_minion"
+  post "/remove-minion", to: "salt#remove_minion"
+  post "/reject-minion", to: "salt#reject_minion"
```
MaximilianMeister: should we put this API endpoints into the `/minions` resource namespace?

app/models/salt_handler/minion_start.rb
```diff
@@ -25,6 +25,9 @@ def process_event
     return false if ["ca", "admin"].include? id
 
     minion_info = Velum::Salt.minions[id]
+    puts Velum::Salt.minions
+    puts id
+    puts minion_info
```
MaximilianMeister: looks like this is only debug output? in general if you want to have logs appear in the velum container logs, just use e.g. `Rails.logger.debug(msg)`

app/models/salt_handler/auth_event.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize(salt_event)
+    @salt_event = salt_event
+  end
+
+  # This method is responsible for all actions needed when a minion starts.
+  # For now we simply store the Minion in our database.
+  def process_event
+    parsed_data = salt_event.parsed_data
+    puts parsed_data
```
MaximilianMeister: same here

app/models/salt_handler/auth_event.rb
```diff
@@ -0,0 +1,39 @@
...
+    # the k8s cluster.
+    id = parsed_data["id"]
+    return false if ["ca", "admin"].include? id
+
+    if "accept" == act
+      minion_info = Velum::Salt.minions[id]
+      return false if minion_info.blank?
+      # false if a minion with this minion_id or fqdn already exists (uniqueness validation)
+      Minion.new(minion_id: id, fqdn: minion_info["fqdn"]).save
+    elsif "pend" == act
```
MaximilianMeister: here i would perhaps use a case statement :

```ruby
case act
when "accept"
  ...
when "pend"
  ...
else
  ...
end
```

lib/velum/salt.rb
chentex: Thanks

app/models/salt_handler/minion_start.rb
chentex: Missed to delete this.

app/models/salt_handler/auth_event.rb
chentex: here too.

app/models/salt_handler/auth_event.rb
chentex: Ok

lib/velum/salt.rb
chentex: Will remove this line since we are only rejecting no accepted nodes.

config/routes.rb
vitoravelino: Maybe we could leave as it is since we left `accept` as that too unless we move accept too.

chentex: I moved all minion related to the resource namespace

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -443,39 +448,64 @@ MinionPoller = {
...
+        roleHtml = '<td class="role-column"></td>';
+        actionHtml = '\
+          <td class="action-column">'+
+            '<div class="remove-minion">\
+              Pending removal\
+            </div>\
+          </td>\
+        ';
+        break;
+      case "pending":
```
vitoravelino: You can remove these unused cases.

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -443,39 +448,64 @@ MinionPoller = {
...
+            masterHtml +
+            minionHtml +
+            '<div class="btn-group role-btn-group">\
+              <button type="button" class="btn btn-default master-btn '+ (isMaster ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="master">Master</button>\
+              <button type="button" class="btn btn-default worker-btn ' + (isWorker ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="worker">Worker</button>\
+              <button type="button" class="btn btn-default unused-btn ' + (isUnused ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '">Unused</button>\
+            </div>\
+          </td>\
+        ';
+        actionHtml = '\
```
vitoravelino: You can move everything from above this line out of the switch statement and for the buttons you can add something like

```js
var rolesDisabled = State.pendingRemovalMinionId != null;
// (...)

// add disabled attribute to the buttons like
// ...
'<button (...) disabled="' + rolesDisabled + '">Master</button>'\
'<button (...) disabled="' + rolesDisabled + '">Worker</button>'\
'<button (...) disabled="' + rolesDisabled + '">Unused</button>'\
// ...
```

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -443,39 +448,64 @@ MinionPoller = {
...
+          <td class="action-column">\
+            <a class="remove-minion"  href="#" data-minion-id="' + minion.minion_id + '" data-minion-fqdn="' + minion.fqdn + '">Remove</a>\
+          </td>\
+        ';
+        break;
+      case "pending_removal":
+        roleHtml = '<td class="role-column"></td>';
```
vitoravelino: With the suggestion above, no need of this.

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -443,39 +448,64 @@ MinionPoller = {
...
+        actionHtml = '\
```
vitoravelino: You can simply 

```js
actionHtml = '<td class="action-column">Pending removal</td>';
```

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -484,6 +514,45 @@ MinionPoller = {
...
+      showAlert(error, 'alert', 'failed-removal-alert');
+    });
+  }
+}
+
+function rejectNode(minionId) {
+  var $alert = $('.failed-rejection-alert');
+  var error = 'Failed to reject node '+ minionId +'. Please try again.';
+
+  setPendingRejection(minionId);
```
vitoravelino: Move it to inside of the `if` statement and remove the `else` block.

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -508,6 +577,24 @@ function removePendingAcceptance(minionId) {
...
+function setPendingRejection(minionId) {
+  sessionStorage.setItem('reject-'+minionId, true);
+  $('.pending-nodes-container a[data-minion-id="' + minionId + '"]').parent().text('Rejection in progress');
+}
+
+function removePendingRejection(minionId) {
+  sessionStorage.removeItem('reject-'+minionId);
+}
+
+function setPendingRemoval(minionId) {
```
vitoravelino: We can enhance what this method is doing like

```js
function setPendingRemoval(minionId) {
  $('.nodes-container a[data-minion-id="' + minionId + '"]').parent().text('Pending removal');
  $('#role-column-' + minionId + ' .btn').prop('disabled', true);
  $('#set-roles').data('previous', $('#set-roles').prop('disabled'));
  $('#set-roles').prop('disabled', true);
}
```

And add a method to revert the `Next` button in case the request fails. Call this in the `fail` callback. 

```js
function removePendingRemoval(minionId) {
  $('#set-roles').data('previous', $('#set-roles').prop('disabled'));
}
```

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -439,43 +444,58 @@ MinionPoller = {
...
+    }
 
     if (isMaster) {
       masterChecked = 'checked';
     }
 
-    masterHtml = '\
-      <input name="roles[master][]" id="roles_master_' + minion.id +
+    var masterHtml = '\
+      <input hidden name="roles[master][]" id="roles_master_' + minion.id +
```
vitoravelino: Any reason to add this `hidden` attribute?

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -439,43 +444,58 @@ MinionPoller = {
...
     }
 
-    minionHtml = '<input name="roles[worker][]" id="roles_worker_' + minion.id +
+    var minionHtml = '\
+      <input hidden name="roles[worker][]" id="roles_worker_' + minion.id +
       '" value="' + minion.id + '" type="checkbox" title="Select node for bootstrapping" ' + minionChecked + '>';
-
     var roleHtml = '\
-      <td class="role-column">' +
+      <td class="role-column" '+ disabledHtml +' id="role-column-'+ minion.minion_id +'">' +
```
vitoravelino: No need of `disabled` attribute in `td` element.

app/assets/javascripts/dashboard/dashboard.js
chentex: When disabling the buttons a checkbox became visible.

app/assets/javascripts/dashboard/dashboard.js
chentex: ok

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -508,6 +566,30 @@ function removePendingAcceptance(minionId) {
...
+  $('.pending-nodes-container a[data-minion-id="' + minionId + '"]').parent().text('Rejection in progress');
+}
+
+function removePendingRejection(minionId) {
+  sessionStorage.removeItem('reject-'+minionId);
+}
+
+function setPendingRemoval(minionId) {
+  $('.nodes-container a[data-minion-id="' + minionId + '"]').parent().text('Pending removal');
+  $('#role-column-' + minionId + ' .btn').prop('disabled', true);
```
vitoravelino: After testing I was able to identify that there was some delay disabled the other rows buttons, so we can safely replace this selector with `.role-column .btn` like

```js
$('.role-column .btn').prop('disabled', true);
```

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -508,6 +566,30 @@ function removePendingAcceptance(minionId) {
...
+function setPendingRejection(minionId) {
+  sessionStorage.setItem('reject-'+minionId, true);
+  $('.pending-nodes-container a[data-minion-id="' + minionId + '"]').parent().text('Rejection in progress');
+}
+
+function removePendingRejection(minionId) {
+  sessionStorage.removeItem('reject-'+minionId);
+}
+
+function setPendingRemoval(minionId) {
```
vitoravelino: We also forgot to disable the other `Remove` links, so adding 

```js
$('.remove-minion').addClass('disabled');
```

would do the trick.

vitoravelino: Remember to do the opposite on the `removePendingRemoval` method.

vitoravelino:
The `Next` button was being enabled after a few seconds, that was because of the `handleBootstrapErrors` method.

You can add something like

```js
if (State.pendingRemovalMinionId) {
  return;
}
```

to avoid enabling the next button while a removal is happening.

chentex:
@vitoravelino I tried adding the lines to disable and enable the next button but it messes with the minions lists and they stop showing anything

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -484,6 +506,50 @@ MinionPoller = {
...
+
+  if (confirm('Are you sure you want to remove ' + hostname + '?')) {
+    setPendingRemoval(minionId);
+    $.ajax({
+      url: '/minions/'+ minionId +'/remove-minion.json',
+      method: 'POST',
+      data: { minion_id: minionId },
+      dataType: 'text'
+    }).done(function(){
+      removePendingRemoval();
```
vitoravelino: By calling this here you'll re-enable things for a few seconds until the next fetch happens, which is an undesired behavior. Leftover while testing?

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -439,43 +444,60 @@ MinionPoller = {
...
+          <button type="button" '+ disabledHtml +' class="btn btn-default master-btn '+ (isMaster ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="master">Master</button>\
+          <button type="button" '+ disabledHtml +' class="btn btn-default worker-btn ' + (isWorker ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '" data-minion-role="worker">Worker</button>\
+          <button type="button" '+ disabledHtml +' class="btn btn-default unused-btn ' + (isUnused ? 'btn-primary' : '') + '" data-minion-id="' + minion.id + '">Unused</button>\
         </div>\
       </td>\
     ';
 
+    var actionHtml = '\
+      <td class="action-column">\
+        <a class="remove-minion"  href="#" data-minion-id="' + minion.minion_id + '" data-minion-fqdn="' + minion.fqdn + '">Remove</a>\
```
vitoravelino: We forgot to check if there's a pending removal to render this already disabled. Just need to add the class `disabled` if there's a pending removal.

chentex: Thank you (Y)

app/assets/javascripts/dashboard/dashboard.js
chentex: Noticed that behavior didn't know what was doing it, thanks.

chentex:
@MaximilianMeister here is the Screen Shot for the discovery step.

screenshot from 2018-09-20 13-40-16

chentex:
Done with your comment @MaximilianMeister

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -443,43 +448,65 @@ MinionPoller = {
...
     var masterChecked = '';
-    var minionHtml;
     var minionChecked = '';
+    var disabledHtml = '';
+    var disabledClass = '';
+
+    if (State.pendingRemovalMinionId != null){
+      disabledHtml = 'disabled="disabled"';
+      disabledClass = 'disabled';
+      $('.select-nodes-btn').prop('disabled', true);
```
chentex: Done @vitoravelino

flavio:
Shouldn't we backport that to the 3.0 branch?

PR #648, opened 2018-09-05
Fix test connection button for ldap external auth

nanoscopic:
Corrected fetching of password for ldap external auth

Corrected grabbing of certificate content for validating ldap external auth

### DIFF
--- a/app/assets/javascripts/settings/ldap-conn-test.js
+++ b/app/assets/javascripts/settings/ldap-conn-test.js
@@ -18,6 +18,61 @@ $(function () {
     $('#ldap_conn_message').html(msg)
   }
 
+  function ldapConnectionTestPart1 () {
+    var file = $('#dex_connector_ldap_certificate').val()
+    if(!file) {
+      ldapConnectionTestPart2( $('#dex_connector_ldap_current_cert').val() )
+      return;
+    }
+    var fileInput = $('#dex_connector_ldap_certificate')[0];
+    var reader = new FileReader();
+
+    reader.onload = function(e) {
+      ldapConnectionTestPart2(reader.result);
+    }
+
+    reader.readAsText(fileInput.files[0]);	
+  }
+  
+  function ldapFormSubmit (event) {
+    var file = $('#dex_connector_ldap_certificate').val()
+    if(!file) {
+      var currentCert = $('#dex_connector_ldap_current_cert').val();
+      if(currentCert) {
+        $('#dex_connector_ldap_certificate').remove();
+      }
+    }
+    return true;
+  }
+
+  function ldapConnectionTestPart2 (cert) {
+    var host     = $('#dex_connector_ldap_host').val()
+    var port     = $('#dex_connector_ldap_port').val()
+    var startTLS = $('#dex_connector_ldap_start_tls_true').parent().hasClass('active')
+    // Flip logic of #anonymous_bind_toggle since expanded div means anonymous binding is 'false'
+    var anonBind = ($('#anonymous_bind_toggle').attr('aria-expanded') === 'false')
+    var dn       = $('#dex_connector_ldap_bind_dn').val()
+    var pass     = $("#dex_connector_ldap_bind_pw[required='required']").val();
+    var baseDN   = $('#dex_connector_ldap_user_base_dn').val()
+    var filter   = $('#dex_connector_ldap_user_filter').val()
+    var paramArray = anonBind ? [host, port, cert, baseDN, filter] : [host, port, cert, dn, pass, baseDN, filter]
+
+    if (noneEmpty(paramArray)) { // Check if form inputs are filled in
+      $.post('/settings/ldap_test', {
+        host: host, port: port, start_tls: startTLS, cert: cert,
+        anon_bind: anonBind, dn: dn, pass: pass, base_dn: baseDN, filter: filter
+      }).done(function (data) {
+        if (data.result.test_pass) renderValidity('Sucessfully validated connection to LDAP server', true)
+        else                       renderValidity('Validation failure. ' + data.result.message, false)
+      }).fail(function (status) {
+        var valMessage = 'Failed to connect to Velum server with following error code/message: ' + status.status + '; ' + status.statusText
+        renderValidity(valMessage, false)
+      })
+    } else {
+      renderValidity('Missing form data', false)
+    }
+  }
+
   $(function () {
     // Elements specific to LDAP Connector page; check for existence before changing
     if ($('#ldap_conn_save').length > 0 && $('#ldap_conn_message').length > 0) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -18,6 +18,56 @@ $(function () {
     $('#ldap_conn_message').html(msg)
   }
 
+  function ldapConnectionTestPart1 () {
+    var file = $('#dex_connector_ldap_certificate').val()
+    if(!file) {
+      ldapConnectionTestPart2( $('#dex_connector_ldap_current_cert').val() )
```
nanoscopic: Note that despite this code getting the current cert contents and using them for testing, "saving" is still broken if a new cert is not chosen via the file upload control. This is a bad user experience and is bound to cause confusion. A bug will be filed for that and a different PR made to address that issue.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -18,6 +18,56 @@ $(function () {
...
+
+    reader.onload = function(e) {
+      ldapConnectionTestPart2(reader.result);
+    }
+
+    reader.readAsText(fileInput.files[0]);	
+  }
+
+  function stripCert (cert) {
+    cert = cert.replace("-----BEGIN CERTIFICATE-----\n",'');
```
nanoscopic: Note that a certificate file could be uploaded without the BEGIN/END blocks. That file would "pass" the connection test button, but would still fail on upload because the backend is expending those to be there. This may be a sub-optimal user experience.

app/assets/javascripts/settings/ldap-conn-test.js
vitoravelino: If the user doesn't select a file, use the current one for testing the connection. If there's no current cert then the test would fail as expected, no? That's right, no? 

What do you mean by "still broken"? Do you mean that there's no way of saving if the test connection fail?

app/assets/javascripts/settings/ldap-conn-test.js
vitoravelino: Can't we validate the params in the controller to force the user to always upload a valid cert file with the BEGIN/END lines? If they are not present, test connection would fail until the user fixes it. 

I thought that this was already being handled since there's a rescue to `OpenSSL::X509::CertificateError`. Do we need to remove the BEGIN/END lines because of `decoded_pem = Base64.decode64(pem_string)` or something else? In the `x509_certificate_validator` we've been using the whole content of the certificate in the constructor method and it works as expected, no? I just want to understand the reason behind the removal of the BEGIN/END lines. Lemme know if I missed something (I probably did :sweat_smile:).

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -18,6 +18,56 @@ $(function () {
...
+    var anonBind = ($('#anonymous_bind_toggle').attr('aria-expanded') === 'false')
+    var dn       = $('#dex_connector_ldap_bind_dn').val()
+    var pass     = $("#dex_connector_ldap_bind_pw[required='required']").val();
+    var baseDN   = $('#dex_connector_ldap_user_base_dn').val()
+    var filter   = $('#dex_connector_ldap_user_filter').val()
+    var paramArray = anonBind ? [host, port, cert, baseDN, filter] : [host, port, cert, dn, pass, baseDN, filter]
+
+    if (noneEmpty(paramArray)) { // Check if form inputs are filled in
+      $.post('/settings/ldap_test', {
+        host: host, port: port, start_tls: startTLS, cert: stripCert(cert),
```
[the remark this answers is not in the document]
vitoravelino: One improvement that could be done in this regard was to instead of declaring a variable for each value, we could declare a single variable as an object and add the values to it. Something like

```js
var params = {
  host: $('#dex_connector_ldap_host').val(),
  port: $('#dex_connector_ldap_port').val(), 
  //...
}
```
This would require a change to the `paramArray` assignment like

```js
if (anonBind) {	
  paramArray = [params.host, params.port, params.cert, params.baseDN, params.filter]	
} else { //... } 
```

I'm suggesting to revert to the if/else block because the line would be too long with the addition of the `params` prefix.

And finally just 

```js
$.post('/settings/ldap_test', params)
```

But this is just cosmetics. Since we don't have eslint to enforce any style, almost everything is fine.

app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -18,6 +18,56 @@ $(function () {
...
+        anon_bind: anonBind, dn: dn, pass: pass, base_dn: baseDN, filter: filter
+      }).done(function (data) {
+        if (data.result.test_pass) renderValidity('Sucessfully validated connection to LDAP server', true)
```
vitoravelino: Could you add the curly braces back to this statement? This is a common sense style in the js community.

app/assets/javascripts/settings/ldap-conn-test.js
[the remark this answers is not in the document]
vitoravelino: The `params` could also be extracted to its own function too. That would reduce the amount of lines of this method. Another possible enhancement. 

And don't worry too much about the codeclimate complains. We don't have an eslint to enforce stuff, so from time to time it's fine to ignore those kind of complains.

app/assets/javascripts/settings/ldap-conn-test.js
vitoravelino: Another improvement that would also increase the readability is to extract the callback to its own function. Like:

```js

// I'm bad at naming sometimes hehe
$.post('/settings/ldap_test', params)
.done(handleSuccessTest)
.fail(handleFailTest)

// ...

function handleSuccessTest(data) {
   // the if/else block here
}
```

That would simplify the method a lot.

app/assets/javascripts/settings/ldap-conn-test.js
nanoscopic: To clarify, the following sequence is a problem:
1. Create an ldap connector ( enter all data, test connection, save )
2. Go to edit the connector ( cert is shown )
3. Change something ( perhaps reconfigure the port )
4. Re-test the connection ( would work; pulling cert from the displayed textarea )
5. Try to save ( you can't; errors out )

That is, this PR allows the 'test connection' button to work in the edit scenario, but the edit scenario still cannot be saved through currently. This PR does not fix that problem.

--

Besides that there is another existing bug with the current code:
1. Enter correct details into an ldap connector
2. Click 'test connection' and succeed; save is enabled
3. Alter the details to be incorrect ( save is not properly then re-disabled )
4. Save the bad data

This scenario also is not corrected by this PR.
Both of the scenarios I am describing are separate bugs that need to be resolved.
This PR is only meant to address the fact that even the 'happy path' for ldap external authentication is currently broken in HEAD.

app/assets/javascripts/settings/ldap-conn-test.js
nanoscopic: The backend already does enforce that the BEGIN/END blocks are there. All I am saying is that the test connection button currently does not enforce that they are there, it simply removing them before testing the connection as the backend test connection code doesn't work with them present.

The backend of the 'test connection' feature could be changed to remove them at that point. This PR is meant to be as quick as possible to fix the broken head state with minimal change. It serves that purpose and should be accepted and merged. If further changes are desired that is fine but this PR should not be held up for that reason.

app/assets/javascripts/settings/ldap-conn-test.js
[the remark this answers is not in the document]
nanoscopic: The function as written is designed to give an error if some field is not filled in. If that is viewed as the goal it does it in an efficient way with a relatively low number of lines of code. It is quite clear how it works. I would not have written it this way myself; the original code here was written by Daniel Orf, but I do think it is accomplishing the purpose in an acceptable fashion.

I don't think the UX flow is good. What really should happen is that fields are marked as required ( typically with a red star ), and that when you attempt to save without any of the fields present, that the fields that are empty are set with a red border on the entire field and that the scroll automatically scrolls up to the first of the fields that has not been entered. That is the common pattern that should be followed for a good user experience.

That needs to happen in the near future. When that happens this code you are suggesting to optimize would go away because it will be deleted and replaced with something else entirely. For this reason optimizing what we have here is premature and should not be done and is a waste of developer time.

app/assets/javascripts/settings/ldap-conn-test.js
nanoscopic: Vitor: Whether braces are used in this sense is, I believe, a personal preference thing. As the author of the PR I have chosen what I prefer and looks good to me.

If SUSE as a company, or CaaSP as a team has a defined code styleguide that I can see and is told to everyone, and everyone agrees to follow, then please point it out and I would be happy to follow it.

Pull requests are not the place to have discussions of personal code styling. You should be checking functionality to determine whether you approve, and possibly blocking if the appearance of the code is such that it is unclear or confusing.

It is not confusing with or without the braces. I will not add them simply because you like it better. That conversation is not one that belongs here.

Additionally, the code is perfectly readable with the 2 lines of "good/fail" within anonymous functions. Moving those out into separate functions does not, in my opinion, increase readability. I additionally will not move those out simply because you want it. There needs to a definitive value to doing so beyond your feelings about it. Present a logical argument for doing so and I will be happy to.

Convince the team as a whole to follow some definitive style and I will be happy to abide by it.

app/assets/javascripts/settings/ldap-conn-test.js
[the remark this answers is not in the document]
danielorf: I agree that denoting required fields with a red star/border should be added in a separate bug.

app/assets/javascripts/settings/ldap-conn-test.js
nanoscopic: The latest updates to this PR fix the first scenario. Saving of an existing connector after changes without re-uploading the cert now works.

app/assets/javascripts/settings/ldap-conn-test.js
nanoscopic: The BEGIN/END removal has now been moved to the backend to prevent the possible issues described.

PR #652, opened 2018-09-08
Prevent changed external ldap values from skipping checks

nanoscopic:
Alter external ldap auth form to redisable save and rerequire 'check connection' when values are changed.

### DIFF
--- a/app/assets/javascripts/settings/ldap-conn-test.js
+++ b/app/assets/javascripts/settings/ldap-conn-test.js
@@ -33,7 +33,7 @@ $(function () {
 
     reader.readAsText(fileInput.files[0]);	
   }
-  
+
   function ldapFormSubmit (event) {
     var file = $('#dex_connector_ldap_certificate').val()
     if(!file) {
@@ -73,12 +73,27 @@ $(function () {
     }
   }
 
+  function setupFieldOnchange ( fieldName ) {
+    var fetch;
+    if( fieldName == 'bind_pw' ) fetch = "#dex_connector_ldap_bind_pw[required='required']";
+    else fetch = '#dex_connector_ldap_' + fieldName;
+
+    var field = $(fetch);
+    field.change( fieldChange );
+  }
+
+  function fieldChange () {
+    showTestConnectionMessage();
+  }
+
+  function showTestConnectionMessage () {
+    renderValidity( 'Test Connection first before saving', false );
+  }
+
   $(function () {
     // Elements specific to LDAP Connector page; check for existence before changing
     if ($('#ldap_conn_save').length > 0 && $('#ldap_conn_message').length > 0) {
-      $('#ldap_conn_save').prop('disabled', 'disabled')
-      $('#ldap_conn_message').addClass('ldap-conn-message-default')
-      $('#ldap_conn_message').html('Test Connection first before saving')
+      showTestConnectionMessage();
     };
 
     // Run LDAP Connection Test with button click
@@ -90,5 +105,13 @@ $(function () {
       var form = $('form.dex-connectors-form');
       form.submit(ldapFormSubmit);
     }
+
+    // prepend with dex_connector_ldap_
+    var fields = [
+      'name', 'host', 'port', 'start_tls_true', 'start_tls_false', 'certificate',
+      'bind_anon_true', 'bind_anon_false', 'bind_dn', 'bind_pw',
+      'user_base_dn', 'user_filter'
+    ];
+    for( var i=0;i<fields.length;i++ ) setupFieldOnchange( fields[i] );
   })
 })
--- a/spec/features/settings/dex_connector_ldap_feature_spec.rb
+++ b/spec/features/settings/dex_connector_ldap_feature_spec.rb
@@ -130,6 +130,75 @@ describe "Feature: LDAP connector settings", js: true do
     end
   end
 
+  # rubocop:disable RSpec/ExpectInHook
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/settings/ldap-conn-test.js
```diff
@@ -73,12 +73,30 @@ $(function () {
...
+    var field = $(fetch);
+    field.change( fieldChange );
+  }
+
+  function fieldChange () {
+    showTestConnectionMessage();
+  }
+
+  function showTestConnectionMessage () {
+    /*$('#ldap_conn_save').prop('disabled', 'disabled')
```
danielorf: Commented code could be removed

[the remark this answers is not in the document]
nanoscopic: Commented code is removed in latest update to this PR.

spec/features/settings/dex_connector_ldap_feature_spec.rb
```diff
@@ -130,6 +130,75 @@
...
+      # choose id: "dex_connector_ldap_start_tls_true"
+      page.execute_script("$('#dex_connector_ldap_start_tls_true').prop('checked', true)")
+      page.execute_script("$('#dex_connector_ldap_start_tls_true').trigger('change')")
+    end
+
+    it "allows change of bind_anon" do
+      # Is false by default - change to true
+      # Following line should work but it does not so using JS
+      # choose id: "dex_connector_ldap_bind_anon_true"
+      page.execute_script("$('#dex_connector_ldap_bind_anon_true').prop('checked', true)")
```
vitoravelino: You can add an id/class to the label element and use `find` and `click` to mimic the user behavior instead of changing the value directly through `execute_script`. By doing that you are bypassing the normal behavior of the engine that might lead into unexpected results like not triggering an event.

spec/features/settings/dex_connector_ldap_feature_spec.rb
```diff
@@ -130,6 +130,75 @@
...
+      page.execute_script("$('#dex_connector_ldap_user_base_dn').trigger('change')")
+    end
+
+    it "allows change of user filter" do
+      fill_in id: "dex_connector_ldap_user_filter", with: "AAA"
+      page.execute_script("$('#dex_connector_ldap_user_filter').trigger('change')")
+    end
+
+    after do
+      expect(page.evaluate_script("$('#ldap_conn_save').prop('disabled')")).to eq(true)
```
vitoravelino: You can use the built-in matcher `have_button("Save", disabled: true)` instead of using `evaluate_script` to check this kind of thing. Always prefer the built-in matchers in favor of more maintainable code.

spec/features/settings/dex_connector_ldap_feature_spec.rb
```diff
@@ -130,6 +130,75 @@
...
+      page.execute_script("$('#dex_connector_ldap_certificate').trigger('change')")
+    end
+
+    it "allows change of bind dn" do
+      fill_in id: "dex_connector_ldap_bind_dn", with: "AAA"
+      page.execute_script("$('#dex_connector_ldap_bind_dn').trigger('change')")
+    end
+
+    it "allows change of password" do
+      page.execute_script("$('#dex_connector_ldap_bind_pw[required=\"required\"]').val('AAA')")
```
vitoravelino: You can use the `fill_in` matcher here using `dex_connector_ldap_bind_pw` as the locator instead of `execute_script`. 

Even having duplicated elements with the same id, `fill_in` will try to find the visible elements and will ignore the duplicated one. This will also benefit the normal engine behavior when filling the input text to avoid unexpected results.

vitoravelino:
From what I could investigate so far, the usage `trigger('change')` was necessary because of `execute_script`. So, instead of using that we can try to use the built-in matchers to avoid unexpected results and forcing us to manually trigger an event.

spec/features/settings/dex_connector_ldap_feature_spec.rb
nanoscopic: Since fill_in does not properly trigger onchange events, which are needed for this test, the exact same code you want to remove using jquery to grab the element still has to exist on the next line. Making a change like this as a result is rather pointless. I changed it and have tested it and will push the update, but I don't think this is a good use of my time, anyone else's time, nor of the test servers time retesting this for an inconsequential change.

"Avoid unexpected results"? What are you talking about? We intend to stop using Ruby and Rails and Capybara in the future entirely so that is an irrelevant optimization.

spec/features/settings/dex_connector_ldap_feature_spec.rb
nanoscopic: More maintainable is an opinion here very much so. More maintainable from whose perspective? From my perspective and anyone who isn't intimately familiar with Capybara the JQuery is actually more readable and easier to maintain.

vitoravelino:
I understand the urgency of getting this merged to unleash the LDAP feature, I understand the lack of convention rule and all the other minor things. I'll start a discussion around having these conventions to lower the contribution bar.

In the meantime, I tried to perform the remaining changes I proposed and I got the tests to pass. Here's the [link](https://github.com/kubic-project/velum/pull/656) to the PR with my changes. Could you test it locally to see if they are not passing on your machine?

I'm bringing this up because I'm concerned we are making things less reliable by exaggerating on the usage of js things inside of capybara tests and lack of proper external services mocking in order to have tests green.

PR #654, opened 2018-09-11
[3.0] enable update channels

MaximilianMeister:
feature#update-channels

### Depends on

https://github.com/kubic-project/salt/pull/658
https://github.com/thkukuk/update-checker

## Backport of

https://github.com/kubic-project/velum/pull/638

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -233,7 +233,8 @@ MinionPoller = {
 
         handleBootstrapErrors();
         handleUnsupportedClusterConfiguration();
-        showRepoMirrorAlert(data.admin);
+        // FIXME: uncomment when v4 gets released
+        //showRepoMirrorAlert(data.admin);
 
         // show/hide "update all nodes" link
         var hasAdminNodeUpdate = data.admin.tx_update_reboot_needed || data.admin.tx_update_failed;
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -54,7 +54,12 @@ MinionPoller = {
         var pendingRendered = "";
         var masterApplied = false;
         var updateAvailable = false;
+        var migrationAvailable = false;
         var updateAvailableNodeCount = 0;
+        var migrationAvailableNodeCount = 0;
+
+        // reset admin migrated
+        State.adminMigrated = adminMigrated(data);
 
         // reset pending
         State.hasPendingStateNode = false;
@@ -156,6 +161,7 @@ MinionPoller = {
         });
 
         State.hasPendingStateNode = !!pendingStateMinion;
+        State.hasPendingStateAdmin = data.admin.highstate == "pending";
 
         // find if there's a node pending removal
         // to set State.pendingRemovalMinionId
@@ -183,6 +189,11 @@ MinionPoller = {
             updateAvailableNodeCount++;
           }
 
+          if (minions[i].tx_update_migration_available) {
+            migrationAvailable = true;
+            migrationAvailableNodeCount++;
+          }
+
           // removes node from the pending acceptance state in the browser
           removePendingAcceptance(minions[i].minion_id);
         }
@@ -193,6 +204,10 @@ MinionPoller = {
         updateAvailable = updateAvailable &&
                           !State.hasPendingStateNode &&
                           !State.pendingRemovalMinionId;
+
+        migrationAvailable = migrationAvailable &&
+                             !State.hasPendingStateNode &&
+                             !State.pendingRemovalMinionId;
 
         // Build Pending Nodes display table
         if (pendingMinions.length) {
@@ -212,20 +227,32 @@ MinionPoller = {
         $('.discovery-nodes-panel').toggleClass('hide', allMinions.length === 0);
         $('.discovery-empty-panel').toggleClass('hide', allMinions.length > 0);
 
-        MinionPoller.handleAdminUpdate(data.admin || {});
+        MinionPoller.handleAdminUpdate(data || {});
+        MinionPoller.handleAdminMigration(data || {});
         MinionPoller.handleRetryableOrchestrations(data);
 
         handleBootstrapErrors();
         handleUnsupportedClusterConfiguration();
+        showRepoMirrorAlert(data.admin);
 
         // show/hide "update all nodes" link
         var hasAdminNodeUpdate = data.admin.tx_update_reboot_needed || data.admin.tx_update_failed;
-        State.updateAllNodes = updateAvailable && !hasAdminNodeUpdate;
+        State.updateAllNodes = updateAvailable && !hasAdminNodeUpdate && !adminMigrated(data);
         $("#update-all-nodes").toggleClass('hidden', !State.updateAllNodes);
+
+        // show/hide "migrate all nodes" link
+        var hasAdminNodeMigration = data.admin.tx_update_migration_available || data.admin.tx_update_failed;
+        var nodeIsPending = State.minions.filter(function (m) { return m.highstate === "pending" }).length > 0;
+        State.migrateAllNodes = migrationAvailable && !hasAdminNodeMigration && !nodeIsPending;
+        $("#migrate-all-nodes").toggleClass('hidden', !State.migrateAllNodes);
 
         MinionPoller.enable_kubeconfig(masterApplied);
 
-        $('#out_dated_nodes').text(updateAvailableNodeCount + ' ');
+        if (updateAvailable) {
+          $('#out_dated_nodes').text(updateAvailableNodeCount + ' ');
+        } else if (migrationAvailable) {
+          $('#out_dated_nodes').text(migrationAvailableNodeCount + ' ');
+        }
 
         $('.assigned-count').text(minions.length);
         $('.master-count').text(MinionPoller.selectedMasters.length);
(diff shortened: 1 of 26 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -3,6 +3,7 @@
 # DashboardController shows the main page.
 class DashboardController < ApplicationController
   include Discovery
+  before_action :update_grains
```
ereslibre: Do we really need to do this? It will impact all requests done to this controller, that now will have to wait for the request to the salt-api.

app/controllers/orchestrations/migration_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    redirect_to root_path
+  end
+
+  def check_mirror
+    Velum::Salt.call(
+      action:      "cmd.run",
+      arg:         "update-checker-migration",
+      targets:     "roles:(admin|kube-(master|minion))",
+      target_type: "grain_pcre"
+    )
```
ereslibre: How long can this process take? Might it timeout the http request?

I wish we had something to background tasks properly :(

app/models/minion.rb
```diff
@@ -6,6 +6,8 @@ class Minion < ApplicationRecord
   scope :cluster_role, -> { where role: [Minion.roles[:master], Minion.roles[:worker]] }
   scope :unassigned_role, -> { where role: nil }
   scope :needs_update, -> { where "tx_update_reboot_needed = true or tx_update_failed = true" }
+  scope :has_migration, -> { where "tx_update_migration_available = true" }
+  scope :needs_mirror_sync, -> { where "tx_update_migration_mirror_synced = false" }
```
ereslibre: Those can be rewritten directly as:

```ruby
scope :has_migration, -> { where tx_update_migration_available: true }
scope :needs_mirror_sync, -> { where tx_update_migration_mirror_synced: false }
```

app/models/orchestration.rb
```diff
@@ -6,14 +6,14 @@ class Orchestration < ApplicationRecord
   class OrchestrationAlreadyRan < StandardError; end
   class OrchestrationOngoing < StandardError; end
 
-  enum kind: [:bootstrap, :upgrade, :removal, :force_removal]
+  enum kind: [:bootstrap, :upgrade, :migration, :removal, :force_removal]
```
ereslibre: Adding `migration` here will confuse rails, because it automatically assigned `bootstrap = 0`, `upgrade = 1`, `removal = 2`, `force_removal = 3`. Either we do add it at the end (`[:bootstrap, :upgrade, :removal, :force_removal, :migration]`), or we convert this array into a hash and do something like:

```ruby
enum kind: { bootstrap: 0, upgrade: 1, migration: 4, removal: 2, force_removal: 3 }
```

Otherwise if we keep this as is, old `removal` orchestrations saved in the database (as `2`) will be identified as `migration`, and old `force_removal` orchestrations saved with (as `3`) will be identified as `removal`.

db/migrate/20181715075511_add_migration_status.rb
```diff
@@ -0,0 +1,8 @@
+class AddMigrationStatus < ActiveRecord::Migration
+  def change
+    add_column :minions, :tx_update_migration_available, :boolean, default: false, after: :tx_update_failed
+    add_column :minions, :tx_update_migration_notes, :string, after: :tx_update_migration_available
```
ereslibre: Depending on the size of the notes we'd need to make this field bigger.

app/controllers/orchestrations/migration_controller.rb
MaximilianMeister: this is really quick, i dont think it should make the request timeout

app/controllers/dashboard_controller.rb
MaximilianMeister: probably not, that's why i have added it as a separate commit. wanted to get some feedback. yes the idea was to always show the latest grains, hence putting it in the application controller. but if that's too invasive, lets drop it again

db/migrate/20181715075511_add_migration_status.rb
MaximilianMeister: it's just a url

MaximilianMeister: however if we would have a way to include the full release notes as text, we might be able to display the notes directly in velum, for air-gapped systems. but for now it's just a link... not sure how this will evolve

app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -607,6 +750,54 @@ $('body').on('click', '.reboot-update-btn', function(e) {
   });
 });
 
+// migrate admin node handler
+$('body').on('click', '.trigger-migrate-admin-btn', function(e) {
+  var $btn = $(this);
+  var $modal = $('.migrate-admin-modal');
+
+  // should any node have an out of sync mirror, open the sync modal
+  if(showMirrorSyncModal() > 0) {
```
vitoravelino: This is preventing `spec/features/admin_node_migration_feature_spec.rb:44` to pass (with the changes we discussed on IRC) as you can see in the screenshot below. I suggest you to also add an expectation to `Migrating` text to indicate that the migration actually started and didn't get stuck in this state.

error_state-1537983578

vitoravelino: This is also another good case to have its own test.

Another good scenario would be the disappearance of this RMT/SMT out of sync alert when user clicks on `Confirm`. Locally during my tests it didn't disappear but I didn't investigate why.

app/assets/stylesheets/pages/nodes_list.scss
```diff
@@ -128,10 +138,15 @@
     .failed-message {
       display: inline-block;
     }
+
+    .actions .btn-link {
```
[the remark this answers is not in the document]
MaximilianMeister: > There's already `visibility: hidden` in line 108, there's no need to override it in the failed state. Is this a new condition?

i think the issue was because in case of a failed admin migration, the migrate admin button still showed up in the error message. i couldnt find any js where we remove it in the failed update case, so i found this css code. i was not really knowing what i am doing, just try and error... do you have an idea why the button is visible in the failed admin migration case?

app/assets/javascripts/dashboard/dashboard.js
MaximilianMeister: how did you get to the screenshot ? :D

MaximilianMeister:
@vitoravelino removed the css bit again, but i need to test the failure case again, i suspect the btn link will still appear in the failure message. i also added some spec for the mirror sync check, and i managed to get the other test green, by simply setting all nodes to `tx_update_mirror_synced: true` not only the admin... thanks a lot for pinning down this issues!

app/assets/stylesheets/pages/nodes_list.scss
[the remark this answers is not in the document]
vitoravelino: The button link will appear if `data.admin.tx_update_migration_available` is `true` or if `data.admin.tx_update_reboot_needed` is true even having both cases failed.

So the requirement is that in case of a failed update/migration, the user won't be able to click on `Update/Migrate admin node` anymore?

app/assets/javascripts/dashboard/dashboard.js
vitoravelino: Copied the `after` hook from `automation`.

```ruby
config.after do |example|
  if example.exception
    save_screenshot("screenshots/error_state-#{Time.now.to_i}.png", full: true)
  end
end
```

The only trick was that it had to be declared after the Capybara declarations, so I had to add this to the `capybara.rb` file.

vitoravelino:
I've pulled the latest changes and what I noticed is that the admin node migration doesn't seem to be happening anymore. The admin node is not rebooting. So I'm able to refresh the page and click on the link again and it fails because there's already an orchestration on going (had to check the devtools). I followed the same instructions as before. Anything changed? Is the no reboot now expected?

MaximilianMeister:
> Anything changed? Is the no reboot now expected?

nothing changed, it looks like your salt orch failed? then usually the event processor should set the orchestration to failed, not sure why that didnt happen in your case. here are the instructions again:

```
# register
docker exec -it $(docker ps | grep "salt-master_velum" | awk '{print $1}') salt "*" cmd.run "SUSEConnect -r [CODE]"

# install 3.0 maintenance updates
docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt -P "roles:(admin|kube-(master|minion))" cmd.run '/usr/sbin/transactional-update cleanup dup salt'
docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt '*' saltutil.refresh_grains
docker exec -it $(docker ps | grep "velum-dashboard" | awk '{print $1}') entrypoint.sh bash -c "export RAILS_ENV=production; bundle exec rails runner 'Minion.update_grains'"
# reboot via velum

## add update-checker-migration
# on the host, download these rpm's and copy them to the nodes
for i in 1.0 2.0 3.0; do scp ~/Downloads/update-checker-1.0+git20180905.814bdd8-1.1.noarch.rpm root@10.17.$i:/root/; done

# on each node install the packages and reboot
transactional-update cleanup pkg install --oldpackage /root/*.rpm
# choose (2) break perl dependency, which is not used by update-checker-migration
reboot

# adapt the update-checker conf to write grains
docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt -P "roles:(admin|kube-(master|minion))" cmd.run "sed -i -e 's|output=.*|output=salt|g' /etc/update-checker.conf"
# write the grains
docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt -P "roles:(admin|kube-(master|minion))" cmd.run "update-checker-migration"
docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt '*' saltutil.refresh_grains
docker exec -it $(docker ps | grep "velum-dashboard" | awk '{print $1}') entrypoint.sh bash -c "export RAILS_ENV=production; bundle exec rails runner 'Minion.update_grains'"
```

MaximilianMeister:
added another fix: when user reloads the page during migration, there was a short timeframe, where `tx_update_reboot_needed` was set, and therefore the `UPDATE ADMIN NODE` link popped up

also got rid of some DRY in the notification selectors by putting the notification cleanup in a reusable function

MaximilianMeister:
rebased

db/migrate/20181715075511_add_migration_status.rb
ereslibre: By default `string` will map to `VARCHAR(255)`, I guess we should be fine?, we might want to tweak`limit` on the `add_column` call though so we are around something bigger (just in case the URL's grow at some point)

jordimassaguerpla:
Can we have an approval on this one?

jordimassaguerpla:
Hi. I am sorry but we are removing the release-3.0 branch. Due to change of priorities, I don't think this fix will be necessary for 3.0 so I am closing this PR.

PR #659, opened 2018-09-17
[3.0][bsc#1108679] use online status during update for update status

MaximilianMeister:
we can make use of the online status to have a more accurate status
during the update

if a node is offline during the update, it likely means that it's
rebooting

(cherry picked from commit 23bb090b5a8bd74a3f3af935701356d9aacf4276)

Backport of https://github.com/kubic-project/velum/pull/641


jordimassaguerpla:
Hi Max,
I would need a bsc or fate number.
If there is none, feel free to create a bug in bugzilla.suse.com and add the reference into your commit.

MaximilianMeister:
> Hi Max,
> I would need a bsc or fate number.
> If there is none, feel free to create a bug in bugzilla.suse.com and add the reference into your commit.

@jordimassaguerpla added -> https://bugzilla.suse.com/show_bug.cgi?id=1108679

PR #660, opened 2018-09-18
[bsc#1108740] dont allow updating admin again when cluster update failed

MaximilianMeister:
bsc#1108740

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -288,8 +288,11 @@ MinionPoller = {
 
   handleAdminUpdate: function(admin) {
     var $notification = $('.admin-outdated-notification');
+    var failedToUpdate = State.minions.filter(function (m) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/dashboard/dashboard.js
```diff
@@ -289,7 +289,7 @@ MinionPoller = {
   handleAdminUpdate: function(admin) {
     var $notification = $('.admin-outdated-notification');
 
-    if (State.hasPendingStateNode || State.pendingRemovalMinionId) {
+    if (State.hasPendingStateNode || State.pendingRemovalMinionId || failedToUpdateCount > 0) {
```
vitoravelino: `failedToUpdateCount` is a function and not a variable. You forgot to call it instead of just referring it. I would also move the comparison to the function and rename the function to just `failedToUpdate`.

MaximilianMeister: oops. updated, thanks :smile:

PR #665, opened 2018-09-25
OIDC Connector support

dannysauer:
This set of changes implements external OIDC connector support within Velum.  It's based upon the LDAP connector, and should incorporate the majority of the relevant lessons learned from that implementation.

### DIFF
--- a/app/controllers/settings/dex_connector_oidcs_controller.rb
+++ b/app/controllers/settings/dex_connector_oidcs_controller.rb
@@ -1,0 +1,98 @@
+# Settings::DexConnectorOidcsController is responsible to manage requests
+# related to OIDC connectors.
+class Settings::DexConnectorOidcsController < SettingsController
+  # populate @data_holder with existing database record
+  before_action :set_data_holder, except: [:index, :new, :create]
+
+  def index
+    @oidc_connectors = DexConnectorOidc.all
+  end
+
+  def new
+    @is_data_valid = false
+    @data_holder = data_holder_type.new
+  end
+
+  def edit
+    @is_data_valid = false
+  end
+
+  def create
+    @data_holder = data_holder_type.new(data_holder_params)
+
+    if params[:validate]
+      @is_data_valid = @data_holder.valid?
+      if @is_data_valid
+        flash.now[:notice] = "#{friendly_name} is valid"
+      else
+        flash.now[:alert]  = "#{friendly_name} is invalid"
+      end
+      render action: :new
+    else
+      @data_holder.save!
+      redirect_to settings_dex_connector_oidcs_path,
+                  notice: "#{friendly_name} was successfully created."
+    end
+  rescue ActiveRecord::RecordInvalid
+    render action: :new, status: :unprocessable_entity
+  end
+
+  def destroy
+    @data_holder.destroy!
+
+    redirect_to settings_dex_connector_oidcs_path,
+                notice: "OIDC Connector was successfully removed."
+  end
+
+  def update
+    if params[:validate]
+      # merge form values with DB record so form redisplays properly
+      data_holder_update_params.each do |key, value|
+        @data_holder[key] = value
+      end
+      @is_data_valid = @data_holder.valid?
+      if @is_data_valid
+        flash.now[:notice] = "#{friendly_name} is valid"
+      else
+        flash.now[:alert]  = "#{friendly_name} is invalid"
+      end
+      render action: :edit
+    else
+      @data_holder.update_attributes!(data_holder_update_params)
+      redirect_to [:settings, @data_holder],
+                  notice: "#{friendly_name} was successfully updated."
+    end
+  rescue ActiveRecord::RecordInvalid
+    render action: :edit, status: :unprocessable_entity
+  end
+
+  protected
+
+  def data_holder_type
+    DexConnectorOidc
+  end
+
+  def data_holder_params
+    oidc_connector_params
+  end
+
+  def data_holder_update_params
+    oidc_connector_params
+  end
+
+  def friendly_name
+    "OIDC Connector"
+  end
+
+  private
+
+  def oidc_connector_params
+    params.require(:dex_connector_oidc).permit(
+      :name, :provider_url, :client_id, :client_secret, :callback_url, :basic_auth
+    )
+  end
+
+  def set_data_holder
+    @data_holder = data_holder_type.find(params[:id])
+  end
+end
--- a/app/views/settings/dex_connector_oidcs/_form.html.slim
+++ b/app/views/settings/dex_connector_oidcs/_form.html.slim
@@ -1,0 +1,70 @@
+= form_for [:settings, @data_holder], html: { class: "dex-connectors-form", autocomplete: "false" } do |f|
+  / Chrome may autocomplete username into text field before a password field, regardless of name
+  / This *should* prevent that (in combination with 'autocomplete=false' on the form_for; "off" no longer works)
+  / Explanation: Chrome looks for the first password field, while newer versions prefer one named "password"
+  / Firefox behaves similarly.  Hopefully they don't figure out the hidden workaround soon...
+  = f.text_field     :username, style: "display:none", value:"autocomplete trap"
+  = f.password_field :password, style: "display:none", value:"autocomplete bait"
+
+
+  .form-group class="#{error_class_for(@data_holder, :name)}"
+    = f.label :name
+    = f.text_field :name, id: "oidc_name", class: "form-control", value: @data_holder.name, required: true
+    = error_messages_for(@data_holder, :name)
+    small.form-text.text-muted
+      | Name shown to user when selecting a connector
+
+  h4 Provider
+
+  .form-group class="#{error_class_for(@data_holder, :provider_url)}"
+    = f.label :provider_url, "Provider URL"
+    = f.text_field :provider_url, id: "oidc_provider", class: "form-control", value: @data_holder.provider_url, required: true
+    = error_messages_for(@data_holder, :provider_url)
+    small.form-text.text-muted
+      | Issuer used by OIDC Provider
+    
+  .form-group class="#{error_class_for(@data_holder, :basic_auth)}"
+    = f.label :basic_auth, "Negotiate client auth"
+    br
+    .btn-group.btn-group-toggle data-toggle="buttons"
+      = f.label :basic_auth, nil, class: "btn btn-default #{'btn-primary active' if @data_holder.basic_auth}"
+        = f.radio_button :basic_auth, "true", checked: @data_holder.basic_auth
+        | Enable
+      = f.label :basic_auth, nil, class: "btn btn-default #{'btn-primary active' unless @data_holder.basic_auth}"
+        = f.radio_button :basic_auth, "false", checked: !@data_holder.basic_auth
+        | Disable
+    = error_messages_for(@data_holder, :basic_auth)
+    br
+    small.form-text.text-muted
+      | Set to "Disable" only if this provider requires client credentials to be sent using POST but doesn't publish that requirement in discovery response (this is uncommon)
+    
+  h4 Client
+  
+  / The 'https' and '3200' seem to be hard-coded elsewhere, so I assume they're not discoverable :/
+  - external_api_url = "https://#{Pillar.value(pillar: :apiserver)}:32000/callback"
+  .form-group class="#{error_class_for(@data_holder, :callback_url)}"
+    = f.label :callback_url, "Callback URL:"
+    span.show
+      = external_api_url
+    small.form-text.text-muted
+      | URL which will be provided to OIDC provider as the callback location
+    = f.hidden_field :callback_url, id: "oidc_callback_url", class: "form-control", value: external_api_url, readonly: true
+
+  .form-group class="#{error_class_for(@data_holder, :client_id)}"
+    = f.label :client_id, "ID"
+    = f.text_field :client_id, id: "oidc_client_id", class: "form-control", value: @data_holder.client_id, required: true
+    = error_messages_for(@data_holder, :client_id)
+    small.form-text.text-muted
+      | Client ID provided by OIDC Provider
+    
+  .form-group class="#{error_class_for(@data_holder, :client_secret)}"
+    = f.label :client_secret, "Secret"
+    = f.password_field :client_secret, id: "oidc_client_secret", class: "form-control", value: @data_holder.client_secret, required: true
+    = error_messages_for(@data_holder, :client_secret)
+    small.form-text.text-muted
+      | Secret associated with Relying Party's Client ID
+
+  .form-actions.clearfix
+    = f.submit "Save", class: "btn btn-primary action", id: "oidc_conn_save", name: "submit", disabled: !@is_data_valid
+    = link_to "Cancel", settings_dex_connector_oidcs_path, class: "btn btn-default action"
+    = f.submit "Test Connection", class: "btn btn-primary", id: "oidc_conn_validate", name: "validate"
--- a/docs/oidc_testing/README.md
+++ b/docs/oidc_testing/README.md
@@ -1,0 +1,66 @@
+# Bring up a cluster
+1. Set up dev environment
+   ```
+   git clone git@github.com:kubic-project/automation
+   cd automation
+   ./caasp-devenv --setup
+   ```
+1. `cd ../automation && ./caasp-devenv --build --bootstrap` (with whatever other options, like `-L provo`)
+
+# Bring up the stand-alone dex OIDC Provider (OP)
+1. `cd ../velum/docs/oidc_testing`
+1. `cp examples/config-dev.yaml{.example,}`
+1. edit `examples/config-dev.yaml` as needed, paying special attention to the two places marked `NOTE`:
+   * set the `issuer:` URL to match the DNS host name of your development machine.  Note that this hostname must resolve _via DNS_ from the master and worker nodes running the dex_dex container.  You can check this via: `docker exec _container_id_ ping -c 1 _host_name_`.  You should get a response back, and it should show the IP address of the host.  The URL used as the issuer _must_ be exactly the URL you use in the OIDC provider URL field in Vellum".
+   * If you do not have a DNS-resolvable name, there are a variety of ways to fix that.  The easiest is to add an entry to the libvirt DNS resolver, which will make the hostname/IP mapping work from the VMs and containers within, but won't affect anything outside the cluster.  Say you're using `dexhost.do.main` as your issuer name, and it's listening on `192.168.0.213`.  Do this, and then retry the ping test above:
+     ```
+     virsh net-update caasp-dev-net add dns-host '<host ip="192.168.0.213"><hostname>dexhost.do.main</hostname></host>' --live --config
+     ```
+     Note that you can add multiple hostnames within the host, should that be necessary.
+1. still in the directory where docker-compose.yml is, run `docker-compose up -d`
+   * output should look roughly like this:
+     ```
+     sauer@lightning:~/dev/dex$ docker-compose up -d
+     Starting dex_base_1 ... done
+     Starting dex_dex_1  ... done
+     sauer@lightning:~/dev/dex$ 
+     ```
+   * logs can be accessed using `docker-compose logs` in the same directory; append `-f` to equivocate `tail -f` on the log
+1. it should be possible to use WebFinger to fetch the important data from Dex now, using `curl http://dexhost.do.main:5556/dex/.well-known/openid-configuration` on the admin node.  You should get back a JSON document listing the expected issuer, some endpoints, supported scopes, etc.  Note specifically that if the issuer URL is broken (`http:/host:5556/dex`, for example -- note the wrong number of slashes), dex may start up without emitting a warning, but the curl check will return a 404. Do not skip this test. :)
+
+# Configure and test the OIDC provider in velum
+1. log in to admin.devenv.caasp.suse.net
+1. under settings, click `new oidc provider`
+1. put an "x" in all the fields, and hit save (these values are currently not saved)
+1. click on the "edit" button next to the newly-created "oidc_provider_1"
+1. fill in the values correctly
+   * name is arbitrary (will be the string the user generally sees)
+   * Endpoint URL is the is the same as the issuer from the config file
+   * callback URL should be Kubernetes callback URL: `https://kube-api-x1.devenv.caasp.suse.net:32000/callback` (in config file's static client list)
+   * leave basic auth on (turning it off is for rare non-compliant OPs)
+   * client id is `example-app` (from the static clients in config file)
+   * client secret is `ZXhhbXBsZS1hcHAtc2VjcmV0` (from the static clients in config file)
+1. save and apply changes
+1. wait for orchestration to complete
+1. Go to https://admin.devenv.caasp.suse.net/oidc/
+1. Select the "Log in with X" button which matches the name you provided in the setup
+1. Select either option
+   * Log in with email: use the email and password from the example config (defaults are admin@example.com/password)
+   * Log in with Example: use the "mock" provider in dex, which doesn't prompt for anything, just returns success
+1. You should be redirected to Velum and be presented with a kubeconfig for the user you selected in the previous step
+1. You should also see a log message logged (use `docker-compose logs`); success with the mock option looks like this:
+   `dex_1   | time="2018-08-24T19:15:15Z" level=info msg="login successful: connector \"mock\", username=\"Kilgore Trout\", email=\"kilgore@kilgore.trout\", groups=[\"authors\"]"`
+
+# Clean up testing environment
+1. Tear down the dex container
+   * In the velum directory containing the docker-compose.yml: `docker-compose down`
+   * should look like this:
+     ```
+     sauer@lightning:~/dev/dex$ docker-compose down
+     Stopping dex_dex_1 ... done
+     Removing dex_base_1 ... done
+     Removing dex_dex_1  ... done
+     Removing network dex_default
+     ```
+1. Tear down the cluster
+   * In the automation directory:`./caasp-devenv --destroy`
--- a/docs/oidc_testing/docker-compose.yml
+++ b/docs/oidc_testing/docker-compose.yml
@@ -1,0 +1,19 @@
+# simple test dex instance
+version: "2.1"
+
+services:
+  base:
+    image: quay.io/coreos/dex:v2.10.0
+
+  dex:
+    extends:
+      service: base
+    volumes:
+      - "./examples/:/go/src/github.com/coreos/dex/examples"
+    # We're using the example developer config with a static entry
+    command: ["serve","/go/src/github.com/coreos/dex/examples/config-dev.yaml"]
+    ports:
+      - 5554:5554 # SSL port, if enabled
+      - 5556:5556 # non-SSL port (http://yourhost:5556/dex)
+      - 5558:5558 # metrics, prometheus style - http://yourhost:5558/metrics
+
--- a/docs/oidc_testing/examples/.gitignore
+++ b/docs/oidc_testing/examples/.gitignore
@@ -1,0 +1,2 @@
+config-dev.yaml
+dex.db
--- a/docs/oidc_testing/examples/config-dev.yaml.example
+++ b/docs/oidc_testing/examples/config-dev.yaml.example
@@ -1,0 +1,79 @@
+################################################################################
+# This is the test Dex config for verifying external OIDC connector support
+#
+# This config will give two options when trying to auth via OIDC:
+# * "mock", which always sends back a fake name ("Kilgore Trout") without any
+# further authentication. (this is hard-coded in the mock callback)
+# * "static", which uses an email (admin@example.com) as the username and a
+# password of "password", then returns a username of "admin"
+################################################################################
+
+################################################################################
+# you need to change this value
+################################################################################
+# The base path of dex and the external name of the OpenID Connect service.
+# This is the canonical URL that all clients MUST use to refer to dex. If a
+# path is provided, dex's HTTP service will listen at a non-root URL.
+# NOTE: This needs to exactly match the URL you'll use (in the Provider URL)
+#       Significantly, it will not work with an IP if you connect to a hostname
+#       It may not work with an IP at all. Just use a hostname.
+issuer: http://your.fqdn.here:5556/dex
+
+################################################################################
+# you probably don't need to change anything below here, but it's worth reading
+################################################################################
+
+connectors:
+- type: mockCallback
+  id: mock
+  name: Example
+
+staticPasswords:
+- email: "admin@example.com"
+  # bcrypt hash of the string "password"
+  hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+  # username which will be returned
+  username: "admin"
+  # Type 4 "random" UUID string; value is insignificant
+  userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
+
+# This should be the Kubernetes callback, not Velum (the defaults should work)
+# The id and secret heere are what you paste in to Velum
+staticClients:
+- id: example-app
+  redirectURIs:
+  - 'https://kube-api-x1.devenv.caasp.suse.net:32000/callback'
+  name: 'Example App'
+  secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+
+# Configuration for the HTTP endpoints.  Only change this if you want to test
+# the use of TLS, which requires you to create a key/cert pair and trust the CA
+web:
+  http: 0.0.0.0:5556
+  # Uncomment for HTTPS options.
+  # https: 0.0.0.0:5554
+  # tlsCert: /etc/dex/tls.crt
+  # tlsKey: /etc/dex/tls.key
+
+################################################################################
+# you almost certainly don't need to change or care about anything below here
+################################################################################
+
+enablePasswordDB: true
+storage:
+  type: sqlite3
+  config:
+    file: /go/src/github.com/coreos/dex/examples/dex.db
+
+# metrics/logging; ideally, leave these unchanged
+telemetry:
+  http: 0.0.0.0:5558
+logger:
+  level: "debug"
+  format: "text" # can also be "json"
+
+# we need this for testing
+oauth2:
+  responseTypes: ["code"]
+  skipApprovalScreen: true
+
--- a/lib/velum/dex/oidc.rb
+++ b/lib/velum/dex/oidc.rb
@@ -1,0 +1,23 @@
+require "base64"
+
+module Velum
+  # This generates the JSON that SaltStack uses to list the OIDC connectors
+  module Dex
+    class << self
+      def oidc_connectors_as_pillar
+        DexConnectorOidc.all.map do |con|
+          {
+            type:          "oidc",
+            id:            "oidc-" + con.id.to_s,
+            name:          con.name,
+            provider_url:  con.provider_url,
+            client_id:     con.client_id,
+            client_secret: con.client_secret,
+            callback_url:  con.callback_url,
+            basic_auth:    con.basic_auth
+          }
+        end
+      end
+    end
+  end
+end
--- a/spec/controllers/internal_api/v1/pillars_controller_spec.rb
+++ b/spec/controllers/internal_api/v1/pillars_controller_spec.rb
@@ -435,7 +435,7 @@ RSpec.describe InternalApi::V1::PillarsController, type: :controller do
   end
 
   context "with Dex LDAP connectors" do
-    def expected_dex_json(conn, certificate)
+    def expected_dex_json_ldap(conn, certificate)
       dn_common = "dc=#{conn.host.chomp(".com")},dc=com"
       connector_json = {
         type:            "ldap",
@@ -492,9 +492,55 @@ RSpec.describe InternalApi::V1::PillarsController, type: :controller do
             CertificateService.create(service: dex_connector_ldap, certificate: certificate)
 
             get :show
-            expect(json).to eq(expected_dex_json(dex_connector_ldap, certificate))
+            expect(json).to eq(expected_dex_json_ldap(dex_connector_ldap, certificate))
           end
         end
+      end
+    end
+  end
+  # rubocop:enable RSpec/ExampleLength
+
+  context "with dex OIDC connectors" do
+    def expected_dex_oidc_json(conn)
+      {
+        registries:          [],
+        kubelet:             {
+          :"compute-resources" => {},
+          :"eviction-hard"     => ""
+        },
+        system_certificates: [],
+        dex:                 {
+          connectors: [
+            {
+              id:            "oidc-#{conn.id}",
+              type:          "oidc",
+              name:          conn.name,
+              provider_url:  conn.provider_url,
+              client_id:     conn.client_id,
+              client_secret: conn.client_secret,
+              callback_url:  conn.callback_url,
+              basic_auth:    conn.basic_auth
+            }
+          ]
+        }
+      }
+    end
+
+    context "with basic auth" do
+      let!(:oidc_connector) { create(:dex_connector_oidc, :skip_validation) }
+
+      it "has dex OIDC connectors" do
+        get :show
+        expect(json).to eq(expected_dex_oidc_json(oidc_connector))
+      end
+    end
+
+    context "without basic auth" do
+      let!(:oidc_connector) { create(:dex_connector_oidc, :skip_validation, basic_auth: true) }
+
+      it "has dex OIDC connectors" do
+        get :show
+        expect(json).to eq(expected_dex_oidc_json(oidc_connector))
       end
     end
   end
--- a/spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
+++ b/spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
@@ -1,0 +1,369 @@
+require "rails_helper"
+
+RSpec.describe Settings::DexConnectorOidcsController, type: :controller do
+  let(:user) { create(:user) }
+
+  before do
+    setup_done
+    sign_in user
+  end
+
+  describe "GET #index" do
+    let!(:connector) { create(:dex_connector_oidc, :skip_validation) }
+
+    before do
+      get :index
+    end
+
+    it "populates an array of oidc dex connectors" do
+      expect(assigns(:oidc_connectors)).to match_array([connector])
+    end
+  end
+
+  describe "GET #new" do
+    before do
+      get :new
+    end
+
+    it "sets data validity to be false" do
+      expect(assigns(:is_data_valid)).to be(false)
+    end
+  end
+
+  describe "GET #edit" do
+    let!(:connector) { create(:dex_connector_oidc, :skip_validation) }
+
+    before do
+      get :edit, id: connector.id
+    end
+
+    it "assigns dex_connector_oidc to @dex_connector_oidc" do
+      expect(assigns(:dex_connector_oidc)).not_to be_a_new(DexConnectorOidc)
+    end
+
+    it "return 404 if oidc connector does not exist" do
+      get :edit, id: DexConnectorOidc.last.id + 1
+      expect(response).to have_http_status(:not_found)
+    end
+
+  end
+
+  # rubocop:disable RSpec/ExampleLength, RSpec/MultipleExpectations
+  describe "GET #create" do
+    good_provider = "http://your.fqdn.here:5556/dex"
+
+    it "fails validation with a bad provider" do
+      get :create, validate: true, dex_connector_oidc: {
+        name:          "good oidc",
+        provider_url:  "dead",
+        callback_url:  "http://well.formed.but.invalid/",
+        basic_auth:    true,
+        client_id:     "client",
+        client_secret: "secret"
+      }
+      expect(response).to render_template("new")
+      expect(assigns(:is_data_valid)).to be(false)
+    end
+
+    it "passes validation with a good provider" do
+      VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+        get :create, validate: true, dex_connector_oidc: {
+          name:          "good oidc",
+          provider_url:  good_provider,
+          callback_url:  "http://well.formed.but.invalid/",
+          basic_auth:    true,
+          client_id:     "client",
+          client_secret: "secret"
+        }
+      end
+      expect(response).to render_template("new")
+      expect(assigns(:is_data_valid)).to be(true)
+    end
+
+    it "fails creating with a bad provider" do
+      get :create, validate: false, dex_connector_oidc: {
+        name:          "good oidc",
+        provider_url:  "dead",
+        callback_url:  "http://well.formed.but.invalid/",
+        basic_auth:    true,
+        client_id:     "client",
+        client_secret: "secret"
+      }
+      expect(response).to render_template("new")
+      expect(response).to have_http_status(:unprocessable_entity)
+    end
+
+    it "passes creating with a good provider" do
+      VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+        get :create, validate: false, dex_connector_oidc: {
+          name:          "good oidc",
+          provider_url:  good_provider,
+          callback_url:  "http://well.formed.but.invalid/",
+          basic_auth:    true,
+          client_id:     "client",
+          client_secret: "secret"
+        }
+      end
+      expect(response).not_to have_http_status(:unprocessable_entity)
+      expect(response).to redirect_to(settings_dex_connector_oidcs_path)
+    end
+
+    context "with rendered views" do
+      render_views
+
+      it "shows an error message for empty name" do
+        VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+          get :create, validate: false, dex_connector_oidc: {
+            name:          "",
+            provider_url:  good_provider,
+            callback_url:  "http://well.formed.but.invalid/", # can't be blank in form
+            basic_auth:    true, # also can't be blank - only true/false
+            client_id:     "client",
+            client_secret: "secret"
+          }
+        end
+        expect(response).to render_template("new")
+        expect(response.body).to match(/Name can('|&#39;)t be blank/)
+        expect(assigns(:is_data_valid)).to be_falsey
+      end
+
+      it "shows an error message for empty provider" do
+        get :create, validate: false, dex_connector_oidc: {
+          name:          "good oidc",
+          provider_url:  "",
+          callback_url:  "http://well.formed.but.invalid/", # can't be blank in form
+          basic_auth:    true, # also can't be blank - only true/false
+          client_id:     "client",
+          client_secret: "secret"
+        }
+        expect(response).to render_template("new")
+        expect(response.body).to match(
+          /Provider Url (can('|&#39;)t be blank|is not a valid HTTP URL)/i
+        )
+        expect(assigns(:is_data_valid)).to be_falsey
+      end
+
+      it "shows an error message for empty client id" do
+        VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+          get :create, validate: false, dex_connector_oidc: {
+            name:          "good oidc",
+            provider_url:  good_provider,
+            callback_url:  "http://well.formed.but.invalid/", # can't be blank in form
+            basic_auth:    true, # also can't be blank - only true/false
+            client_id:     "",
+            client_secret: "secret"
+          }
+        end
+        expect(response).to render_template("new")
+        expect(response.body).to match(/Client (Id )?can('|&#39;)t be blank/i)
+        expect(assigns(:is_data_valid)).to be_falsey
+      end
+
+      it "shows an error message for empty client secret" do
+        VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+          get :create, validate: false, dex_connector_oidc: {
+            name:          "good oidc",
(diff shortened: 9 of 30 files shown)


app/views/settings/dex_connector_oidcs/_form.html.slim
```diff
@@ -0,0 +1,73 @@
...
+
+
+  .form-group class="#{error_class_for(@data_holder, :name)}"
+    = f.label :name
+    = f.text_field :name, id: "oidc_name", class: "form-control", value: @data_holder.name, required: true
+    = error_messages_for(@data_holder, :name)
+    small.form-text.text-muted
+      | Name shown to user when selecting a connector
+
+  hr
```
vitoravelino: I know that adding this creates a larger space between the sections in order to create some context but the `h4` naturally does that.

vitoravelino: Maybe creating a specific class for the `h4` and increase the margin for that would be better but I guess just removing `hr` is fine. In the ideal world it would be nice to use `fieldset` with `legend` but that will probably require more work which is not the focus of this PR.

app/views/settings/dex_connector_oidcs/_form.html.slim
```diff
@@ -0,0 +1,73 @@
...
+  h4 Provider
+
+  .form-group class="#{error_class_for(@data_holder, :provider_url)}"
+    = f.label :provider_url, "Provider URL"
+    = f.text_field :provider_url, id: "oidc_provider", class: "form-control", value: @data_holder.provider_url, required: true
+    = error_messages_for(@data_holder, :provider_url)
+    small.form-text.text-muted
+      | Issuer used by OIDC Provider
+    
+  form-group class="#{error_class_for(@data_holder, :basic_auth)}"
```
vitoravelino: I guess you forgot to add `.` here.

app/views/settings/dex_connector_oidcs/_form.html.slim
```diff
@@ -0,0 +1,73 @@
...
+  - external_api_url = "https://#{Pillar.value(pillar: :apiserver)}:32000/callback"
+  .form-group class="#{error_class_for(@data_holder, :callback_url)}"
+    = f.label :callback_url, "Callback URL:"
+    br
+    form-text
+      = external_api_url
+    br
+    small.form-text.text-muted
+      | URL which will be provided to OIDC provider as the callback location
+    = f.hidden_field :callback_url, id: "oidc_callback_url", class: "form-control", value: external_api_url, readonly: true
```
vitoravelino: Instead of showing the url as plain text, I would simply show it as a read-only text input. If we go with that, you can remove the lines from 48 to 51.

spec/features/settings/dex_connector_oidc_feature_spec.rb
```diff
@@ -0,0 +1,142 @@
+require "rails_helper"
+
+# rubocop:disable RSpec/ExampleLength
+# TODO: do we need js?
+# describe "Feature: OIDC connector settings", js: true do
+describe "Feature: OIDC connector settings", js: true do
+  let!(:user) { create(:user) }
+  let!(:dex_connector_oidc) do
+    VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
```
vitoravelino: You don't need to use VCR when calling factories to create objects. VCR should be used to wrap a line or a block of code that interacts with external HTTP requests.

vitoravelino: Just realized that you've created custom validators for attributes that indeed make external HTTP requests and that's probably the reason you added this. A workaround for this would be to skip validation when creating via factories.

I wonder if it's a good idea to have a custom validator for attributes making external HTTP requests. I'd appreciate if someone else with more experience could give an input about it.

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: I initially went that way, but changed my mind. :)  My reasoning for changing to a hidden field and showing it as plain text was that a disabled text field conveys the impression that the data might vary, but isn't directly mutable by the user.  In this case, we're showing the user a static piece of information they will very likely need to use when configuring their OIDC Provider.  Because we're providing the information to the user rather than requesting it from the user, the use of a form field didn't feel right.  My anticipation was that the wording around the URL might be built up a bit more during a user experience review, or at least expanded upon in some end-user documentation.  I further located it in the "Client" section of the page because it just didn't look right separated on the top or of the page, and relates to the client setup.

With that said, I'm really not strongly tied to either approach.  Is there a UX guide or some similar convention documented which would point to one approach v/s the other?

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: Whoops - good catch.  Added.

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: Removing the `hr` does result in a more compact form, which I generally like.  I also agree that long-term it would be better to use CSS applied to the `h4` rather than the somewhat clunky `hr`s.  However, this change will be inconsistent with the LDAP connector.  Maybe that's ok, since the LDAP connector form is significantly longer.  Removing the other `hr` on Line 41 for consistency also brings the "Client" heading uncomfortably close to the "Basic Auth" toggle.

I'm not sure, then.  Would it be better to remove only this `hr`, to remove both for code consistency, or to retain them and maintain the visual consistency with the LDAP connector form?

vitoravelino:
Does it make sense to have this `Validate` button if the validation happens for both `Save` and `Validate` submissions? The only difference is that when clicking on `Save` besides validating it also tries to save the data in the database. Why do we need this additional step if we are not doing anything special?

app/views/settings/dex_connector_oidcs/_form.html.slim
vitoravelino: > Removing the hr does result in a more compact form, which I generally like. I also agree that long-term it would be better to use CSS applied to the h4 rather than the somewhat clunky hrs. However, this change will be inconsistent with the LDAP connector. Maybe that's ok, since the LDAP connector form is significantly longer. 

This is something we should also change on the LDAP connector to keep consistency.

> Removing the other hr on Line 41 for consistency also brings the "Client" heading uncomfortably close to the "Basic Auth" toggle.

I thought I had suggested to add a `br` after the toggle and before the hint text to break the line. Forgot that. :sweat_smile: 

If you do that, it looks fine. See below

screenshot-20180926183159-168x154

spec/features/settings/dex_connector_oidc_feature_spec.rb
dannysauer: That is indeed why it's there. :)

I guessed that the cost of littering VCR wrappers all around the test suite was a worthwhile trade for putting the data validation as close as possible to the database.  Velum's future isn't completely certain, so part of my reasoning was that doing all validation in one place would provide for cleaner support of an alternate interface which interacted via Velum programmatically through the web API rather than through the HTML, for example.  The other reasoning is that this is more reliable and less brittle than a client-side solution.

To explain why the external http call is made to begin with:  the validation duplicates the check Dex makes when it's started (requests webfinger data and compares the issuer).  If this value is inaccurate in the database and gets pushed out to the Dex config, orchestration will fail because Dex will fail to restart.  We therefore want to do as much as possible to ensure that the database never receives data which could cause that orchestration failure.  Longer-term, it might make sense to additionally do a check when updating the Dex config, but that's _way_ beyond the scope of this PR. :)

In any event, since the external call must be made, it makes sense to me to do that in the testing as well. Removing it seems like it would decrease the validity of the testing.

dannysauer: If someone can provide a better way to do this particular test without decreasing the validity of the tests, I'm definitely all for it.  My Ruby is not particularly strong, so there was not much finesse used in this test code. :)

dannysauer:
> Does it make sense to have this `Validate` button if the validation happens for both `Save` and `Validate` submissions? The only difference is that when clicking on `Save` besides validating it also tries to save the data in the database. Why do we need this additional step if we are not doing anything special?

The button is primarily there for consistency with the LDAP provider.  Personally, I kind of like that it gives the user some more explicit indication that we are going to ensure that their values make sense before saving.  But it is technically redundant.

Long term, we discussed potentially adding an additional authentication step to validation, where we would connect to the OIDC Provider using the provided client information and attempt to do a real authentication with a real user.  Right now, minimal validation only verifies that the issuer matches the provider URL.  In that possible future state, we would want to be able to isolate the interactive authentication from the database validation.  Keeping the validate button in place gives the user a consistent experience (normally, they won't know that "save" actually revalidates since they can't click the save button until data is valid) while making it easier to integrate this enhanced validation later.

spec/features/settings/dex_connector_oidc_feature_spec.rb
```diff
@@ -0,0 +1,142 @@
+require "rails_helper"
+
+# rubocop:disable RSpec/ExampleLength
+# TODO: do we need js?
```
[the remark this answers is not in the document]
dannysauer: Whoops - forgot to remove that one.  Another good catch. :)  Running the tests again locally and will push up after making sure it doesn't break.

dannysauer:
What's the right place to record the desired change to the LDAP connector to match 048a640 (there are three `hr+h4` instances)?  Would that be an issue opened in Github?

spec/features/settings/dex_connector_oidc_feature_spec.rb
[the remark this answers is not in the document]
dannysauer: Looks like JS=true is still required to handle clicking on the alert popup when deleting a connector.  So, only removing the comment. :)

spec/features/settings/dex_connector_oidc_feature_spec.rb
vitoravelino: The workaround suggestion, which might not be the final solution, does not decrease the validity of the tests. Actually, it will decrease the verbosity and enable us to write better tests and focus on what is important to be tested. 

The purpose of `create(:dex_connector_oidc)` is to seed the database with a record. We can assume and ensure that the data produced by the factory is valid. This way bypassing the validation wouldn't be an issue. Here's an example [0] of what I was suggesting. That would fix the issue with `create(:dex_connector_oidc)` but wouldn't fix the issue of `FactoryGirl.lint` inside `DatabaseCleaner.cleaning`.

[0] https://stackoverflow.com/a/45965846

app/views/settings/dex_connector_oidcs/_form.html.slim
vitoravelino: Not sure if it was on purpose but the `br` you added was after the hint text. The hint text is still on the right of the toggle buttons. 

screenshot-20180926200430-624x82

The `br` needs to come above `error_messages_for` line to produce the suggestion on the previous comment. You can remove the `br` you added. Sorry if I wasn't clear enough. :sweat_smile:

dannysauer:
The validate button is a technical redundancy, to be sure, but it maintains user-visible workflow consistency with the LDAP interface already deployed.  As such, I'm pretty strongly inclined to at least initially make new stuff look the same as existing.

I'd suggest that integrating validation into the save process from a UI perspective is probably better done in a future change which updates both interfaces simultaneously, rather than mixing new and old approaches in the same build.  Maybe at the same time as those `hr`s are removed from LDAP. :)

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: Oh, whoops.  The `br` I added was to separate the header from the button - moving the text down didn't even occur to me.  Fixed. :)

vitoravelino:
> The validate button is a technical redundancy, to be sure, but it maintains user-visible workflow consistency with the LDAP interface already deployed. As such, I'm pretty strongly inclined to at least initially make new stuff look the same as existing.

The consistency you are talking about is both pages having a button that validates something, right? Because both buttons do different things. In the LDAP, it doesn't refresh the page and validates the connection based on the input fields. In the OIDC, it refreshes the whole page (you probably don't notice because it's localhost and your resolution is large and the form small), it validates the reachability of the OIDC provider based on a single field and also validates any data from the model that will be saved in the database. 

Are you still sure that there's consistency between both pages?

If you really want to have a button to validate/check the provider reachability, I would move it next to the provider url input field and rename it to `Check reachability`. When the user clicks on it, the check would happen without refreshing the page just like the LDAP. Otherwise I would just remove it and rely on the `Save` button as it is. I think that makes much more sense but it's probably out of the scope for this PR

> I'd suggest that integrating validation into the save process from a UI perspective is probably better done in a future change which updates both interfaces simultaneously, rather than mixing new and old approaches in the same build. Maybe at the same time as those hrs are removed from LDAP. :)

You mean "integrating validation into the save process" for the LDAP page, right? Because for OIDC it's already done and we should stick with that IMO.

app/views/settings/dex_connector_oidcs/_form.html.slim
vitoravelino: > I initially went that way, but changed my mind. :) My reasoning for changing to a hidden field and showing it as plain text was that a disabled text field conveys the impression that the data might vary, but isn't directly mutable by the user. In this case, we're showing the user a static piece of information they will very likely need to use when configuring their OIDC Provider. Because we're providing the information to the user rather than requesting it from the user, the use of a form field didn't feel right. My anticipation was that the wording around the URL might be built up a bit more during a user experience review, or at least expanded upon in some end-user documentation. I further located it in the "Client" section of the page because it just didn't look right separated on the top or of the page, and relates to the client setup.

Fair enough. Could you add `span.show` to the text instead of using `form-text` and remove both `br` then? It will make things cleaner.

> With that said, I'm really not strongly tied to either approach. Is there a UX guide or some similar convention documented which would point to one approach v/s the other?

Not really.

One thing I noticed was that the info is only available in the new/edit pages. It's not displayed in the details page, for example. Would it be relevant to also have it there?

spec/features/settings/dex_connector_oidc_feature_spec.rb
dannysauer: Good point; the validation is explicitly tested elsewhere, and these instances are all really used to test something other than creation. 

Adding a trait to the factory (thanks for the example) allowed a number of creations to be cleaned up.  In the process, I noticed that the pillar test cases were written in such a way that they'd always pass, so that's also been cleaned up now.

I'm not sure that it's a problem to have the `lint` done inside datbaase cleaner actually run validation.  That's why the lint is there, right - to validate that the factories work?  One could exclude this particular factory from the normal linting, and instead run it explicitly with the validation skipped.  But IMHO that's quite a bit uglier than just leaving the VCR cassette definition in the database cleaner. :)

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: Converted to `span.show`.  That's cleaner. :)

I'm not sure about showing the callback URL on the details page; we don't show any of the other client details there, and if there's a client defined, the user should have already set up their provider.  It's really only useful before the end user adds a new connector, since they'll need that when setting up their provider.  The only reason it shows up on the edit page is that It didn't seem worth the work to hide it in the shared template, and it's conceivable that the reminder would be useful at that point.

I could see adding it on the connector `index` page, since the user will be there before they click on "add".  But I'm not really sure _where_ would be a good place to put that information on that page.  Maybe if there was a help link which provided a more prosaic explanation of what this connector is for, it would go in there.  Since we don't have that help function and the info should only be needed at the point where "new" is happening, I'm thinking that having it only on new/edit is probably sufficient.

All that said, I'm not adamantly against having the info elsewhere, so I can stick it wherever if someone has a good idea for how to present it. :)

dannysauer:
> The consistency you are talking about is both pages having a button that validates something, right? 

Correct: I'm referring to consistency in user workflow.  Fill out form, click validate, click save.  Granted the OIDC page does reload, but I'm inclined to think the different buttons are more likely to cause a user to think "why isn't this tested like LDAP is?" while they're less likely to notice (or at least think much about) a reload.

Speaking of which, perhaps the validate button on OIDC should be changed to "Test Connection" to match LDAP.  Whoops. :)

> You mean "integrating validation into the save process" for the LDAP page, right? Because for OIDC it's already done and we should stick with that IMO.

Yes - I personally think validating server-side is the right decision in both cases.  Honestly, moreso with LDAP, since there's no reason that the web client needs to reach the LDAP server to begin with; it's within reason that an LDAP auth server might be firewalled and not accessible by desktops.  With OIDC, the browser will almost certainly also have to reach the provider at some point so the user can auth.

To that end, when that future PR which moves LDAP validation server-side happens, it would be trivial to both remove this one's `enabled` attribute from the save button & remove the validate button entirely at the same time, keeping both pieces visually congruent with minimal effort.

lib/velum/dex/oidc.rb
```diff
@@ -0,0 +1,25 @@
...
+            "name"          => con.name,
+
+            "provider_url"  => con.provider_url,
+            "client_id"     => con.client_id,
+            "client_secret" => con.client_secret,
+            "callback_url"  => con.callback_url,
+            "basic_auth"    => con.basic_auth
+          }
+        end
+        oidc_connectors
```
MaximilianMeister: you can just omit the variable decalaration, as the map will be returned implicitly

```ruby
      def oidc_connectors_as_pillar
        DexConnectorOidc.all.map do |con|
          {
            "type"          => "oidc",
            "id"            => "oidc-" + con.id.to_s,
            "name"          => con.name,

            "provider_url"  => con.provider_url,
            "client_id"     => con.client_id,
            "client_secret" => con.client_secret,
            "callback_url"  => con.callback_url,
            "basic_auth"    => con.basic_auth
          }
        end
      end
```

app/controllers/settings/dex_connector_oidcs_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+  def edit
+    @is_data_valid = false
+  end
+
+  def create
+    @data_holder = data_holder_type.new(data_holder_params)
+
+    if params[:validate]
+      @is_data_valid = @data_holder.valid?
+      # TODO: figure out how to make notice actually display :)
```
MaximilianMeister: i think they should be displayed/rendered via the layout, see https://github.com/kubic-project/velum/blob/master/app/views/layouts/settings.html.slim#L28 - @vitoravelino any idea here?

vitoravelino:
> PR #647 seems to have introduced a CodeClimate failure, which slipped in here when I rebased on current master. So those commits (084b072 and the two following) have been un-merged from this branch to keep things clean.

IIRC rebasing won't show past commits unless the target branch doesn't have them. What I usually do to rebase is:

```sh
# supposing `upstream` is the name of the main repository
$ git fetch upstream
$ git rebase upstream/master
$ git push origin my-branch --force
```

Hope it helps! :wink:

dannysauer:
> IIRC rebasing won't show past commits unless the target branch doesn't have them. What I usually do to rebase is:

That's precisely what I did. :)  The relevant commit was merged into master after this PR was opened, so they weren't originally in my branch.  Since they merged (and were `rebase -i` drop'd) cleanly, I decided to leave them removed from this branch for now to prevent muddying the water re: whether the checks fail because of this changeset or something else.  I figure I'll (re)rebase on master at the end when I squash the commits before merge.

lib/velum/dex/oidc.rb
dannysauer: Whoops.  Fixed in 1240800. :)

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: I assume it's a difference between `redirect_to` v/s `render`.  This is my first time working with Rails, so it wasn't obvious, and then I forgot to look that up after I got the rest working. :)

I did a little research this morning, now that I was reminded of this TODO.  It appears that render doesn't take the flash as input?  Either way, if I directly assign to `flash[:notice]` or `flash[:alert]` before the render, that works properly with the `render`.  I'm not positive if that's the "right" way to do it, though.  How about this, @MaximilianMeister @vitoravelino?
```ruby
      if @is_data_valid
        flash.now[:notice] = "#{@data_holder.class} is valid"
      else
        flash.now[:alert]  = "#{@data_holder.class} is invalid"
      end
      render action: :edit
```

I needed to use flash.now so the "is invalid" message doesn't stay visible when it's corrected (or vice-versa).

dannysauer: Well, `render action: :new` here; I tested it on the edit page. :)

lib/velum/dex/oidc.rb
dannysauer: And now Rubocop rightfully complains about the useless assignment.  Whoops, again.  I'll fix that when I push up the fix for the alerts in the other conversation. :)

dannysauer: Actually fixed in 539a210

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: There - works properly in f0fbdcf.

app/controllers/settings/dex_connector_oidcs_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    @is_data_valid = false
+  end
+
+  def create
+    @data_holder = data_holder_type.new(data_holder_params)
+
+    if params[:validate]
+      @is_data_valid = @data_holder.valid?
+      if @is_data_valid
+        flash.now[:notice] = "#{@data_holder.class} is valid"
```
vitoravelino: We can simplify with just `flash[:notice]`. 

Also the current message is `DexConnectorOidc is valid`, wouldn't it be better if we used `OIDC Connector is valid`? What do you think?

app/controllers/settings/dex_connector_oidcs_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    @oidc_connectors = DexConnectorOidc.all
+  end
+
+  def new
+    @is_data_valid = false
+    @data_holder = data_holder_type.new
+  end
+
+  def edit
+    @is_data_valid = false
```
vitoravelino: This is more of a personal preference but if we omit `@is_data_valid`, whenever we refer it in the templates it would be already a falsy value.

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: If we just use `flash`, then if the user doesn't clear the alert, it persists. So, enter an invalid value, validate, then enter a valid value. On that page, there will be both the "invalid" warning and the "valid" alert. Using `flash.now` mitigates that issue.

The message should probably be changed; I'll do that when I get back. :)

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: I'll have to stare at that one to see if it makes the code significantly cleaned; even after years of Perl, I really don't like depending on implicit initialization. _Maybe_. :)

app/controllers/settings/dex_connector_oidcs_controller.rb
vitoravelino: You are right about `now`. Never had to use it though.

vitoravelino:
When I fill an invalid provider url I get duplicated error message

screenshot-20180928174538-246x156

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: Looking at it, I'm not quite sold on this one.  The var is explicitly set to the result of`.valid?` in create and update, so I think in this specific case I like the consistency a little bit better when it's being set for all four views that use it.  I'm not completely against taking the two in `new` and `edit` out if you're absolutely adamant about making things as compact as possible, though.

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: I've only known about it for a few hours. ;)

Better friendly_name implemented; will push after re-running the test cases.

dannysauer:
The duplicated error message is because the model was using the `http_url` validator, and the OIDC connection test also validates that it's a http url first.  So it was validating twice, causing the message to be added twice.  I saw that in testing but thought I remembered fixing it.  Guess not. :) Thanks.

Fixed (in 8e02aa8) by removing the explicit http_url validator on the  provider URL's field in the model.

app/controllers/settings/dex_connector_oidcs_controller.rb
dannysauer: ad72e37

dannysauer:
So, what remains to get this merged in?  Just remove the branch references in the OIDC testing document and collapse all the commits?

vitoravelino:
A `validators` folder exists, just realized that the validators created here are not there. Any special reason for that?

I'm still not agreeing with the unnecessary test/validate button workflow and the VCR line wrapping the DatabaseCleaner call.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -527,4 +528,45 @@ def expected_dex_json(num, certificate)
     end
   end
   # rubocop:enable RSpec/ExampleLength
+
+  context "with dex OIDC connectors" do
+    let!(:oidc_connector) { create(:dex_connector_oidc, :skip_validation, name: con_name) }
+    # because name increments every time factory is called, it's not predictable
+    let(:con_name) { "pillar test OIDC connector" }
```
vitoravelino: This is the same as explained at https://github.com/kubic-project/velum/pull/672#issuecomment-425859643. The factory generates the data, use that in our favor.

dannysauer:
> A `validators` folder exists

Not sure about the validators; the few examples I looked at said that custom database validators go in the `concerns` folder, so I didn't ask questions. :)  Looking at https://www.sitepoint.com/a-quick-study-of-the-rails-directory-structure/ and a few other similar sites, the `controllers/concerns` directory is well-known for this general sort of purpose (locating common code used by controllers).  On the other hand, `validators` not only isn't something I can find documented outside of one now-deleted blog post, it's located away from the data controllers that use it.  I guess the better question might be "why is there a `validators` directory?" :)

I don't have a huge problem with moving the files, as the name "concerns" doesn't make a lot of sense to me, but it'd be nice if there was some definitive document or similar guidance on the subject.

dannysauer:
> I'm still not agreeing with the unnecessary test/validate button workflow

I'm in agreement with ultimately changing this to only "validate on save", but I'd really rather see that in a separate PR where the LDAP connector is changed the same way, and making that LDAP change concurrently would be outside the scope of this PR.

dannysauer:
> and the VCR line wrapping the DatabaseCleaner call.

I started to change that, but the only way I can see to make that happen is to disable database validation on the factory by default.  I decided not to use that solution, because anyone overriding any parameter in a factory call elsewhere would also need to be aware of the "not validated by default" nature.  Normally, one would expect the database validation to happen when using a database record-generating factory, I think, so that would be a bad solution.

The other option that I thought of is to explicitly pick the dex_connector_oidc factory out of the list of factories and override how it's called in the cleaner definition.  But that, to me, would be quite a bit uglier than the VCR wrapper.  I'm very much open to discussing any other solution.  I don't really like the way it is, but it's the best I've come up with so far.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
dannysauer: I was literally just fixing that, having cleaned up the LDAP test. :)

vitoravelino:
>  Not sure about the validators; the few examples I looked at said that custom database validators go in the concerns folder, so I didn't ask questions. :) Looking at https://www.sitepoint.com/a-quick-study-of-the-rails-directory-structure/ and a few other similar sites, the controllers/concerns directory is well-known for this general sort of purpose (locating common code used by controllers). On the other hand, validators not only isn't something I can find documented outside of one now-deleted blog post, it's located away from the data controllers that use it. I guess the better question might be "why is there a validators directory?" :)

Because we want to group all the validators in one place. There are other approaches of grouping but the one we've been using is based on functionality.

> I don't have a huge problem with moving the files, as the name "concerns" doesn't make a lot of sense to me, but it'd be nice if there was some definitive document or similar guidance on the subject.

Explanation below extracted from [0]: 

> All subdirectories of app in the application and engines present at boot time. For example, app/controllers. They do not need to be the default ones, any custom directories like app/workers belong automatically to autoload_paths.

[0] https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths-and-eager-load-paths

dannysauer:
> They do not need to be the default ones, any custom directories

Ok, that information is what I was looking for.  Thanks. :)  Moved to `validators` in 628f0b9

spec/controllers/internal_api/v1/pillars_controller_spec.rb
dannysauer: Updated in 25a6d12.

dannysauer:
> > I'm in agreement with ultimately changing this to only "validate on save", but I'd really rather see that in a separate PR where the LDAP connector is changed the same way, and making that LDAP change concurrently would be outside the scope of this PR.
> 
> So, if I understood correct, you prefer something that is not better just because of something that is already there? It's just a matter of removing stuff at this point, things will become even simpler.

In the case of UI/UX, maybe. :)  I'm not sure that having the separate validation button is actually _not_ better.  There's no disputing that it's technically redundant, but the end user doesn't see that.  If the button is simply removed, then an end user who's seen the LDAP connector may wonder if the OIDC connector is also validated.  The only way that validation will be communicated is if they enter a bad value and see the failure on the form.  If they enter good information and hit save -- which is hopefully the common case -- they can't be sure.  Including an explicit, *visible* validation step removes that doubt.

An alternative which addresses the same concern might be adding a note either in the documentation or on the form itself indicating that the connector will be tested before being saved, and errors will be flagged.

To be fair, I really wouldn't be sad if it's removed nor happy if it's kept; it's minimal work either way.  It seems like the main sticking point here is whether or not the redundant button is justified by a desire for consistency.  Do we have a user interface expert or team who has a more vested interest in the user experience, and who can contribute how important it actually is to maintain that visual consistency?  I think that relative weighting between technical simplicity and user interface is what this discussion is missing.

vitoravelino:
> I started to change that, but the only way I can see to make that happen is to disable database validation on the factory by default. I decided not to use that solution, because anyone overriding any parameter in a factory call elsewhere would also need to be aware of the "not validated by default" nature. Normally, one would expect the database validation to happen when using a database record-generating factory, I think, so that would be a bad solution.

Indeed that wouldn't be a good solution. 

My guts says that having a validation that reaches an external service in a model is not good. And what makes me think that is the dependency created whenever the model is used. Disabling the validation on the factory during the creation improved a lot that problem but was not enough IMO.

The LDAP page will need to also solve this problem when we move to the "validate on save" approach. Maybe we could address this issue when LDAP "validate on save" is done, as long as the solution is different from the one proposed in this PR unless we find an alternative in the meantime.

dannysauer:
> My guts says that having a validation that reaches an external service in a model is not good. And what makes me think that is the dependency created whenever the model is used. 

In this particular case, it's the database structure and the larger system which I think justifies it.  As-is, any value in the database gets provided to Dex during orchestration, and Dex verifies by actually connecting on startup.  If the issuer doesn't match what the external OIDC Provider returns (or if the OP isn't reachable), Dex fails to start, and orchestration fails.  It's been suggested that we could alter the database structure to add a "validated" field which indicates whether or not a connector has been verified (or maybe when it was validated) and is thus safe to deploy.  At that point, it would be safe to reduce the validation to only format checks, and the "second-stage validation" could make the external calls.

If this second-stage validation was initiated when orchestration happened, all connectors could be revalidated each time.  That would also help ensure that Dex didn't fail to restart if something changes between initial validation and a later reorchestration.  While I don't love the idea of potentially "bad" data in the database, I do like the potential increase in robustness that would provide.

Moving the candidate connectors to a separate table and replacing the current tables with a read-only view showing only "validated" data would allow this to be mostly transparent to the data consumers, and would allow the second-stage validation to happen entirely in Velum.  I think.

> The LDAP page will need to also solve this problem when we move to the "validate on save" approach. Maybe we could address this issue when LDAP "validate on save" is done, as long as the solution is different from the one proposed in this PR unless we find an alternative in the meantime.

Yes, it would make sense to combine the back-end validation rework with the front-end interface appearance, since they're fairly interdependent changes.

dannysauer:
After:
 * rebase on current master
 * squash the commits
 * track "tweak validation workflow" somewhere
Are we in good shape now?

app/views/settings/dex_connector_oidcs/_form.html.slim
```diff
@@ -0,0 +1,70 @@
...
+    = error_messages_for(@data_holder, :client_id)
+    small.form-text.text-muted
+      | Client ID provided by OIDC Provider
+    
+  .form-group class="#{error_class_for(@data_holder, :client_secret)}"
+    = f.label :client_secret, "Secret"
+    = f.password_field :client_secret, id: "oidc_client_secret", class: "form-control", value: @data_holder.client_secret, required: true
+    = error_messages_for(@data_holder, :client_secret)
+    small.form-text.text-muted
+      | Secret associated with Relying Party's Client ID
```
nanoscopic: I think this is meant to say "Relaying" not "Relying"

docs/oidc_testing/README.md
```diff
@@ -0,0 +1,81 @@
...
+1. Set up dev environment (which you've probably already done)
+   ```
+   git clone git@github.com:kubic-project/automation
+   cd automation
+   ./caasp-devenv --setup
+   ```
+1. check out OIDC-aware velum
+   ```
+   cd ../velum
+   git remote add dannysauer git@github.com:dannysauer/velum
```
nanoscopic: I don't think that readme in kubic-project should refer to personal github accounts. Especially not to one that is a copy of velum from within the velum repo itself.

docs/oidc_testing/README.md
```diff
@@ -0,0 +1,81 @@
+# Bring up a cluster w/ the OIDC code enabled.
+1. Set up dev environment (which you've probably already done)
```
nanoscopic: "which you've probably already done" can simply be removed.

docs/oidc_testing/README.md
```diff
@@ -0,0 +1,81 @@
...
+   git fetch dannysauer
+   git checkout dannysauer/oidc_support
+   ```
+1. `cd ../automation && ./caasp-devenv --build --bootstrap` (with whatever other options, like `-L provo`)
+
+# Bring up the stand-alone dex OIDC Provider (OP)
+1. `cd ../velum/docs/oidc_testing`
+1. `cp examples/config-dev.yaml{.example,}`
+1. edit `examples/config-dev.yaml`
+   * there is a "NOTE:" in each place you need to review, which will show up an alternate color in vim w/ syntax hilighting
```
nanoscopic: Mention of 'alternate color in vim' is overly verbose here. Mention of 'NOTE' even is also verbose. This can be dropped. Saying "modify as needed to suit your configuration" or something like that is sufficient.

docs/oidc_testing/README.md
```diff
@@ -0,0 +1,81 @@
...
+   * set the `issuer:` URL to match the DNS host name of your development machine.  Note that this hostname must resolve _via DNS_ from the master and worker nodes running the dex_dex container.  You can check this via: `docker exec _container_id_ ping -c 1 _host_name_`.  You should get a response back, and it should show the IP address of the host.  The URL used as the issuer _must_ be exactly the URL you use in the OIDC provider URL field in Vellum".
+   * If you do not have a DNS-resolvable name, there are a variety of ways to fix that.  The easiest is to add an entry to the libvirt DNS resolver, which will make the hostname/IP mapping work from the VMs and containers within, but won't affect anything outside the cluster.  Say you're using `dexhost.do.main` as your issuer name, and it's listening on `192.168.0.213`.  Do this, and then retry the ping test above:
```
nanoscopic: Using a non-dns resolvable name is outside the scope of what we wish to support. Please remove this.

docs/oidc_testing/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+services:
+  base:
+    image: quay.io/coreos/dex:v2.10.0
+
+  dex:
+    extends:
+      service: base
+    volumes:
+      - "./examples/:/go/src/github.com/coreos/dex/examples"
+      #- "/var/lib/dex:/var/lib/dex"
```
nanoscopic: Commented out lines should be dropped.

docs/oidc_testing/examples/.gitignore
```diff
@@ -0,0 +1,2 @@
+config-dev.yaml
+dex.db
```
nanoscopic: This should be dropped. It was mentioned elsewhere that this file is no longer present.

app/views/settings/dex_connector_oidcs/_form.html.slim
dannysauer: I truly appreciate someone looking closely enough to catch that. :)

While "relaying" would totally make sense, the app which federates authentication to an OIDC Provider (OP) is referred to as an RP, or "relying party".  I presume that's because it _relies_ upon information from the OP.  I was hesitant to use any of the "real" OIDC names, but it felt necessary here for precision; saying "application" or similar here seemed too vague.

Reference: https://openid.net/connect/faq/ (about halfway down; dang page doesn't use anchors on FAQ entries)

docs/oidc_testing/examples/config-dev.yaml.example
```diff
@@ -0,0 +1,100 @@
+# The base path of dex and the external name of the OpenID Connect service.
```
nanoscopic: This entire file should be made more minimal to suit our needs. All upstream comments for customization do not need to be included.

lib/velum/dex/oidc.rb
```diff
@@ -0,0 +1,24 @@
+require "base64"
+
+module Velum
+  # This class offers the integration between ruby and the Saltstack API.
```
nanoscopic: This comment doesn't make sense and/or doesn't seem to belong. Copied from elsewhere?

docs/oidc_testing/README.md
dannysauer: For testing purposes, I suspect that most of our developers will be standing up the docker container on their local dev system, which very well might not resolve via DNS.  I'd really rather leave that instruction in there, since it's most likely to benefit one of us. :)

lib/velum/dex/oidc.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Dex
+    class << self
+      def oidc_connectors_as_pillar
+        DexConnectorOidc.all.map do |con|
+          {
+            "type"          => "oidc",
```
nanoscopic: It's generally recommended to use the new json format.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
```diff
@@ -0,0 +1,371 @@
...
+  # rubocop:disable RSpec/ExampleLength
+  describe "GET #create" do
+    good_provider = "http://your.fqdn.here:5556/dex"
+
+    it "fails validation with a bad provider" do
+      get :create, validate: true, dex_connector_oidc: {
+        name:          "good oidc",
+        provider_url:  "dead",
+        callback_url:  "http://well.formed.but.invalid/",
+        basic_auth:    true,
```
nanoscopic: Is there a test with basic_auth set to false?

docs/oidc_testing/README.md
dannysauer: I agree with removing the vim part.  Later steps assume that only the `NOTE`-delimited sections have been modified, so suggesting only modifying those two places seemed like a good idea.  I think I can word this better, though.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
```diff
@@ -0,0 +1,371 @@
...
+    context "with rendered views" do
+      render_views
+
+      it "shows an error message for empty name" do
+        VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+          get :create, validate: false, dex_connector_oidc: {
+            name:          "",
+            provider_url:  good_provider,
+            callback_url:  "http://well.formed.but.invalid/", # can't be blank in form
+            basic_auth:    true, # also can't be blank - only true/false
```
nanoscopic: There doesn't appear to be any test with basic_auth set to false. There should be a test to ensure if it is false it outputs as false.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
```diff
@@ -0,0 +1,371 @@
...
+            client_id:     "client",
+            client_secret: ""
+          }
+        end
+        expect(response).to render_template("new")
+        expect(response.body).to match(/Client Secret can('|&#39;)t be blank/i)
+        # TODO: expect save button to be disabled
```
nanoscopic: TODO's should be resolved before merge / accept of the PR.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
```diff
@@ -0,0 +1,371 @@
...
+            client_secret: "secret"
+          }
+        end
+        expect(response).to render_template("new")
+        expect(response.body).to match(/is not a valid OIDC provider/)
+        expect(response.body).to match(/discovery failure/)
+      end
+      # rubocop:enable RSpec/MultipleExpectations
+
+      # it "shows an error message for 404-generating OIDC issuer"
```
nanoscopic: This appears to be a todo?

spec/features/settings/dex_connector_oidc_feature_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+  end
+
+  describe "#new" do
+    before do
+      visit new_settings_dex_connector_oidc_path
+    end
+
+    it "allows a user to create an OIDC connector" do
+      VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
```
nanoscopic: Is a VCR needed? Reviewing the changes I saw that the other VCR's were able to be done away with. Is it possible to do similar here?

spec/features/settings/dex_connector_oidc_feature_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it "allows a user to edit an oidc connector clicking validate and then save" do
+      VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+        fill_in id: "oidc_name", with: "a new name"
+
+        click_button("Test Connection")
+        click_button("Save")
+        expect(page).to have_content("OIDC Connector was successfully updated.")
+      end
+    end
+
```
nanoscopic: Extra carriage return here.

spec/features/settings/dex_connector_oidc_feature_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      VCR.use_cassette("oidc/validate_connector", allow_playback_repeats: true, record: :none) do
+        accept_alert do
+          click_on("Delete")
+
+          expect(page).not_to have_content(dex_connector_oidc.name)
+          expect(page).to have_content("OIDC Connector was successfully removed.")
+          expect(page).to have_current_path(settings_dex_connector_oidcs_path)
+        end
+      end
+
```
nanoscopic: Extra carriage return here also.

spec/models/dex_connector_oidc_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    "",
+    "http:://x",
+    "fqdn.is.invalid",
+    "http://user:fqdn.is.invalid/"
+  ]
+  bad_but_well_formed_http_urls = [
+    "http://fqdn.is.invalid",
+    "http://user@fqdn.is.invalid/",
+    "https://user:pass@fqdn.is.invalid/",
+    # "https://192.0.2.0",      # RFC5737 TEST-NET-1, should trigger timeout
```
nanoscopic: Does this line need to be here?

spec/spec_helper.rb
```diff
@@ -20,6 +20,13 @@
   c.ignore_request do |request|
     request.uri =~ /__identify__/
   end
+  # Don't try to playback under the RFC6761 .invalid TLD (should always return NXDOMAIN)
```
nanoscopic: Are these cases for weird IPs important? Why are they here? This seems like premature optimization / handling cases that will never occur for the user.

docs/oidc_testing/README.md
dannysauer: Correct.  Those two references need to be removed, since they won't be relevant after this is merged.  Thanks. :)

Fixed in cbed0d6

docs/oidc_testing/README.md
dannysauer: Fixed in cbed0d6

docs/oidc_testing/README.md
dannysauer: Reworded in cbed0d6

docs/oidc_testing/examples/.gitignore
dannysauer: Good catch.  That file gets created dynamically when the test container is started up.  It's in .gitignore now so that someone who does some testing won't accidentally add their test database, potentially with a sensitive password, and so it doesn't show up in "git status" as untracked files. :)

docs/oidc_testing/docker-compose.yml
dannysauer: Cleaned up in 9ebfe6c

docs/oidc_testing/examples/config-dev.yaml.example
dannysauer: I agree.  My original line of thinking was to leave it as close to upstream as possible to make it easier to merge future upstream changes in, but readability suffers as a result.  I'll clean that up.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
dannysauer: It's a comment which was implemented elsewhere and then I forgot to remove it. :)

fixed in e337b2b

spec/features/settings/dex_connector_oidc_feature_spec.rb
dannysauer: Sure is.

The others don't need it because the data in those cases is being directly inserted into the database by a known-good, static source, so exercising the connection validation was redundant.  This one actually needs VCR, since it's testing the actual "user clicks validate before save" flow in a simulated browser, which takes data from the form.  This is also known-good data, but there's no way to bypass the validation in this one because it's coming in from the web form instead of direct to the DB via the factory.

spec/features/settings/dex_connector_oidc_feature_spec.rb
dannysauer: Fixed in 4410137

spec/features/settings/dex_connector_oidc_feature_spec.rb
dannysauer: Fixed in 4410137

spec/models/dex_connector_oidc_spec.rb
dannysauer: Whoops.  I was going to investigate reducing the timeout in the test cases, then uncomment that one.

spec/spec_helper.rb
dannysauer: The idea was to verify that the most common typos are caught.  The `invalid` TLD was specifically chosen for the "validation fails when OIDC server domain is typo'd" tests, because the NXDOMAIN reply can't be simulated using VCR. This is both cleaner (IMO) and less brittle than using mock to override a method hidden a few calls deep within the OIDC connector gem.  It'd be really annoying to track down a test case which spontaneously started failing because someone upstream refactored their gem. :)

dannysauer: The spec test cases test the database model, which includes the validators.  So most of these weird URLs are to exercise the validator behavior on the most plausible things someone might stick in.  To provide individual explanations of each one:

This set should cause failures in the http URL format validator, all of which are things we want to catch:
* `ftp://fqdn.is.invalid`
  * wrong protocol - ensure that http URL validator catches a protocol error
* `""`
  * empty - ensure that "present" check works properly
* `http:://x`
  *  too many colons (this happened more than once in testing, and not just by me) - caught by http URL validator match
* `fqdn.is.invalid`
  * no protocol specified, just bare hostname - caught by http URL validator
* `http://user:fqdn.is.invalid/`
  * username:password format URL with no hostname; unlikely, but since this is mostly testing the http URL validator, worthwhile to make sure parsing works as expected.

These should pass the http validator but fail the OIDC validator, thus simultaneously acting as positive tests for http and negative tests for OIDC:
* `http://fqdn.is.invalid`
  * test NXDOMAIN, as above
* `http://user@fqdn.is.invalid/`
  * NXDOMAIN with a username provided
* `https://user:pass@fqdn.is.invalid/`
  * NXDOMAIN with a username and password provided, the third of the three most-likely http URL formats
* `https://192.0.2.0`
  * Address from RFC5737 TEST-NET-1, should trigger timeout because it should not exist anywhere routeable, thus validating handling of a connection timeout
* `http://192.0.2.0:65537`
  * too-high port, which should ultimately trigger a socket error (double-typing a port number is not unlikely)
* `https://1.2.3.256`
  * invalid IP address, which could easily happen due to a typo


I could likely drop the basic auth URL format checks, but they're cheap to do and are things that someone could _technically_ use (though I definitely would not).  I'm basically following the "be liberal in what you accept and strict on what you return" philosophy here.  Since it's in the test cases and not the app and the loops will be there either way to test a couple examples both ways, the relatively trivial extra work to catch the majority of potential mistakes feels justified.

spec/features/settings/dex_connector_oidc_feature_spec.rb
nanoscopic: Ok understood. Thank you for the explanation.

docs/oidc_testing/examples/.gitignore
nanoscopic: That reasoning makes sense. Thanks.

docs/oidc_testing/README.md
nanoscopic: I am fine with readme being present but it should be more minimal compared to what it is now. A lot of it won't be correct once it is actually merged in.

I do agree that knowing how to set the IP for the hostname is useful information to know ( it has helped me certainly ), my concern is that mentioning it may imply that we support people doing so which we do not.

app/views/settings/dex_connector_oidcs/_form.html.slim
nanoscopic: Thanks for the extra information. I was not aware of that. Good job on using the correct terminology here and other places also.

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
dannysauer: Fixed in 86830cb

docs/oidc_testing/examples/config-dev.yaml.example
dannysauer: Completely restructured to be a bit more useful in 4766fbf

lib/velum/dex/oidc.rb
dannysauer: I'll point at the `git blame` output to explain that one. ;)

This is the file which generates the list of OIDC connectors for SaltStack to consume.  Slightly more clear comment moved to top of file in fb2d6f5.

lib/velum/dex/oidc.rb
dannysauer: Changed in 4f7811b

spec/controllers/settings/dex_connector_oidcs_controller_spec.rb
dannysauer: Pillar test added in 78ace6e.  More useful tests "coming soon".

dannysauer: Upon further discussion, the pillar test is sufficient, as that's the only place it's consumed.  Any further testing would be in Salt and Dex.

app/views/settings/dex_connector_oidcs/_form.html.slim
```diff
@@ -0,0 +1,70 @@
...
+      = f.label :basic_auth, nil, class: "btn btn-default #{'btn-primary active' if @data_holder.basic_auth}"
+        = f.radio_button :basic_auth, "true", checked: @data_holder.basic_auth
+        | On
+      = f.label :basic_auth, nil, class: "btn btn-default #{'btn-primary active' unless @data_holder.basic_auth}"
+        = f.radio_button :basic_auth, "false", checked: !@data_holder.basic_auth
+        | Off
+    = error_messages_for(@data_holder, :basic_auth)
+    br
+    small.form-text.text-muted
+      | Set to "Off" only if this OIDC provider requires credentials to be sent using GET (this is uncommon)
```
dannysauer: This text needs to be adjusted to better reflect what the parameter does and why it's needed.

dannysauer: Fixed in cad2fc9

dannysauer:
Rebased on current master yesterday.

Walked through the full config again.  Build and auto-bootstrap using KVM build 454, then follow the OIDC testing instructions to set up a test dex and log in with that OIDC provider by visiting `/oidc` on the admin node.  This results in the kubeconfig download page.  I did not use kubectl with that configuration file, but the contents look reasonable.

suseclee:
I verified this velum code with salt (https://github.com/dannysauer/salt.git) and updated master automation repo (11/5). It worked as it is expected. 

After I created ,surely using kubconfig from admin at first ,clusterrolebinding such as "kubectl create clusterrolebinding oidc-test  --namespace=default --clusterrole=edit --user=<email.from.oidc.provider>", I could use kubectl from the oidc user for the default namespace

Great work!!! David and Danny.

Everything looks good to me.

PR #666, opened 2018-09-25
[bsc#1109720] test for salt 'alter_time' attributes

bear454:
Incorporate the `alter_time` patch into the schema by default, and test that conditions are met.

Instead of having to remember to update a non-obvious patch when adding a migration, developers must remember not to include changes to the affected schema lines, and will be assisted by a spec that verifies the proper conditions are maintained.

BONUS: test for the conditions _without_ analyzing the text content of the schema file, so that whenever velum moves to Rails 5, we can incorporate the proper upstream implementation[1]

[1] https://stackoverflow.com/a/40647191


flavio:
@MaximilianMeister can you take a look at that please?

PR #672, opened 2018-09-28
Fix broken LDAP connector test case syntax

dannysauer:
I noticed that a couple of pillar test cases were passing despite an error in the test.  This PR adjusts those test cases so they correctly fail upon discrepancies.

Ultimately, this ended up being a fairly significant refactor of the test code, but the test still actually works the same way; it's just a bit more concise.

### DIFF
--- a/spec/controllers/internal_api/v1/pillars_controller_spec.rb
+++ b/spec/controllers/internal_api/v1/pillars_controller_spec.rb
@@ -434,46 +434,44 @@ RSpec.describe InternalApi::V1::PillarsController, type: :controller do
     end
   end
 
-  def expected_dex_json(num, certificate)
-    {
-      id:              num,
-      name:            "LDAP Server #{num}",
-      root_ca_data:    Base64.encode64(certificate.certificate),
-      bind:            {
-        anonymous: false,
-        dn:        "cn=admin,dc=ldap_host_#{num},dc=com",
-        pw:        "pass"
-      },
-      username_prompt: "Username",
-      user:            {
-        base_dn:  "cn=users,dc=ldap_host_#{num},dc=com",
-        filter:   "(objectClass=person)",
-        attr_map: {
-          username: "uid",
-          id:       "uid",
-          email:    "mail",
-          name:     "name"
-        }
-      },
-      group:           {
-        base_dn:  "cn=groups,dc=ldap_host_#{num},dc=com",
-        filter:   "(objectClass=group)",
-        attr_map: {
-          user:  "uid",
-          group: "member",
-          name:  "name"
+  context "with Dex LDAP connectors" do
+    def expected_dex_json(conn, certificate)
+      dn_common = "dc=#{conn.host.chomp(".com")},dc=com"
+      connector_json = {
+        type:            "ldap",
+        id:              conn.id,
+        name:            conn.name,
+        root_ca_data:    Base64.encode64(certificate.certificate),
+        server:          "#{conn.host}:#{conn.start_tls ? 389 : 636}",
+        start_tls:       conn.start_tls,
+        bind:            {
+          anonymous: false,
+          dn:        "cn=admin,#{dn_common}",
+          pw:        "pass"
+        },
+        username_prompt: "Username",
+        user:            {
+          base_dn:  "cn=users,#{dn_common}",
+          filter:   "(objectClass=person)",
+          attr_map: {
+            username: "uid",
+            id:       "uid",
+            email:    "mail",
+            name:     "name"
+          }
+        },
+        group:           {
+          base_dn:  "cn=groups,#{dn_common}",
+          filter:   "(objectClass=group)",
+          attr_map: {
+            user:  "uid",
+            group: "member",
+            name:  "name"
+          }
         }
       }
-    }
-  end
-
-  # rubocop:disable RSpec/ExampleLength
-  context "with dex LDAP connectors tls" do
-    it "has dex LDAP connectors" do
-      dex_connector_ldap = create(:dex_connector_ldap, :tls, :regular_admin)
-      CertificateService.create(service: dex_connector_ldap, certificate: certificate)
-
-      expected_json = {
+      connector_json = connector_json.merge(bind: { anonymous: true }) if conn.bind_anon
+      {
         registries:          [],
         kubelet:             {
           :"compute-resources" => {},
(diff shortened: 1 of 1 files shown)


spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -434,97 +434,107 @@
...
-    it "has dex LDAP connectors" do
-      dex_connector_ldap = create(:dex_connector_ldap, :tls, :regular_admin)
-      CertificateService.create(service: dex_connector_ldap, certificate: certificate)
+  context "with Dex LDAP connectors" do
+    before do |example|
+      # calculate a likely-unique 0-100 integer hash value by adding up the
+      # integer values of chars in the example name (context + example)
+      u = example.metadata[:full_description].codepoints.reduce(0, :+) % 100
+      @conn_name = "LDAP connector #{u}"
+      @host_name = "ldap_host_#{u}.com"
```
vitoravelino: Why do we need this? The factory is supposed to take care of that as we can see [here](https://github.com/kubic-project/velum/blob/master/spec/factories/dex_connectors_ldap_factory.rb#L3). Is it not working?

dannysauer:
The factory doesn't return the number assigned by the internal sequence (or rather it doesn't do so in a way obvious to me), so the auto increment database ID doesn't necessarily reflect the value the factory uses. This provides predictable uniqueness.

vitoravelino:
> The factory doesn't return the number assigned by the internal sequence (or rather it doesn't do so in a way obvious to me), so the auto increment database ID doesn't necessarily reflect the value the factory uses. This provides predictable uniqueness.

What do you mean by "The factory doesn't return the number assigned by the internal sequence"? Maybe an example of what is not happening and you would like it to happen?

dannysauer:
Sure. The factory has a sequence which is used in generating the hostname and connector name. Each time the factory is used, that number increments. In these tests, we need to be able to match those fields. When this test is run stand-alone, the database auto-increment id field and sequence both start at the same point, so we can get the hostname using the database id. Because of the db cleaning, when the whole test suite runs, though, the database ID might be 5 while the sequence is 27.

We could use a static string. That would fail a unique name constraint, though, and makes debugging harder. What we want here is to ensure that the row we inserted is what we get back, which means a unique name. So this uses the test name to generate that unique ID separate from the factory sequence, saving us from relying on a relationship between the auto-increment field in the database and the factory sequence.

vitoravelino:
You are passing the generated manually through the constructor and expecting it. Woudn't it be the same as using `dex_connector_ldap.name` and `dex_connector_ldap.host`? It's already done with `dex_connector_ldap.id`.

dannysauer:
This is verifying that what we expected to store is returned. That would be 
verifying that was was returned is returned. :)

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -434,97 +434,107 @@
...
-          :"eviction-hard"     => ""
+    def expected_dex_json(num, certificate, c_name, host)
+      {
+        type:            "ldap",
+        id:              num,
+        name:            c_name,
+        root_ca_data:    Base64.encode64(certificate.certificate),
+        bind:            {
+          anonymous: false,
+          dn:        "cn=admin,dc=#{host.chomp(".com")},dc=com",
```
nanoscopic: If the already passed in 'num' matches the id, as it is still constructing a json object checking that, then simply using number and constructing the hostname in place is much cleaner. So long as 'num' is available there to use, I fail to see how this is an improvement to the expected_dex_json function.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -434,97 +434,107 @@
...
-  # rubocop:disable RSpec/ExampleLength
-  context "with dex LDAP connectors tls" do
-    it "has dex LDAP connectors" do
-      dex_connector_ldap = create(:dex_connector_ldap, :tls, :regular_admin)
-      CertificateService.create(service: dex_connector_ldap, certificate: certificate)
+  context "with Dex LDAP connectors" do
+    before do |example|
+      # calculate a likely-unique 0-100 integer hash value by adding up the
+      # integer values of chars in the example name (context + example)
+      u = example.metadata[:full_description].codepoints.reduce(0, :+) % 100
```
nanoscopic: A much larger range needs to be used; supposing this number is depended on it could produce false positives during error conditions. 1-100 is not an adequate range. Also; as discussed in the other comment, if 'num' is used then this is avoided. This before block makes the code more confusing and harder to read. Please change it back and use 'num'.

To my understanding the only actual fix in this commit is changing the expect blocks to be after the get lines. These changes to the way the json is checked don't seem to be required to make the needed change.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
nanoscopic: The attempt here is to avoid using the sequence for the names; there is a conflict where the sequence and db ids don't match up. Really; if this change is done the setting of the host should be removed from the factory and/or a comment added there as that is the purpose of that code being there.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -434,97 +434,107 @@
...
+              expected_dex_json(dex_connector_ldap.id, certificate, @conn_name, @host_name).merge(
+                server:    "#{@host_name}:636",
+                start_tls: false
+              )
+            ]
+          }
+        }
+        get :show
         expect(json).to eq(expected_json)
-        delete(dex_connector_ldap)
```
nanoscopic: This is left in the test database and effects later tests if this line is removed. It should not be removed.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
dannysauer: While 100 should be more than sufficient to handle not only the input domain here (two strings) but also the foreseeable future of "maybe a couple more" test cases, an established digest function like MD5 is somewhat less likely to collide given similar-length ASCII strings.  I didn't realize that Digest::MD5 is in the standard library, so it's easy enough to switch over.

dannysauer:
As we go farther down the road of using the factory-generated object for data, we get closer to validating only the format of the response and not the data content. While that's fine for what this test ought to be doing, is that still a sufficient test with the other connector tests that are defined elsewhere? In other words, Is the model being fully exercised elsewhere such that it's ok to basically validate only the JSON structure here? My initial update did not make that assumption.

vitoravelino:
> As we go farther down the road of using the factory-generated object for data, we get closer to validating only the format of the response and not the data content. While that's fine for what this test ought to be doing, is that still a sufficient test with the other connector tests that are defined elsewhere? In other words, Is the model being fully exercised elsewhere such that it's ok to basically validate only the JSON structure here? My initial update did not make that assumption.

The spec changes you've submitted belongs to the `PillarsController`. The controller has a single endpoint and its only purpose is return a JSON. Our responsibility in this context is to test whatever affects the return of that JSON.

dannysauer:
> The spec changes you've submitted belongs to the `PillarsController`. The controller has a single endpoint and its only purpose is return a JSON. Our responsibility in this context is to test whatever affects the return of that JSON.

Agreed.  With commit 2dba58d I've moved all of the JSON assembly logic into the expected_json definition.  That reduced the content of the test examples to merely varying by a couple of constant parameters, so I pulled those out into a loop.  The code has less repetition now.  As a side benefit, it also now tests all four possible combinations of bind method and security negotiation rather than just two.

I also agreed with @nanoscopic's reasoning and elected to leave the destroy call in as a comment, so some future person who finds a data persistence issue might have a starting point.  I think that (and the above) addresses all of the requested changes.

If we're good here, I'll squash the commits in the branch and repush.

dannysauer:
Oh; I didn't switch to the `let` block because I don't see it adding anything here.  The object changes across every example and is used in every example, so the lazy evaluation doesn't save anything.  The example is so compact that, IMHO, the parens and colons and braces in `let` add more distracting visual clutter than simply using an `=` for a local var.

spec/controllers/internal_api/v1/pillars_controller_spec.rb
```diff
@@ -434,97 +434,70 @@
...
+      [:regular_admin, :anon_admin].each do |bind_method|
+        context "with #{security}" do
+          it "has dex LDAP connectors for bind as #{bind_method}" do
+            dex_connector_ldap = create(:dex_connector_ldap, security, bind_method)
+            CertificateService.create(service: dex_connector_ldap, certificate: certificate)
+
+            get :show
+            expect(json).to eq(expected_dex_json(dex_connector_ldap, certificate))
+            # destroy should be unecessary, but here it is if that assumtion is invalidated
+            # dex_connector_ldap.destroy
```
vitoravelino: You can safely remove these lines. No need to worry about that.

dannysauer: Removed, and squashed commits.

PR #680, opened 2018-10-18
Add database index on alter_time column (bsc#1111787)

mjura:
Add new index to velum_production database on table salt_events on column
alter_time. This will improve cleanings for old salt events and salt returns.


mjura:
We can close it, there was bug which was fixed in https://github.com/kubic-project/velum/pull/688

PR #681, opened 2018-10-18
[3.0] ui: don't show failed alert for repeated orchestration

MaximilianMeister:
A failed alert is shown whenever one of the nodes fails to bootstrap.
This is an optimization to only show this alert once for a specific
orchestration run. If the user closes the alert, it won't come back
unless a new orchestration happens.

bsc#1097752

(cherry picked from commit e943e073e813cdea58bf0943e1caba226cb60131)

Backoprt of https://github.com/kubic-project/velum/pull/589

### DIFF
--- a/app/assets/javascripts/dashboard/dashboard.js
+++ b/app/assets/javascripts/dashboard/dashboard.js
@@ -149,6 +149,7 @@ MinionPoller = {
         }
 
         State.minions = minions;
+        State.lastOrchestrationAt = data.last_orchestration_at;
 
         var pendingStateMinion = minions.find(function (minion) {
           return minion.highstate == "pending";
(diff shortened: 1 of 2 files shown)


dannysauer (commented):
This uses the `created_at` field from the database, which is a `datetime` type rather than a `timestamp`.  While the odds are good that no one will create a second deployment at 1AM on the day when DST ends, I'm pretty sure this introduces a (trivial) race condition *if* the customer has also changed their time zone from the default UTC (by default MySQL/Maria uses the system timezone).

How much would break if the column type was changed?

ereslibre:
> How much would break if the column type was changed?

As far as I can tell MariaDB should not be mapping timezones at all. We control how we run `MariaDB` and we should not be setting any timezone in it. On top of that, ActiveRecord should be using UTC for all our usages (because we are not mapping them either), so I don't see a timezone shift in any part of the stack (we can ask for the customer timezone in the future and show events on the views to their local time, but that's up on the stack, `MariaDB` and `ActiveRecord` should be storing everything in plain UTC).

By default Rails creates `created_at` and `updated_at` as `datetime` instead of `timestamp`. I think it would be good if we could continue using Rails' defaults.

ereslibre (approved):
LGTM

ereslibre:
Not adding the `awaiting-merge` label until @dannysauer confirms that everything is fine regarding their objections.

PR #700, opened 2018-11-28
Support for cloud node removal

jgleissner:
This adds support for removing cluster nodes in the public cloud. It uses salt-cloud to terminate the instance associated with the given minion, and uses a salt event processor to react on the salt-cloud destroyed event to remove the minion from the database.

This fixes https://bugzilla.suse.com/show_bug.cgi?id=1117152.


jgleissner:
This requires https://github.com/kubic-project/caasp-container-manifests/pull/207 to work properly.

lib/velum/salt.rb
```diff
@@ -72,11 +72,25 @@ def self.accept_minion(minion_id: "")
...
+      res = perform_request(endpoint: "/", method:   "post",
+                            data:     { client: "local",
+                                        tgt:    "admin",
+                                        fun:    "cloud.query" })
+      if res.body.include?(minion_id)
+        res = perform_request(endpoint: "/", method:   "post",
+                              data:     { client: "local_async",
+                                          tgt:    "admin",
+                                          fun:    "cloud.destroy",
+                                          arg:    [ minion_id ] })
```
flavio: I looked at the salt documentation and it's not clear to me whether this API call is going to drop the minion key as well.

I wonder if we shouldn't always execute the other branch of this `if/else` statement.

lib/velum/salt.rb
```diff
@@ -72,11 +72,25 @@ def self.accept_minion(minion_id: "")
...
+      else
+        res = perform_request(endpoint: "/", method: "post",
+                              data: { client: "wheel",
+                                      fun:    "key.delete",
+                                      match:  minion_id })
+      end
+      JSON.parse(res.body)
```
flavio: This changed from parsing the `res.code` to parsing the whole body of the response. Doesn't it have side effects in other parts of the code?

flavio (commented):
I didn't test the code with these changes. I'm not sure our CI has automated tests in place simulating node removal, that's something we should test to ensure no regression is introduced.

lib/velum/salt.rb
jgleissner: `cloud.destroy` does remove the minion key.

lib/velum/salt.rb
jgleissner: I don't think the return code is used at the moment, and actually even couldn't, as the existing code `JSON.parse(res.code)` throws an exception since `res.code` is not JSON (see also https://bugzilla.suse.com/show_bug.cgi?id=1117152).

MalloZup:

PR #706, opened 2018-12-21
Include the name on the Pillar API response for the registry list

ereslibre:
This allows us to include the descriptive name the user gave to each
registry, so when creating the resources on the cluster (specially when
we create custom resources) we can give the custom resources meaningful
names.


MalloZup:

PR #707, opened 2019-01-08
[WIP] Added External Certificates Feature

danielorf:
Added External Certs controller to Settings; Added associated views, routes and models
Added ext_cert_feature_spec.rb, added example certs/keys to spec/fixtures


app/assets/stylesheets/pages/settings.scss
```diff
@@ -110,3 +110,27 @@ $sidebar-item-highlight: #00c081;
 .ldap-conn-message-valid {
   color: $brand-success;
 }
+
+.cert-table {
+  text-indent: 10px;
```
vitoravelino: You can remove this because we are going to use `word-break` and `width` that will align stuff better.

app/assets/stylesheets/pages/settings.scss
```diff
@@ -110,3 +110,27 @@ $sidebar-item-highlight: #00c081;
...
+  td.field {
+    font-weight: bold
```
vitoravelino: Add `width: 180px`

app/assets/stylesheets/pages/settings.scss
```diff
@@ -110,3 +110,27 @@ $sidebar-item-highlight: #00c081;
...
+  }
+}
```
vitoravelino: Add `td.value` selector with `word-break: break-word`.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,138 @@
...
+            .cert-element 
+              <strong>#{k}</strong>:  <span class=#{textcolor}>#{v}</span>
+      - else
+        .cert-table
+          p
+          | No Velum SSL Key found.  Please add an SSL key.
+      br
+      .form-group.form-group-certificate
+        p Select "Choose File" to upload a new Velum SSL Key:
+        = f.file_field :velum_key, class: "form-control", required: false
```
vitoravelino: Remove `form-control` from all the file inputs.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,138 @@
...
+
+h2 External Certificates
+
+p Enable the use of External SSL Certificates for CaaSP and Kubernetes services.  Certificates and matching keys must be in armored PEM format.
+.alert.alert-warning role='alert'
+ p Remember to add a Root CA Certificate to the System Wide Certificates corresponding to any new External Certificates
+
+= form_for :external_certificate, url: settings_ext_cert_index_path, method: :post do |f|
+  br
+  .panel.panel-default
```
vitoravelino: Extract all the panels to its own partial.

app/assets/stylesheets/pages/settings.scss
vitoravelino: Add `vertical-align: top`

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,254 @@
...
+    @velum_cert = cert_parse(Pillar.value(pillar: :external_cert_velum_cert))
+    @velum_key = key_parse(Pillar.value(pillar: :external_cert_velum_key))
+    @kubeapi_cert = cert_parse(Pillar.value(pillar: :external_cert_kubeapi_cert))
+    @kubeapi_key = key_parse(Pillar.value(pillar: :external_cert_kubeapi_key))
+    @dex_cert = cert_parse(Pillar.value(pillar: :external_cert_dex_cert))
+    @dex_key = key_parse(Pillar.value(pillar: :external_cert_dex_key))
+  end
+
+  def get_val_from_form(form, param)
+    if params[form][param].present?
```
vitoravelino: Isn't better if we hard code `params[:external_certificate]` since it will always be the same param prefix? it will make the call shorter. Or do we plan to pass other params outside `:external_certificate`?

spec/features/settings/ext_cert_feature_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+  describe "#index" do
+    before do
+      visit settings_ext_cert_index_path
+    end
+
+    # Success Conditions
+
+    it "barely does the bare minimum" do
+      expect(page).to have_http_status(:success)
+      expect(page).to have_content("External Certificates", wait: 1)
```
vitoravelino: You can remove all the `wait: 1`.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,254 @@
...
+    return OpenSSL::PKey::RSA.new(key_string)
+  rescue OpenSSL::PKey::RSAError, NoMethodError, TypeError
+    # Push error handling to calling method for flexibility
+    return nil
+  end
+
+  # Common method to simplify failure renders
+  def render_failure_event(message)
+    set_instance_variables
+    flash[:notice] = message
```
vitoravelino: Use `:alert` instead of `:notice` since it's an error.

app/controllers/settings/ext_cert_controller.rb
vitoravelino: Also, if the user submits the form without selecting any file, a nil exception occurs.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,138 @@
...
+          h5 Current Velum SSL Certificate Info:
+          table
+            - @velum_cert.each do |k, v|
+              tr
+                td.field = k.to_s
+                td.value = v.to_s
+      - else
+        .cert-table
+          p
+          | No Velum certificates found.  Please add a certificate.
```
vitoravelino: You can simply write `p No Velum certificates found.  Please add a certificate.`. Also applies to the other occurrences.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,138 @@
...
+      br
+      .form-group.form-group-certificate
+        p Select "Choose File" to upload a new Velum SSL Certificate:
+        = f.file_field :velum_cert, class: "form-control", required: false
+      br
+      - if @velum_key
+        .cert-table
+          h5 Current Velum SSL Key Info:
```
vitoravelino: `h5` is `14px` and `h4` is `18px`. I'd say `16px` would be better to highlight more the sections. You can add a class and set `16px`.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,138 @@
...
+  br
+  .panel.panel-default
+    .panel-heading
+      h3.panel-title Velum SSL Certificate and Key
+    .panel-body
+      - if @velum_cert
+        .cert-table
+          h5 Current Velum SSL Certificate Info:
+          table
+            - @velum_cert.each do |k, v|
```
vitoravelino: I wouldn't show `Error Certificate not available, please upload a certificate`. Maybe showing only `Certificate not available, please upload a certificate`. `Error` is definitely not a good word to be used.

vitoravelino: I would probably use a simple `p` instead of the `table`.

app/assets/stylesheets/pages/settings.scss
danielorf: Done

app/assets/stylesheets/pages/settings.scss
danielorf: Done

app/views/settings/ext_cert/index.html.slim
danielorf: Done

app/controllers/settings/ext_cert_controller.rb
danielorf: Hard-coding `:external_certificate` is fine with me.

spec/features/settings/ext_cert_feature_spec.rb
danielorf: The current wait time on RSpec fails is terribly long right now.  Obviously not a problem when there are no failures though.  Is there a reason the global wait value is so long?

app/controllers/settings/ext_cert_controller.rb
danielorf: Done

spec/features/settings/ext_cert_feature_spec.rb
vitoravelino: This is some legacy decision when we had issues writing capybara specs. We are safe to to reduce the global timeout to 10-30 seconds.

app/views/settings/ext_cert/index.html.slim
danielorf: I made it `Notice` instead.

danielorf:
> Regarding the mirrors/registries specs failing, you can remove `expect(page).not_to have_content("Certificate")` lines that it's causing that. No big deal.

I had to change a few tests (mirrors_feature_spec and registries_feature_spec) because they were looking for "Certificate" on the page which contains the word twice in the sidebar.

app/views/settings/ext_cert/index.html.slim
danielorf: Done

app/assets/stylesheets/pages/settings.scss
danielorf: It's not clear to me what needs to be removed.  We use ".cert-table" below.  Are you referring to the "text-indent"?

danielorf: Done

app/views/settings/ext_cert/index.html.slim
danielorf: I presume you're looking for something like this:
```
  .panel
    .panel-default
      .panel-heading
        h3.panel-title Dex SSL Certificate and Key
```

vitoravelino: No. It's about extracting the content to a new file and use `= render "settings/ext_cert/_partial_name"`. This way the templates become more readable. 

Remember to pass the form reference as a parameter.

app/controllers/settings/ext_cert_controller.rb
danielorf: Nil exception fixed

spec/features/settings/ext_cert_feature_spec.rb
danielorf: OK.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,135 @@
...
+h2 External Certificates
+
+p Enable the use of External SSL Certificates for CaaSP and Kubernetes services.  Certificates and matching keys must be in armored PEM format.
+.alert.alert-warning role='alert'
+ p Remember to add a Root CA Certificate to the System Wide Certificates corresponding to any new External Certificates
+
+= form_for :external_certificate, url: settings_ext_cert_index_path, method: :post do |f|
+  br
+  .panel
+    .panel-default
```
vitoravelino: When extracting to partials, do not forget to revert this to `.panel.panel-default`.

app/views/settings/ext_cert/index.html.slim
vitoravelino: screenshot-20190109143716-523x85

This is the problem that I've mentioned that was going to happen if we don't differentiate the way to display the content.

app/assets/stylesheets/pages/settings.scss
```diff
@@ -110,3 +110,35 @@ $sidebar-item-highlight: #00c081;
 .ldap-conn-message-valid {
   color: $brand-success;
 }
+
+.cert-table {
+  td.field {
+    font-weight: bold;
+    width: 180px;
+    vertical-align: top;
```
vitoravelino: Could you add `padding-bottom: 3px` here?

vitoravelino:
screenshot-20190109145254-352x84

screenshot-20190109145455-232x67

I'm fine with `Valid Key: true` and `true` being green but the other case should be black.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,135 @@
...
+          p Select "Choose File" to upload a new Velum SSL Certificate:
+          = f.file_field :velum_cert, required: false
+        br
+        - if @velum_key
+          .cert-table
+            h5.cert-parse-heading Current Velum SSL Key Info:
+            - @velum_key.each do |k, v|
+              - if v
+                - textcolor = "ext-cert-success"
+              - else
```
vitoravelino: Is there a way to reproduce this scenario?

app/views/settings/ext_cert/index.html.slim
danielorf: I'm currently using the same mechanism to inject an error message [here](https://github.com/danielorf/velum/blob/external_certs_front_end/app/controllers/settings/ext_cert_controller.rb#L163).  Do we need to make sure this unlikely state has the ideal spacing?  If so, I will have to add logic to the slim code looking for each of these states - and I'm hesitant to spread the logic across multiple files if I can keep it in the controller.

app/assets/stylesheets/pages/settings.scss
danielorf: Done

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+    @velum_key = key_parse(Pillar.value(pillar: :external_cert_velum_key))
+    @kubeapi_cert = cert_parse(Pillar.value(pillar: :external_cert_kubeapi_cert))
+    @kubeapi_key = key_parse(Pillar.value(pillar: :external_cert_kubeapi_key))
+    @dex_cert = cert_parse(Pillar.value(pillar: :external_cert_dex_cert))
+    @dex_key = key_parse(Pillar.value(pillar: :external_cert_dex_key))
+  end
+
+  def get_val_from_form(param)
+    if params.key?(:external_certificate) && params[:external_certificate].key?(param)
+      params[:external_certificate][param].read.strip
```
nanoscopic: I don't think the extra 'strip' belongs. The direct contents of the file should be used. The strip was there originally in the code this was copied from to handle paste issues. The same will not be true for a file upload.

Also, what happens if a zero line file is uploaded? Am curious.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+        }
+      }
+    }
+  end
+
+  # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+  def upload_validate(key_cert_map)
+    # Do nothing if both cert/key are empty
+    if key_cert_map[:cert][:cert_string].empty? && key_cert_map[:key][:key_string].empty?
+      true
```
nanoscopic: An early return without return looks really terrible. I would advise to override rubocop on this as it looks nutty.

I can see why it is the case, because it is the last thing evaluated, but because the code below goes down so far it is not immediately obvious that is the last evaluated value.

At the very least a 'true # return true' is in order.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+  end
+  # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+  def cert_parse(cert_string)
+    params = {}
+
+    # Check if certificate exists, assume that validation has already occured
+    if !cert_string
```
nanoscopic: Doesn't ruby style prefer unless instead of 'if !' ?

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+      params["Not Valid After".to_sym] = cert.not_after
+      params[fingerprint[0]] = fingerprint[1]
+    end
+    params
+  end
+
+  def cert_fingerprint(cert)
+    # Error-checking ignored, cert already validated and verified
+    fingerprint = OpenSSL::Digest::SHA256.new(cert.to_der)
+    fingerprint_string = fingerprint.to_s.scan(/../).map(&:upcase).join(":")
```
nanoscopic: This needs a comment saying "Add colons every 2 characters" or something similar

Also, is there no ruby string 'to_upper'? Using map(&:upcase) is a really goofy way to uppercase.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+  end
+
+  def key_parse(key_string)
+    params = {}
+    if !key_string
+      params[:Error] = "Key not available, please upload a certificate"
+    # :nocov:
+    # Key has aleady been valiated when entered, an error here would require a failure
+    # in connection from velum to salt or an unintended change in the salt pillar
+    else
```
nanoscopic: This whole block has no coverage? Isn't is possible to test by passing in a key that cannot be parsed?

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+      params["Valid Key"] = key_valid
+      # :nocov:
+    end
+    params
+  end
+
+  # Common method to build cert object from string
+  def read_cert(cert_string)
+    return OpenSSL::X509::Certificate.new(cert_string)
+  rescue OpenSSL::X509::CertificateError, NoMethodError, TypeError
```
nanoscopic: I'd recommend throwing a new exception rather than returning nil here. Something like 'CertCantRead'. You would need to define the new exception type but I think it would look cleaner. Maybe it isn't worth the effort though. Unsure.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+      else
+        false
+      end
+      params["Valid Key"] = key_valid
+      # :nocov:
+    end
+    params
+  end
+
+  # Common method to build cert object from string
```
nanoscopic: "Common method" can be dropped from the description. "Builds cert object from string" is good and succinct.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+    # Push error handling to calling method for flexibility
+    return nil
+  end
+
+  # Common method to simplify failure renders
+  def render_failure_event(message)
+    set_instance_variables
+    flash[:alert] = message
+    render action: :index, status: :unprocessable_entity
+    false
```
nanoscopic: Since this always returns false, I'm not sure what the point is here of returning anything.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+  end
+
+  # Get SubjectAltName field buried in a certificate
+  def get_san_array(cert)
```
nanoscopic: get_altnames_array would be much easier to read and understand.

app/controllers/settings/ext_cert_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+    subject_alt_name = cert.extensions.find { |e| e.oid == "subjectAltName" }
+    return nil unless subject_alt_name
+    subject_alt_name.value.gsub("DNS:", "").delete(",").split(" ")
+  end
+
+  def cert_date_check(cert)
+    if Time.now.utc > cert.not_after || Time.now.utc < cert.not_before
+      message = "Certificate out of valid date range"
+      render_failure_event(message)
```
nanoscopic: As mentioned above; it would be cleaner here to call the render_failure_event and then explicitly return false. It's unclear false is being returned otherwise.

app/views/settings/ext_cert/index.html.slim
```diff
@@ -0,0 +1,135 @@
...
+
+  .panel
+    .panel-default
+      .panel-heading
+        h3.panel-title Kubernetes API SSL Certificate and Key
+      .panel-body
+        - if @kubeapi_cert
+          .cert-table
+            h5.cert-parse-heading Current Kubernetes API SSL Certificate Info:
+            table
```
nanoscopic: Using a table here is good. It keeps the multiple data points from mushing together in a messy visual way.

config/routes.rb
```diff
@@ -75,6 +75,7 @@
     resources :ldap_test
     resources :dex_connector_ldaps, path: :ldap_connectors
     resources :dex_connector_oidcs, path: :oidc_connectors
+    resources :ext_cert, only: [:index, :create]
```
nanoscopic: ext_cert looks kind of ugly. Would be great to changed it everywhere to 'external_cert'. This is a personal preference thing, as people will agree/disagree on whether 'ext' clearly indicates 'external' versus 'cert' clearly being 'certificate'. I think it is worth changing for clarity. 'cert' is fine in my mind because it is very obvious. 'ext' could also be 'extend'

spec/features/settings/ext_cert_feature_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+  end
+
+  describe "#index" do
+    before do
+      visit settings_ext_cert_index_path
+    end
+
+    # Success Conditions
+
+    it "barely does the bare minimum" do
```
nanoscopic: 'bare' essentially used twice here; 'does the bare minimum' would be sufficient

spec/features/settings/ext_cert_feature_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      expect(page).to have_http_status(:success)
+      expect(page).to have_content("External Certificates")
+    end
+
+    it "saves the form with nothing attached" do
+      click_button("Save")
+      expect(page).to have_http_status(:success)
+      expect(page).to have_content("External Certificate settings successfully saved.")
```
nanoscopic: This string is checked against manually repeatedly. A const for the string being checked against would be good.

spec/features/settings/ext_cert_feature_spec.rb
```diff
@@ -0,0 +1,233 @@
+require "rails_helper"
+
+# rubocop:disable RSpec/ExampleLength
+describe "Feature: External Cerificate settings", js: true do
+
+  let!(:user) { create(:user) }
+  let!(:fixture_path) { RSpec.configuration.fixture_path }
+
+  let(:ssl_cert_file_a) { File.join(fixture_path, "ext_cert_ssl_a.pem") }
```
nanoscopic: What these different files indicate should be explained in a comment, and something more descriptive than a/b would be good. Like "valid/invalid"

spec/fixtures/expired_cert.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
nanoscopic: It would be nice to have a expired_cert.pem.to_create.txt file that has the commands used to generate the pem cert.

app/controllers/settings/ext_cert_controller.rb
danielorf: Agree completely, but as you noted this is what rubocop wants.  I'll add the `# return true` comment.

app/controllers/settings/ext_cert_controller.rb
danielorf: Apparently this is only true with a hanging `if` statement.  If the `else` block is included, rucocop expects `if`.  This is the error message from rubocop if you use `unless` with `else`:
`Do not use unless with else. Rewrite these with the positive case first.`

app/controllers/settings/ext_cert_controller.rb
danielorf: Adding descriptive comment.  `upcase` is the way to go:  https://ruby-doc.org/core-2.2.3/String.html#method-i-upcase
I also rearranged the command to get rid of `map` and make more readable.

app/views/settings/ext_cert/index.html.slim
vitoravelino: Unfortunately the logic must go somewhere in this case unless you see a way to differentiate them without adding that. We could differentiate using a css class but to define the css class we need some logic. So it's inevitable. You can use helpers to do that instead of doing it in the template or controller. That's the concept behind rails helpers.

app/controllers/settings/ext_cert_controller.rb
danielorf: It shouldn't be possible to upload an unreadable key.  This just protects for the unlikely scenario where the key is missing or corrupted after storing.

spec/features/settings/ext_cert_feature_spec.rb
danielorf: Done

spec/features/settings/ext_cert_feature_spec.rb
danielorf: Done

spec/features/settings/ext_cert_feature_spec.rb
danielorf: Done

app/views/settings/ext_cert/index.html.slim
danielorf: Ideally no.  It shouldn't be possible to upload an unreadable key. This just protects for the unlikely scenario where the key is missing or corrupted after storing.  I removed the coloring for now.  Unnecessary complication.

config/routes.rb
danielorf: Sounds good to me.  Done.

app/views/settings/ext_cert/index.html.slim
danielorf: Done

app/views/settings/ext_cert/index.html.slim
danielorf: I fixed this at the expense of making the key section a table:
image

app/controllers/settings/ext_cert_controller.rb
danielorf: Fair enough.  .strip on an empty string is safe.

danielorf:
@vitoravelino All of your comments should be addressed now.

app/views/settings/external_cert/_form.html.slim
```diff
@@ -0,0 +1,48 @@
+  .panel
+    .panel-default
```
vitoravelino: You forgot to revert it to `.panel.panel-default`.

app/views/settings/ext_cert/index.html.slim
vitoravelino: The key info was fine for me, the problem was with the case of `Certificate not available, please upload a certificate` that you've fixed that and looks great.

Do we display any other info besides "valid key" for the keys to treat this data as a table? Couldn't we simply display "Your key is valid"? Why is it important to have it displayed like that? If there's other info to be displayed I'll understand to have it as tabulated data, otherwise revert it and kill the spacing for key as you did for the not uploaded file.

app/views/settings/external_cert/_form.html.slim
danielorf: You are right.  Done now.

danielorf:
I am stripping out the `expect(page).to have_http_status(:success)` from the tests where this comes first.  There appears to be a race condition where Capybara sometimes detects the redirect (30X) and not the following successful http status (20X).

spec/fixtures/expired_cert.pem
danielorf: This involved changing the system time on my workstation, I couldn't get any alternative to work.  Not the sort of thing you want to automate.  This will get solved when we include a lib for CA/cert/key generation in a subsequent PR.

danielorf:
Screenshot of current state with a sample cert uploaded for Velum:
image

danielorf:
Moving this work to a feature branch ([feature_external_certs](https://github.com/kubic-project/velum/tree/feature_external_certs)) to better manage commits from multiple users and keep history clean.

PR #708, opened 2019-01-08
Expose pillars managing kubernetes extra args

flavio:
The salt backend already has pillars that can be used to provide extra
arguments to the different kubernetes components (api-server,
controller-manager, scheduler, kubelet and proxy).

This commit allows to provide values for these settings. This is
something we should **not** expose into the UI. It can be handy with
certain customers who are looking for a way to pass not-yet-handled
flags to kubernetes.

Signed-off-by: Flavio Castelli <fcastelli@suse.com>

feature#hidden-kubernetes-pillars

(cherry picked from commit 62237bef06bc0cbb3f67320d7cdb95045b00ccdf)


flavio:
ping @jordimassaguerpla to figure out how to push this out

jordimassaguerpla:
Hi,

first of all we need a bug in bugzilla. Could you create one?

flavio:
@jordimassaguerpla  I've created https://bugzilla.suse.com/show_bug.cgi?id=1121459 and added the proper bsc reference inside of the commit message

PR #710, opened 2019-01-10
remove hack, not anymore needed

MalloZup:
backport from master


jordimassaguerpla:
Hi @MalloZup ! I don't know about this... why are you removing the "gem update" command? If you could be more specific I may be able to review it :)

PR #711, opened 2019-01-10
Remove gem-install workaround

MalloZup:
# Description:

this is a backport from https://github.com/kubic-project/velum/pull/709

 Don't run gem update --system

Related to `gem update --system`:
- travis-ci/travis-ci#8978
- rubygems/rubygems#2123

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ rvm:
   - 2.1.9
 
   # Other ruby versions
-  - 2.2.7
+  - 2.3.0
   - 2.3.4
   - 2.4.1
 
@@ -29,7 +29,6 @@ cache:
     - $PWD/travis_phantomjs
 
 before_install:
-  - gem update --system
   - phantomjs --version
   - export PATH=$PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
   - if [ $(phantomjs --version) != '2.1.1' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi


MalloZup:
@hwoarang  i was asked to the backport. Initially the fix was in `master`.

You can see https://github.com/travis-ci/travis-ci/issues/8978 and in this commit description https://github.com/plataformatec/devise/commit/8e49661fd167eed7a97281aa09aea7bc6bfd1bf5

hwoarang:
> @hwoarang i was asked to the backport. Initially the fix was in `master`.
> 
> You can see [travis-ci/travis-ci#8978](https://github.com/travis-ci/travis-ci/issues/8978) and in this commit description [plataformatec/devise@8e49661](https://github.com/plataformatec/devise/commit/8e49661fd167eed7a97281aa09aea7bc6bfd1bf5)

I understand this is a backport but that does not mean the commit message has to be cryptic. Since we can't fix the commit message on `master` anymore lets fix this instead.

I would prefer it if you amend you commit message to explain the problem and the solution.

MalloZup:
@hwoarang ok, yop you are right for that. I will fix that. thx

MalloZup:
one thing to investigate, is why this is failing only on 3.0 and not on head. https://travis-ci.org/kubic-project/velum/jobs/477778826

MalloZup:
Yep sure. Before squashing I think we need agreement what we do with the version, should we increment it and do the same in master? Is OK for everybody?

flavio:
I have nothing against dropping this version of ruby, as far as I can see both v3 and v4 are using ruby 2.1.9, which is still being tested by travis.

Maybe others more involved with velum can express their opinion on the topic?

MalloZup:
@ereslibre @MaximilianMeister if it is of for you, we can stick to that version and before merging this PR i will send also a PR to `head` branch so we have 1:1 along branches. TIA. :motor_scooter:

MaximilianMeister:
We still have ruby `2.1.9` in v4 so we don't need to test `2.2` i'd say

MalloZup:
thx @MaximilianMeister feel free to merge theb tgx

PR #715, opened 2019-01-18
[3.0][CPI] Fix regression and re-add option to CPI, bsc#1122439

MaximilianMeister:
This change is reverting ignore_vol_az option back to Velum CPI.


hwoarang:
I am sorry to be this annoying person again, but can you add `This change is reverting ignore_vol_az option back to Velum CPI.` to the commit message as well.

abonillasuse:
@flavio can we please ensure this is pushed out this week?

bergmannf:
Some additional changes were requested in the bug that would impact the salt-states and add an additional layer of security for this feature: https://bugzilla.suse.com/show_bug.cgi?id=1122439

PR #719, opened 2019-01-29
WIP: setup chromium headless for velum instead of phantomjs

MalloZup:
no worth for reviwiing atm 

#### Todos:

- [ ] enable on local-devenv mode too

### DIFF
--- a/.ci/chromium-setup.sh
+++ b/.ci/chromium-setup.sh
@@ -1,0 +1,27 @@
+#! /bin/bash
+
+set -e
+# this script install chromium headless for a debian system used in CI.
+
+apt-get update
+apt-get install -y unzip libnss3-tools
+
+# download chromedriver
+CHROME_DRIVER_VERSION="2.35" #ChromeDriver 2.35 -> Supports Chrome v62-64 ( used from travis)
+
+wget https://chromedriver.storage.googleapis.com/$CHROME_DRIVER_VERSION/chromedriver_linux64.zip
+unzip chromedriver_linux64.zip
+mv chromedriver /usr/bin/chromedriver
+chmod +x /usr/bin/chromedriver
+
+
+# create needed pki for chrome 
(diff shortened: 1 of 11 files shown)


.ci/chromium-setup.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
+
+set -e
+# this script install chromium headless for a debian system used in CI.
+snap install chromium
```
vitoravelino: Doesn't adding 

```
addons:
  chrome: stable
```

fix that?

MalloZup: @vitoravelino i'm trying to use `chromium` which is more updated atm.

Anywyas is kind wip, i need to find out the repository of latest chromium for ubuntu so it's compatible with the latest chrome-driver :racehorse:

MalloZup: thx @vitoravelino anyways for that suggestion

vitoravelino: No problem. :wink:

MalloZup:
@ereslibre seems that Travis kill the job because to much time. Can you run this locally and post the output here if you have 3 min time? I think locally the job will not killed. Chrome and the driver are good setup but I cannot see where it is failing.

vitoravelino:
Try to decrease the capybara default wait time. It's probably that.

MalloZup:
Thx Vito for hint. 🎸

bear454:
Maybe you can derive some value from https://github.com/openSUSE/osem/pull/2225/files .

MalloZup:
This GitHub PR is unactive since more then 30 days. Is this GitHub PR still needed? Please close or update it accordingly.

thehejik:
@MalloZup I don't think we need it, please close

MalloZup:
@thehejik  yes, i think we can close thank you for your help here tomas. 

thx all for feedback and help.

PR #721, opened 2019-01-30
Certificate Chain Checking for External Certificates

nanoscopic:
Added full certificate chain checking. Created all in one unified certificate generation bash to create the certs used by the spec tests for external certificates.


spec/fixtures/external_certs/badalt.crt
```diff
@@ -0,0 +1,72 @@
+Certificate:
```
nanoscopic: It is notably interesting that the system accepts certificates with the clean textual version of the certificate at the top. This is a common pattern for certificates but I had previously thought that this is not allowed. The documentation on what is acceptable may need to be adjusted to mention this.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
+#!/bin/bash
+
+# This script re-generates all of the keys and certs within the external_certs directory.
+# The files there are used by the spec/features/settings/external_cert_feature_spec test
+
+function init() {
+  mkdir -p /tmp/ca_gen/csr
```
nanoscopic: It may make more sense to put all the mkdir commands below the variable setting, and mkdir based on the variables, rather than duplicating the paths.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  mkdir -p /tmp/ca_gen/conf
+  DIR_CONF=/tmp/ca_gen/conf
+  
+  mkdir -p /tmp/ca_gen/keys
+  DIR_KEYS=/tmp/ca_gen/keys
+  
+  mkdir -p /tmp/ca_gen/cadata
+  DIR_CADATA=/tmp/ca_gen/cadata
+  
+  echo "dir cadata: $DIR_CADATA"
```
nanoscopic: This looks to be a debugging output It should be removed.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+
+function init_ca() { # $1 ca_name
+  echo "Initiating CA  \"$1\""
+  generate_conf $1
+  # Create / clear certs directory for CA
+  mkdir -p $DIR_CADATA/$1/db.certs
+  rm $DIR_CADATA/$1/db.certs/* 2>/dev/null
+  
+  # Create / clear index file of certs
+  rm $DIR_CADATA/$1/db.inde* 2>/dev/null
```
nanoscopic: It should be mentioned why this is db.inde*, and not db.index; it is because this actually deleted two files. It deletes both db.index and db.index.attr

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  touch $DIR_CADATA/$1/db.index
+  
+  # Other basic needed files
+  echo unique_subject = yes > $DIR_CADATA/$1/db.index.attr
```
nanoscopic: In some cases the db.index.attr gets automatically created. In others it does not. Unsure exactly why it doesn't get created in some cases so this is here to ensure it exists. This is the default content of the file when it does get automatically created.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  echo 01 > $DIR_CADATA/$1/db.serial
```
nanoscopic: The serial can start at 1 for every CA. The serial numbers of not GUIDs; they are specific to the issuing CA, so this is fine.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  openssl req -sha256 -new -key $DIR_KEYS/ca_$1.key -out $CSR_DIR/ca_$1.csr -subj "$3"
+  openssl ca -batch \
+    -config $DIR_CONF/ca_$2.conf \
+    -extensions v3_ca \
+    -out $DIR_CERTS/ca_$1.crt \
+    -infiles $CSR_DIR/ca_$1.csr 2>/dev/null
+}
+
+function append_altnames() { # $1 altnames, $2 filename
+  IFS='-' read -r -a altnames <<< "$1"
```
nanoscopic: This should have a comment explaining that this just splits a dash delimited string into an array.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  cat >>$1 <<EOL
+[req_distinguished_name]
+EOL
+  
+  echo -e "[req_ext]\nsubjectAltName = @alt_names\n" >> $1
+  echo -e "[alt_names]" >> $1
+}
+
+function gen_start_end() { # $1 good/expired
+  # Right now we are just using harded good/bad for the date range validity.
```
nanoscopic: This should say hardcoded not 'harded'

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  then
+    START_DATE=$(TZ=UTC date +"%Y%m%d%H%M%SZ" -d "-1 day")
+    END_DATE=$(TZ=UTC date +"%Y%m%d%H%M%SZ" -d "+5 years")
+  else
+    START_DATE=$(TZ=UTC date +"%Y%m%d%H%M%SZ" -d "-10 day")
+    END_DATE=$(TZ=UTC date +"%Y%m%d%H%M%SZ" -d "-5 days")
+  fi
+}
+
+function create_site_crt() { # $1 domain, $2 ca, $3 message_digest, $4 rsa_bits, $5 good/expired, $6 subject, $7 altnames, $8 ipaltnames
```
nanoscopic: There are a lot of parameters here. Named parameters could potentially be used by way of getopts, but has been avoided here to make it plain bash without requiring getopts. Additionally using getopts for function calls within a bash script is somewhat non-standard and weird.

nanoscopic: It may be reasonable to use more environment variables for the arguments so that the code is more clear to read. I dislike polluting environment variable space so I have not done so here.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  
+  echo "Generating site CRT \"$1\""
+  
+  FILE=/tmp/ca_gen/site_$1.req.conf
+  site_request_conf $FILE $4
+  if [ "$7" != "0" ]; then append_altnames    $7 $FILE; fi
+  if [ "$8" != "0" ]; then append_ip_altnames $8 $FILE; fi
+  
+  openssl req -new -sha256 -key $DIR_KEYS/site_$1.key -subj "$6" -out $CSR_DIR/site_$1.csr -config $FILE
+  # It is possible to specify altnames the following way instead of constructing a conf file.
```
nanoscopic: This alternate way is weird in my mind since it depends on coping the default openssl.cnf and appending stuff to it. It does work but I feel could have strange issue specific to what setup this is being run on. ( OS etc )

nanoscopic: Probably this commented out other method should just be deleted and not kept in the code.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  #  -subj "$6" \
+  #  -reqexts SAN \
+  #  -config <(cat /etc/ssl/openssl.cnf \
+  #   <(printf "\n[SAN]\nsubjectAltName=DNS:example.com,DNS:www.example.com")) \
+  #  -out $CSR_DIR/site_$1.csr
+  
+  # Cleanest way to see the actual contents of the CSR to verify it was created correctly.
+  #openssl req -in $CSR_DIR/site_$1.csr -text -noout
+  
+  # Unclean very raw way to see the contents of the CSR
```
nanoscopic: This is barely if ever useful; the openssl text output is much better so this mentioned method could be deleted.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  # openssl asn1parse < $CSR_DIR/site_$1.csr
+  
+  gen_start_end $5
+  
+  # Irritatingly openssl ca command does not, by default, use the altnames specified in the CSR.
+  # It ignores them. They have to be re-specified. It may be pointless to specify them in the CSR as a result.
+  # The altnames can be copied from the request potentially, by using the "copy_extensions = copy" option of openssl conf
```
nanoscopic: The copy_extensions method should be investigated and would make a bunch of this code cleaner potentially if it was used. It is pretty odd to have to append the altnames twice.

At the very least it only makes sense when doing it twice to do it the second time since the first time has no real effect without copy_extensions being in use.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+  # Generate an extensions file; not extending default config since x509 command doesn't have a way to specify a
+  # full config file that I see.
+  #echo -e "[req_ext]\nsubjectAltName = @alt_names\n" >> conf/site_$1.conf
+  #echo -e "[alt_names]" >> conf/site_$1.conf
+  if [ "$6" != "0" ]; then append_altnames    $6 $DIR_CONF/site_$1.conf; fi
+  if [ "$7" != "0" ]; then append_ip_altnames $7 $DIR_CONF/site_$1.conf; fi
+  
+  gen_start_end $4
+  
+  #openssl x509 -req \
```
nanoscopic: This method has issues. It is not explained clearly why it is not used here and this commented out block should just be removed as a result. It was left initially because it seemed to be the desirable and suggested way to do it; but it has problems...

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+new_certs_dir = $DIR_CADATA/$1/db.certs
+database = $DIR_CADATA/$1/db.index
+serial = $DIR_CADATA/$1/db.serial
+certificate = $DIR_CERTS/ca_$1.crt
+private_key = $DIR_KEYS/ca_$1.key
+default_days = 365
+default_crl_days = 30
+default_md = sha256
+preserve = no
+RANDFILE = $DIR_CADATA/$1/db.random
```
nanoscopic: The RANDFILE option should actually not be specified here. It defaults to system random, and I've seen some problems where the file does not get created automatically in some cases.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+subjectAltName = @alt_names
+[alt_names]
+EOL
+}
+
+read -r -d '' ALTNAMES_VELUM_LN << EOM
+admin.devenv.caasp.suse.net
+admin
+admin.infra.caasp.local
+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
nanoscopic: It might be a good idea to use a more recognizable string to match as it is hard to match and see that the right number of a's are in use. Something like aaaabbbbccccddddeeee perhaps? Maybe even just using a short string. The long name is used because when the system generates a name it is a long randomly generated string of this length.

Additionally this name is the long uuid of the master node. So perhaps "master_long_uuid_24_chars" would be another option ( assuming it is a uuid; I haven't looked carefully at what generates the random strings to see it is in fact a uuid; it may or may not be )

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.infra.caasp.local
+testdomain.com
+EOM
+IP_ALTNAMES_VELUM=10.17.1.0
+
+read -r -d '' ALTNAMES_KUBEAPI_LN << EOM
```
nanoscopic: Would sure be nice to move this read magic into a function instead of using it repeatedly like magic.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb.infra.caasp.local
+EOM
+IP_ALTNAMES_DEX=172.24.0.1
+
+ALTNAMES_VELUM=$(echo "$ALTNAMES_VELUM_LN" | tr "\n" -)
+ALTNAMES_KUBEAPI=$(echo "$ALTNAMES_KUBEAPI_LN" | tr "\n" -)
+ALTNAMES_DEX=$(echo "$ALTNAMES_DEX_LN" | tr "\n" -)
+
+# Strip extra dashes at end of line
```
nanoscopic: Surely there is a way to avoid the extra dash??

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,337 @@
...
+create_site_crt ka      intermed2 sha256 2048 good    "/CN=ka$SITE_SUBJ"      $ALTNAMES_KUBEAPI $IP_ALTNAMES_KUBEAPI
+create_site_crt kb      intermed2 sha256 2048 good    "/CN=kb$SITE_SUBJ"      $ALTNAMES_KUBEAPI $IP_ALTNAMES_KUBEAPI
+create_site_crt da      intermed2 sha256 2048 good    "/CN=da$SITE_SUBJ"      $ALTNAMES_DEX     $IP_ALTNAMES_DEX
+create_site_crt db      intermed2 sha256 2048 good    "/CN=db$SITE_SUBJ"      $ALTNAMES_DEX     $IP_ALTNAMES_DEX 
+create_site_crt weak    intermed2 sha256 1024 good    "/CN=weak$SITE_SUBJ"    $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt sha1    intermed2 sha1   2048 good    "/CN=sha1$SITE_SUBJ"    $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt expired intermed2 sha256 2048 expired "/CN=expired$SITE_SUBJ" $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt badalt  intermed2 sha256 2048 good    "/CN=badalt$SITE_SUBJ"  $ALTNAMES_BAD     $IP_ALTNAMES_BAD
+
+cp $DIR_CERTS/ca_root.crt external_certs/ca_root.crt
```
nanoscopic: Q: Why not just generate the certs and keys into the folder, instead of moving them there after the fact?
A: Originally the files were intermixed, and the automatically generated names are prepended with 'ca_' or 'site_' etc, but the end desired names may be different. This code is generically useful as a standalone script, so it seems bad to customize it specific to this use case. I agree customizing it could be a bunch cleaner though.

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,52 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+    return _trust_chain_verify(_cert,_build_system_wide_certs_store())
+  end
+
+  def _build_system_wide_certs_store
+    cert_store = OpenSSL::X509::Store.new
+
+    system_certs = SystemCertificate.all
+    for system_cert in system_certs
+      cert = system_cert.certificate
+      cert_text = cert.certificate#get_certificate_text
```
nanoscopic: Comment at end of line is old code; should be dropped.

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,52 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+      cert_obj = read_cert(cert_text)
+      next unless cert_obj
```
nanoscopic: Isn't this an error condition? Shouldn't it throw an exception?

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,52 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+  def _trust_chain_verify(cert_obj,cert_store)
+    verify_error = ''
+
+    # setup callback
+    cert_store.verify_callback = proc do |preverify_ok, ssl_context|
+      begin
+        if preverify_ok != true || ssl_context.error != 0
+          cert_being_checked = ssl_context.chain[ssl_context.error_depth]
+          failed_cert_subject = cert_being_checked.subject
+          err_msg = "SSL Verification failed: #{ssl_context.error_string} (#{ssl_context.error}) while verifying #{failed_cert_subject}"
```
nanoscopic: Note that this error message is later actually output to the user in the UI. It is probably desirable to output some generic block here that is parsed later and keep the desired message within the UI output code instead of doing it this way.

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,52 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+        else
+          true
+        end
+      rescue Exception => e
+        verify_error += err_msg
+        false
+      end
+    end
+
+    if cert_store.verify( cert_obj ) == true
```
nanoscopic: It looks like these 5 lines could be collapsed to a ternary operator.

app/controllers/settings/external_cert_controller.rb
danielorf: Do we want it to error or just warn the user?

spec/fixtures/generate_external_certs.sh
nanoscopic: Added comment just that it is created with default content

app/controllers/settings/external_cert_controller.rb
nanoscopic: Dropped the old comment.

app/controllers/settings/external_cert_controller.rb
nanoscopic: Added a comment in code mentioning this is okay as a cert cannot get into the SystemCertificate table without having already passed a read_cert check.

app/controllers/settings/external_cert_controller.rb
nanoscopic: Leaving this as is for now. Will consider revising when messages are enhanced to be cleaner.

app/controllers/settings/external_cert_controller.rb
nanoscopic: Rubocop does pass and this is easy to read and understand. Not condensing as doing so would reduce code readability.

spec/fixtures/generate_external_certs.sh
nanoscopic: Changed to just stripping 'site_' off the names and using the full names to begin with so that any additional added certs don't need extra copy lines.

spec/fixtures/generate_external_certs.sh
nanoscopic: Added a comment to explain this.

spec/fixtures/generate_external_certs.sh
nanoscopic: Added a comment

spec/fixtures/generate_external_certs.sh
nanoscopic: Fixed

spec/fixtures/generate_external_certs.sh
nanoscopic: Removed the commented out portion.

spec/fixtures/generate_external_certs.sh
nanoscopic: Removed the unclean raw way comment and line of code.

spec/fixtures/generate_external_certs.sh
nanoscopic: Removed the commented out portion and fixed the comment above it.

spec/fixtures/generate_external_certs.sh
nanoscopic: Dropped the RANDFILE line.

spec/fixtures/generate_external_certs.sh
nanoscopic: Leaving this alone for now as I don't think it is worth changing. It works fine as is and isn't all that confusing.

spec/fixtures/generate_external_certs.sh
nanoscopic: Added comments to explain what this is doing. It's ugly but it's bash. It's already ugly...

spec/fixtures/generate_external_certs.sh
nanoscopic: Maybe. But not important. Leaving it alone. It works.

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,56 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+
+  def _build_system_wide_certs_store
+    cert_store = OpenSSL::X509::Store.new
+
+    system_certs = SystemCertificate.all
+    system_certs.each do |system_cert|
+      cert = system_cert.certificate
+      cert_text = cert.certificate
+      cert_obj = read_cert(cert_text)
+      # Certs must pass parsing to get into the SystemCertificate table; so we ignore failing parse
```
dannysauer: This is true in the current `update_ca_certificates` on SUSE (and IIRC RedHat), but some other scripts (Ubuntu, as a notable example) just blindly shove in anything with a .crt extension, assuming the files were put there on purpose.    Probably not important given Velum's likely future, but FWIW...

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,56 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+  end
+
+  # rubocop:disable Lint/RescueException
+  def _trust_chain_verify(cert_obj, cert_store)
+    verify_error = ""
+
+    # setup callback
+    cert_store.verify_callback = proc do |preverify_ok, ssl_context|
+      begin
+        if preverify_ok != true || ssl_context.error != 0
```
dannysauer: This `||` should probably be `or`.  I _think_ they both have lower precedence than comparison, but it always makes me nervous to see `||` near `=` :)

app/controllers/settings/external_cert_controller.rb
```diff
@@ -355,10 +357,56 @@ def warning_subjectaltname(cert, warning_messages, required_san_array, service_n
...
+        else
+          true
+        end
+      rescue Exception
+        verify_error += err_msg
+        false
+      end
+    end
+
+    if cert_store.verify(cert_obj) == true
```
dannysauer: Will this ever return a true value which isn't literally `true`?  If not, the comparison is probably unnecessary.  And this gives me an opportunity to use the `suggest change` button!
```suggestion
    if cert_store.verify(cert_obj)
```

spec/fixtures/external_certs/ca_intermed2.crt
```diff
@@ -0,0 +1,76 @@
...
+    Data:
+        Version: 3 (0x2)
+        Serial Number: 1 (0x1)
+    Signature Algorithm: sha256WithRSAEncryption
+        Issuer: O=US2, OU=US2 Unit, CN=US2
+        Validity
+            Not Before: Jan 30 23:59:02 2019 GMT
+            Not After : Jan 30 23:59:02 2020 GMT
+        Subject: O=US3, OU=US3 Unit, CN=US3
+        Subject Public Key Info:
```
dannysauer: Does nothing really complain about having an empty subject in the CA/intermediate CA?  Weird; I would've expected something to get upset about that. :D

app/controllers/settings/external_cert_controller.rb
nanoscopic: The difference is only that "or" short circuits. || does not short circuit. I will make the change; short circuiting does make sense here.

app/controllers/settings/external_cert_controller.rb
nanoscopic: I agree it is cleaner without the "== true". Adding.

spec/fixtures/external_certs/ca_intermed2.crt
nanoscopic: I think even the things there ( O and OU ) are only required because the configuration specifies them to be required. The only thing I think is needed is the CN ultimately.

app/controllers/settings/external_cert_controller.rb
nanoscopic: It does do the check, so it does ignore bad certs. It just doesn't complain about them. I feel like that is okay so at least this code doesn't continue doing something with an unparseable cert.

spec/features/settings/external_cert_feature_spec.rb
```diff
@@ -67,66 +154,66 @@
...
 
       click_button("Save")
       expect(page).to have_http_status(:success)
       expect(page).to have_content(success_message)
     end
 
     it "sucessfully lists Subject Alternative Names" do
-      attach_file("external_certificate_velum_cert", ssl_cert_file_b)
-      attach_file("external_certificate_velum_key", ssl_key_file_b)
+      attach_file("external_certificate_velum_cert", crt_v_b)
```
dannysauer: What are the difference between set "a" and "b"?  Maybe a comment somewhere so someone regenerating or otherwise following up knows what the intended difference is?
(it's entirely possible that I missed this in another script)

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
+#!/bin/bash
+
+# This script re-generates all of the keys and certs within the external_certs directory.
+# The files there are used by the spec/features/settings/external_cert_feature_spec test
+
+function init() {
```
dannysauer: Function declarations in bash/ksh/zsh should be either `function fname {` or `fname() {`; combining the two works in horrible-language bash, but will fail in other shells.  This is a bash script, so it doesn't currently matter, but is worth bearing in mind for general shell scripting purposes.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+  ROOT_DIR=/tmp/ca_gen
```
dannysauer: Should this perhaps use `${TMPDIR:-/tmp}` to allow a test environment which has exported `$TMPDIR` to generate wherever it's using for temporary files?  Or even better, let mktemp deal with that for you, and generate a guaranteed unique directory name at the same time.
```suggestion
  ROOT_DIR=$( mktemp -d )
```

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+  
+  mkdir -p $CSR_DIR
+  mkdir -p $DIR_CERTS
+  mkdir -p $DIR_CONF
+  mkdir -p $DIR_KEYS
+  mkdir -p $DIR_CADATA
+}
+
+function cleanup() {
+  rm -rf $ROOT_DIR
```
dannysauer: I like sticking a `-v` on the cleanup removal for log purposes, but that's up to you.
```suggestion
  rm -vrf $ROOT_DIR
```

dannysauer: Makes nice verbose output on the `mkdir` calls, too. ;)

spec/fixtures/generate_external_certs.sh
dannysauer: I get nervous about this unquoted `01`, because some shells will treat that as an octal number, which makes for weird issues later on.  Quotes are good. ;)

spec/fixtures/generate_external_certs.sh
dannysauer: That's a hyphen, not a dash. :p
I actually really like this idiom, though.  And I appreciate the proper use of `@` rather than `*` when the array is used later.  👍

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+    -infiles $CSR_DIR/ca_$1.csr 2>/dev/null
+}
+
+function append_altnames() { # $1 altnames, $2 filename
+  # Split up dash delimited string into array
+  IFS='-' read -r -a altnames <<< "$1"
+  for index in "${!altnames[@]}"
+  do
+      echo "DNS.$(($index+1)) = ${altnames[index]}" >> $2
+  done
```
dannysauer: It would be a little more efficient to put the concatenation after the while loop, so the file is only opened once instead of once per iteration.  Just move the `>> $2` after the `done`.

Also, in math expressions, variable names don't need the `$` prefix, which I think makes it a little easier to read.

spec/fixtures/generate_external_certs.sh
nanoscopic: As you might imagine this whole script is meant to be able to be used in a generic fashion but is not yet to that point. When it is made into a fully standalone script accepting a DSL for the list of roots and certs etc, then it will do more of what you are suggesting. For now your mktemp suggestion is great. Adding that. Thanks.

spec/fixtures/generate_external_certs.sh
nanoscopic: Sounds good to me. Adding.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+  # Various prompts for the portions of the subject would go here, as well as defaults.
+  # They can, though, be left blank when the subject is provided on the command line.
+  cat >>$1 <<EOL
+[req_distinguished_name]
+EOL
+  
+  echo -e "[no_ext]\nsubjectKeyIdentifier = hash\n" >> $1
+  echo -e "[req_ext]\nsubjectAltName = @alt_names\n" >> $1
+  echo -e "[alt_names]" >> $1
+}
```
dannysauer: You can wrap a bunch of commands inside braces to group I/O filehandles, as in
```bash
{
  echo "a"
  sort file
} > somefile
```
Then the file is only opened once, and it's easier to make changes moving lines around in the future / avoid mistakes by not accidentally using a `>` when someone meant `>>` .  It also makes things a bit easier to read, IMO.

spec/fixtures/generate_external_certs.sh
nanoscopic: do/done blocks collect all of their stdout and can be redirected as a whole?

spec/fixtures/generate_external_certs.sh
nanoscopic: But that's how it looks in Javascript. Just kidding. I'll change them.

spec/fixtures/generate_external_certs.sh
dannysauer: getopts is implemented as a builtin in bash, so there's really no extra requirement...  I agree that using global env vars is the worst option, though.

I prefer to document this in my scripts by assigning the positional parameters to descriptive variable names, which makes the rest of the function a tad easier to read.
```bash
function fname {
  something=$1
  anotherthing=$2 # another thing
}
```

spec/fixtures/generate_external_certs.sh
dannysauer: Examples often don't use copy extensions because it's unsafe for a "real" CA; someone could add the "basicconstraints:CA=true" extension or something like that, and completely wreck a CA's trustworthiness by becoming an intermediate.  Here, it'd be a good idea.

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+
+function gen_start_end() { # $1 good/expired
+  # Right now we are just using hardcoded good/bad for the date range validity.
+  # For most purposes it would be better to expose the clean text specification
+  # of both start and end as parameters.
+  if [ $1 = "good" ]
```
dannysauer: Generally, shell comparisons using bash should use `[[ ]]` instead of `[`, because the single bracket is implemented in bash as an alias for `test` while the `[[` is implemented as a first-class construct (or whatever it's called).  The single brackets are only good when maintaining compatibility with baseline POSIX sh, which this script already doesn't do. :D

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+    -md $2 \
+    -startdate $START_DATE -enddate $END_DATE \
+    -infiles $CSR_DIR/site_$1.csr
+}
+
+# Note that an altname is specified below. If req_ext is used, then this file cannot be used as is, because an empty
+# alt_names section is not allowed. It is done this way so the file can be copied and a list of altnames
+# be appended to the end of the copy.
+function generate_conf() { # $1 ca_name
+  cat >$DIR_CONF/ca_$1.conf << EOL
```
dannysauer: EOL, not EOF? :p

spec/fixtures/generate_external_certs.sh
dannysauer: The node name doesn't have to be included in any of these certs, as far as I can tell, so this is optional anyway.

spec/fixtures/generate_external_certs.sh
nanoscopic: I thought EOL made no sense too. It was copy pasted. ¯\_(ツ)_/¯

spec/fixtures/generate_external_certs.sh
dannysauer: You could read into an array and then join the array using a hyphen.  When you use the `*` expansion in double quotes, the elements of the array are joined using the first char of `$IFS`.  For example:
```
sauer@lightning:~$ a=(1 2 a b pie)
sauer@lightning:~$ g=$(IFS=-; echo "${a[*]}")
sauer@lightning:~$ echo $g
1-2-a-b-pie
```

spec/fixtures/generate_external_certs.sh
```diff
@@ -0,0 +1,322 @@
...
+create_site_crt dex_a       intermed2 sha256 2048 good    "/CN=da$SITE_SUBJ"      $ALTNAMES_DEX     $IP_ALTNAMES_DEX
+create_site_crt dex_b       intermed2 sha256 2048 good    "/CN=db$SITE_SUBJ"      $ALTNAMES_DEX     $IP_ALTNAMES_DEX 
+create_site_crt weak        intermed2 sha256 1024 good    "/CN=weak$SITE_SUBJ"    $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt sha1_digest intermed2 sha1   2048 good    "/CN=sha1$SITE_SUBJ"    $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt expired     intermed2 sha256 2048 expired "/CN=expired$SITE_SUBJ" $ALTNAMES_VELUM   $IP_ALTNAMES_VELUM
+create_site_crt badalt      intermed2 sha256 2048 good    "/CN=badalt$SITE_SUBJ"  $ALTNAMES_BAD     $IP_ALTNAMES_BAD
+create_site_crt noalt       intermed2 sha256 2048 good    "/CN=noalt$SITE_SUBJ"   0                 0
+
+for cert in $DIR_CERTS/ca_*.crt; do
+  mv "$cert" "external_certs/$(basename "$cert")"
```
dannysauer: These might benefit from a `mv -v` for log output, so one can easily see what's being put in place?

spec/fixtures/generate_external_certs.sh
dannysauer: Yup - from the `do` to the `done` is a block that can have redirection (including a pipe) applied after the `done`, kinda like its a separate script. Super handy for lots of things.

spec/fixtures/generate_external_certs.sh
nanoscopic: Added verbose output.

spec/fixtures/generate_external_certs.sh
nanoscopic: Have used this; thanks for the tip.

spec/fixtures/generate_external_certs.sh
nanoscopic: Moved the append to after the done.

spec/fixtures/generate_external_certs.sh
nanoscopic: Have dropped the 'function' at the start of each function.

spec/fixtures/generate_external_certs.sh
nanoscopic: Changed to using double brackets on all the if conditions. Also changed this specific if to use double equals as well.

spec/fixtures/generate_external_certs.sh
nanoscopic: Added quotes around it. Notably tried just 1 by itself without the 0 and that does not work.

spec/fixtures/generate_external_certs.sh
nanoscopic: Have changed to using named parameters using associative arrays thanks to your help in chat on how to do that. Thanks for all the ideas; it enhances the readability of the script a bunch.

spec/fixtures/generate_external_certs.sh
nanoscopic: Thanks for the explanation. Have additionally done away with all of this thanks to your help. :)

spec/features/settings/external_cert_feature_spec.rb
dannysauer: Ah, that makes sense, and is definitely a good test to include.

dannysauer: Ah, that makes sense, and is definitely a good test to include.

PR #722, opened 2019-02-01
init: added check for ldap

vitoravelino:
Added a new check to verify if the ldap instance is ready before running
velum dashboard in order to avoid a connection error when creating a new
user.

bsc#1121064

### DIFF
--- a/bin/check_ldap.rb
+++ b/bin/check_ldap.rb
@@ -1,0 +1,21 @@
+# This is a rails runner that will print the status of the database
+# Possible outcomes:
+#
+#   * `LDAP_READY`: connection successfull
+#   * `LDAP_FAIL`: cannot authenticate within ldap server
+#   * `LDAP_DOWN`: cannot connect to the ldap server
+
+def ldap_ready?
+  ldap_config = Velum::LDAP.ldap_config
+  ldap = Net::LDAP.new(ldap_config)
+
+  if ldap.bind
+    puts "LDAP_READY"
+  else
+    puts "LDAP_FAIL"
+  end
+rescue Errno::ECONNREFUSED, Net::LDAP::Error
+  puts "LDAP_DOWN"
+end
+
+ldap_ready?
--- a/bin/run
+++ b/bin/run
@@ -8,5 +8,34 @@ setup_root_ca() {
   update-ca-certificates
 }
 
+check_ldap() {
+  set +e
+
+  TIMEOUT=90
+  COUNT=0
+  RETRY=1
+
+  while [ $RETRY -ne 0 ]; do
+    case $(bundle exec rails r bin/check_ldap.rb | grep LDAP) in
+      "LDAP_DOWN")
+        if [ "$COUNT" -ge "$TIMEOUT" ]; then
+          printf " [FAIL]\n"
+          echo "Timeout reached, exiting with error"
+          exit 1
+        fi
+        echo "Waiting for ldap to be ready in 5 seconds"
+        sleep 5
+        COUNT=$((COUNT+5))
+        ;;
+      *)
+        echo "LDAP is ready"
+        break
+        ;;
+    esac
+  done
+  set -e
+}
+
 setup_root_ca
+check_ldap
 bundle exec "puma -C config/puma.rb"


bin/check_ldap.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+#   * `LDAP_READY`: connection successfull
+#   * `LDAP_FAIL`: cannot authenticate within ldap server
+#   * `LDAP_DOWN`: cannot connect to the ldap server
+
+def ldap_ready?
+  ldap_config = Velum::LDAP.ldap_config
+  ldap = Net::LDAP.new(ldap_config)
+
+  if ldap.bind
```
vitoravelino: I also thought about simplifying this to just 

```
ldap.bind
puts "LDAP_READY"
```

because what we want to know is if ldap instance is up or not. It doesn't matter if it's able to authenticate or not. If we do that change we can change the `*` in the bash script to `LDAP_READY`.

bin/check_ldap.rb
```diff
@@ -0,0 +1,21 @@
...
+    puts "LDAP_READY"
+  else
+    puts "LDAP_FAIL"
+  end
+rescue Errno::ECONNREFUSED
```
mssola: I'm afraid this is not enough. You should be using `Net::LDAP::Error`, since this is the error class that wraps all kinds of errors from LDAP connections.

bin/check_ldap.rb
```diff
@@ -0,0 +1,21 @@
...
+  puts "LDAP_DOWN"
+end
+
+ldap_ready?
```
mssola: Is it really needed to have it as a method?

bin/init
```diff
@@ -66,6 +67,35 @@ setup_cpi() {
...
+check_ldap() {
+  set +e
+
+  TIMEOUT=90
+  COUNT=0
+  RETRY=1
+
+  while [ $RETRY -ne 0 ]; do
+    case $(bundle exec rails r bin/check_ldap.rb | grep LDAP) in
+      "LDAP_DOWN")
```
mssola: Shouldn't you have "LDAP_FAIL" here as well?

mssola (changes requested):

bin/check_ldap.rb
vitoravelino: The method need to be called somehow unless we don't use a method declaration,

bin/init
vitoravelino: See https://github.com/kubic-project/velum/pull/722#discussion_r253027250.

vitoravelino: I wasn't able to check if the authentication was working properly with my local debug.

bin/check_ldap.rb
vitoravelino: I saw `Deprecation warning: Net::LDAP::ConnectionRefused will be deprecated. Use Errno::ECONNREFUSED instead.` The reported error was a connection refused but if I recall correctly using the general ldap error exception also works.

mssola: Then I'd say to keep them both.

bin/check_ldap.rb
mssola: What I meant is that you don't need to implement it as a method and then call it: you can simply have the code as-is :)

vitoravelino: I know, I just followed the same pattern as `check_db.rb` file.

bin/init
mssola: Well, I'd say it's a matter of deciding what it would make sense. Hopefully credentials should be alright, but if they aren't, then it means that at least the LDAP server is reachable, which is enough for this purposes. Hence ignore my comment :grin:

bin/check_ldap.rb
vitoravelino: The authentication is probably not working because if I add `Net::LDAP::Error` velum doesn't start. So a generic error is happening. I'll try to debug a bit more.

flavio (changes requested):
Overall LGTM, just update the error handling as requested by Miquel.

vitoravelino:
I've realized that I added the check to the wrong file. It should be checked in the `run` file that belongs to the velum-dashboard container and not in the `init` that belongs to the dashboard initContainer.

mssola (dismissed):

MalloZup (approved):
Thx Vito. Lgtm please retrigger the ci and pray, (from famous book eat pray love and retrigger Ci)

MalloZup:
@mssola  can we merge this?

PR #726, opened 2019-02-08
oidc: fixed instructions for manual configuration (bsc#1123711)

mssola:
bsc#1123711


mssola:
It will need to be backported to 3.0 as well.

PR #732, opened 2019-02-20
Check if last orchestration is still in progress, bsc#1120752

mjura:


app/models/orchestration.rb
```diff
@@ -35,12 +35,12 @@ def run
   # rubocop:enable Rails/SkipsModelValidations,Metrics/CyclomaticComplexity
 
   def self.run(kind: :bootstrap, params: nil)
-    raise OrchestrationOngoing unless runnable?
+    raise OrchestrationOngoing if runnable?
```
MaximilianMeister: to be correct i would keep this `raise OrchestrationOngoing unless runnable?`

otherwise it read as raise Error when runnable

app/models/orchestration.rb
```diff
@@ -35,12 +35,12 @@ def run
...
     Orchestration.create!(kind: kind, params: params).tap(&:run)
   end
 
   def self.runnable?
-    Orchestration.in_progress.empty?
+    Orchestration.last.try(:status) == "in_progress"
```
MaximilianMeister: this then should be `! Orchestration.last.try(:status).in_progress?`

app/models/orchestration.rb
mjura: ok

app/models/orchestration.rb
mjura: `Orchestration.last.try(:in_progress?)` works fine as well

app/models/orchestration.rb
```diff
@@ -40,7 +40,7 @@ def self.run(kind: :bootstrap, params: nil)
   end
 
   def self.runnable?
-    Orchestration.in_progress.empty?
+    ! Orchestration.last.try(:in_progress?)
```
MalloZup: `!` imho is more rubyst and readable with `unless`

MaximilianMeister (approved):

mjura:
Backport to 3.0 https://github.com/kubic-project/velum/pull/735

PR #733, opened 2019-02-20
Jenkinsfile: Switch to dynamic library fetching and drop branch

hwoarang:
Instead of having the library hardcoded to Jenkins master, we can fetch
it dynamically. We also drop the usage of library branches since it does
not make sense to maintain such a thing in the CI. The master branch
should be able to handle both development and release branches.


hwoarang:
@MalloZup @ereslibre any clue what the travis error is? i do not think it's related to the PR

ereslibre:
@hwoarang Looks like a flake.

MalloZup:
i ddin't check it :robot:

MalloZup (approved):

